PR #44, opened 2020-10-09
Add linter github workflow

jenniebrown:


mayaspivak (approved):
LGTM, just add documentation on what needs to be updated when a new directory with python code is added.

PR #47, opened 2020-10-09
Copy relevant prototype code over to the main repo

aaronsn:
- Drop helloworld, sql, and testing_data_files folders as those were just for testing/examples
- Drop python/main.py, python/requirements.in, and python/requirements.txt. These were just for testing with cloud functions.
- Copy README over, minus some cloud function details
- Update .github/workflows to match prototype
- Remove terraform setup for cloud functions and BQ testing tables.


prototype.tf
```diff
@@ -0,0 +1,346 @@
...
+    percent         = 100
+    latest_revision = true
+  }
+  autogenerate_revision_name = true
+}
+
+/* [END] Cloud Run Setup */
+
+
+/* TODO: replace scheduler setup with Airflow */
```
jenniebrown, line 224: nit: prefer to create an issue instead of leaving TODOs in code

jenniebrown (changes requested):
Do we still need to keep the python/ingestion directory structure now that we've removed cloud functions? What's stopping us from bumping ingestion to a top level directory?

ldbnr4 (changes requested):
Can we put files that are specific to a service in the owning service directory? For instance all the `_to_bq.py` code can live in the `run_gcs_to_bq` directory right?

Similar to Jennie's comment, I think we can move util.py to the top level directory and use the ingestion directory for resources shared between services.

aaronsn:
> Can we put files that are specific to a service in the owning service directory? For instance all the `_to_bq.py` code can live in the `run_gcs_to_bq` directory right? Similar to Jennie's comment, I think we can move util.py to the top level directory and use the ingestion directory for resources shared between services.

I think there are some good reasons to keep almost all python code in the python/ folder. It simplifies refactoring/reusing code, which especially given the micro-services approach we've taken I think is really valuable. Eg, you can spin up a new service and immediately start reusing code without any extra work, you can refactor code to share between two services without moving/restructuring files and deps, etc. This is loosely modeled after Google's "monorepo" strategy, but instead of it being a "monorepo" it's a "monofolder" where all python code is under one directory. Also worth noting that this setup doesn't preclude splitting out packages within the python/ folder. Eg you could have:
```
python/
  -> gcs_to_bq/
  -> upload_to_gcs/
  // for shared ingestion logic
  -> ingestion/
```
I just started with everything in ingestion/ because I wasn't sure which things belonged where and it was easier to start with this, but I do think it likely makes sense to split this up at some point.


> Do we still need to keep the python/ingestion directory structure now that we've removed cloud functions? What's stopping us from bumping ingestion to a top level directory?

It could definitely go at the top-level. I do somewhat prefer this setup for a few reasons, though I'm less opinionated on this piece:
- Easy to add or split up packages without polluting the root-level. For example, it's easy to distinguish at the root level which directories are entry points for cloud run services, whereas business logic is under python/
- Less boilerplate for the Dockerfile - you can just copy the whole python directory over instead of doing once per package. This is a minor point though since you have to modify the Dockerfile to install packages anyway.

Lmk what you all think or if I'm missing some reason it makes sense to do this differently.

ldbnr4:
> Lmk what you all think or if I'm missing some reason it makes sense to do this differently.

If this structure will set us up for extensibility, lgtm.

ldbnr4 (approved):

jenniebrown:
> > Do we still need to keep the python/ingestion directory structure now that we've removed cloud functions? What's stopping us from bumping ingestion to a top level directory?
> 
> It could definitely go at the top-level. I do somewhat prefer this setup for a few reasons, though I'm less opinionated on this piece:
> 
> * Easy to add or split up packages without polluting the root-level. For example, it's easy to distinguish at the root level which directories are entry points for cloud run services, whereas business logic is under python/
> * Less boilerplate for the Dockerfile - you can just copy the whole python directory over instead of doing once per package. This is a minor point though since you have to modify the Dockerfile to install packages anyway.
> 
> Lmk what you all think or if I'm missing some reason it makes sense to do this differently.

The main reason I see for moving up a level would be that the current setup requires all services to import all libraries from every file even if they don't need them. It could be that it's not a big deal, but importing large libraries like pandas, etc. could cause unnecessary bloat in the image files. When I was testing, my cloud run images were hitting upwards of 200MB.

I think it's okay to table this for now and come back to it if it becomes a problem.

aaronsn:
> The main reason I see for moving up a level would be that the current setup requires all services to import all libraries from every file even if they don't need them. It could be that it's not a big deal, but importing large libraries like pandas, etc. could cause unnecessary bloat in the image files. When I was testing, my cloud run images were hitting upwards of 200MB.

As far as having ingestion at the root level vs inside the python directory, that shouldn't make a difference. The reason they're bloated is because I moved all the ingestion code into one package python/ingestion. I do think it makes sense to split that up, but I think the way it was split earlier wasn't allowing for shared logic so I think a little more work needs to happen there. Eg have an `ingestion` package that has shared logic, a `data_processing` package that just has the pandas code, etc. Also as a side note, copying the whole python/ folder in the Dockerfile just copies our code, doesn't actually install the deps, so I think that's fairly harmless.

I filed https://github.com/SatcherInstitute/health-equity-tracker/issues/48 for this

prototype.tf
aaronsn: I filed an issue and linked to it from the TODO

aaronsn:
Fixed linter issues and resolved comments, take a look and let me know if there are any more changes I should make. Thanks!

jenniebrown (approved):
Thanks for doing this!

PR #51, opened 2020-10-13
Updating primary care access

kkatzen:
Updating error handling, consolidating all states to one BigQuery job, moving state names to constant file

### DIFF
--- a/python/ingestion/primary_care_access_to_bq.py
+++ b/python/ingestion/primary_care_access_to_bq.py
@@ -1,4 +1,3 @@
-import logging
 import math
 from pandas import DataFrame, read_excel
 from .gcs_to_bq_util import append_dataframe_to_bq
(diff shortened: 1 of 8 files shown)


python/ingestion/primary_care_access_to_bq.py
```diff
@@ -1,108 +1,51 @@
...
-                data.append([row[0], row[1], row[2],
-                             row[108], row[109], row[110]])
-            new_dataframe = DataFrame(
-                data=data,
-                columns=('county_fips_code',
-                         'state_name',
-                         'county_name',
-                         'num_primary_care_physicians',
-                         'primary_care_physicians_rate',
-                         'primary_care_physicians_ratio'))
```
aaronsn, line 95: Any particular reason for dropping the ratio field?

aaronsn (approved):
You have some lint errors to fix, other than that looks good! Did you get it working?

python/ingestion/primary_care_access_to_gcs.py
```diff
@@ -1,18 +1,20 @@
...
+from .constants import STATE_NAMES
 
 
 _FILEPATH = '{}-{}.xlsx'
-_URL1 = 'https://www.countyhealthrankings.org/sites/default/files/media/document/2020 County Health Rankings {} Data - v1_0.xlsx'
-_URL2 = 'https://www.countyhealthrankings.org/sites/default/files/media/document/2020 County Health Rankings {} Data - v1.xlsx'
+_URL1 = """https://www.countyhealthrankings.org/sites/default/files/media/document/
+    2020 County Health Rankings {} Data - v1.xlsx"""
+_URL2 = """https://www.countyhealthrankings.org/sites/default/files/media/document/
+    2020 County Health Rankings {} Data - v1_0.xlsx"""
```
aaronsn, line 9: I don't think this works - it adds extra spacing into the string because of the indent. See https://stackoverflow.com/questions/10660435/pythonic-way-to-create-a-long-multi-line-string

I think you want to do:
```
_URL1 = ("first part of string"
         "second part of string")
```

PR #57, opened 2020-10-15
Creates the Airflow local development infrastructure with an initial boilerplate DAG

ldbnr4:


mayaspivak (changes requested):
Looks good! My only comments are on directory structure. I'd put all the new files under an airflow or ingestion or whatever you want to name it directory. If a file is only used for local dev, either prefix it with dev or put in under a dev subdirectory.

.github/workflows/linter.yml
```diff
@@ -48,3 +48,4 @@ jobs:
           mypy --config-file=.github/linters/mypy.ini data_server/
           mypy --config-file=.github/linters/mypy.ini run_ingestion/ python/ingestion/
           mypy --config-file=.github/linters/mypy.ini run_gcs_to_bq/ python/ingestion/
+          mypy --config-file=.github/linters/mypy.ini dags/
```
mayaspivak, line 51: Does the path now need to be updated here?

mayaspivak (approved):

.github/workflows/linter.yml
ldbnr4: yes. good catch

PR #60, opened 2020-10-16
Add terraform steps to deployToTest workflow

mayaspivak:
1. Move terraform files into their own config directory.
2. Refactor into multiple terraform files for readability.
3. Remove scheduler resources from terraform config since they wont be used. The ingestion service can still be triggered manually or through pubsub.
4. Add the test values for the inputs.
5. Add the terraform steps to the workflow.
6. Add condition so that workflow is not run in forked repos. This is currently commented out along with the run on push trigger so that the workflow can be tested.


.github/workflows/deployToTest.yml
```diff
@@ -39,3 +42,29 @@ jobs:
...
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v1
+        working-directory: ./config
+      - name: Terraform Init
+        working-directory: ./config
+        run: terraform init
+      - name: Terraform Format
+        working-directory: ./config
+        run: terraform fmt -check
+      - name: Terraform Plan
```
jenniebrown, line 54: Is there benefit to running terraform plan here if it's just going to auto approve? Does this give any information to human users?

config/run.tf
```diff
@@ -0,0 +1,107 @@
+/* [BEGIN] Cloud Run Setup */
+
+# Create a Pub/Sub topic to trigger the GCS-to-BQ service.
+resource "google_pubsub_topic" "notify_data_ingested" {
```
jenniebrown, line 4: I think we also need the upload_to_gcs topic here

config/main.tf
```diff
@@ -0,0 +1,30 @@
+# Specify the provider, here
+provider "google" {
+  project     = var.project_id
+  region      = var.compute_region
+  credentials = var.gcp_credentials
```
jenniebrown, line 5: Are the credentials just so github can deploy via terraform?

jenniebrown (changes requested):

config/main.tf
mayaspivak: Yep. I downloaded them from the test project and saved them to a github secret.

config/run.tf
mayaspivak: Good catch, thanks! I must have accidentally deleted it in the refactor.

.github/workflows/deployToTest.yml
mayaspivak: terraform plan was included in the example starter workflow. I assumed it was necessary to refresh the state before apply, but reading the documentation more closely, it looks like terraform refreshes the state before any command. I will remove it.

jenniebrown (approved):

PR #61, opened 2020-10-16
Add some unit tests and fix imports

jenniebrown:
- Adds unit tests for gcs_to_bq.py
- Moves setup.py outside of ingestion/
- Fixes import style (relative -> absolute) and some lint errors


.gitignore
```diff
@@ -1,12 +1,13 @@
+.terraform/plugins
 .vscode
 .venv
 __pycache__
-data_ingestion/__pycache__
-.terraform/plugins
 apply_terraform
-jenniebrown.tfvars
+jenniebrown*
```
aaronsn, line 6: Btw, you can have a personal git exclude rule, which I think is preferable for things like this. https://www.atlassian.com/git/tutorials/saving-changes/gitignore#personal-git-ignore-rules

README.md
```diff
@@ -37,6 +41,17 @@ Install Docker Desktop ([Get Docker](https://docs.docker.com/get-docker/))
...
+## Testing
+
+Unit tests can be run using pytest. Running pytest will recursively look for and execute test files.
+
+```bash
+pip install pytest
+pytest
+```
+
+To test from the packaged version of the ingestion library, run `pip install -e python/` before testing.
```
aaronsn, line 53: does `-e python/` install all the packages within `python/`?

python/setup.py
```diff
@@ -0,0 +1,5 @@
+from setuptools import setup, find_packages
+
+setup(name='ingestion',
+      packages=find_packages(),
```
aaronsn, line 4: Neat trick! There's some stuff in the README about putting a setup.py file in each package, can you update that?

python/setup.py
```diff
@@ -0,0 +1,5 @@
+from setuptools import setup, find_packages
+
+setup(name='ingestion',
```
aaronsn, line 3: does this make sense since it's using all packages in the folder? Like if we have `python/data_serving`, what would this be?

Can you still install just `python/<some_package>` independently or now you have to install the whole thing?

run_gcs_to_bq/Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY ./python ./python
 # Install production dependencies.
 RUN pip install -r run_gcs_to_bq/requirements.txt
 # Install shared packages
-RUN pip install ./python/ingestion
+RUN pip install ./python
```
aaronsn, line 18: What was the motivation for changing this to install everything in `python/`?

python/tests/test_gcs_to_bq.py
```diff
@@ -0,0 +1,132 @@
...
+        with patch('ingestion.gcs_to_bq_util.bigquery.Client') as mock_client:
+            # Set up mock calls
+            mock_instance = mock_client.return_value
+            mock_table = Mock()
+            mock_instance.dataset.return_value = mock_table
+            mock_table.table.return_value = 'test-project.test-dataset.table'
+
+            gcs_to_bq_util.append_dataframe_to_bq(
+                test_frame.copy(deep=True), "test-dataset", "table",
+                col_modes=['REPEATED', 'REQUIRED'])
```
aaronsn, line 76: col_modes should be a dictionary of column to mode

aaronsn (approved):
Thanks for doing this!

.gitignore
jenniebrown: Cool, I didn't know about that!

README.md
jenniebrown: It will install all of the packages defined in setup.py in python/ in edit mode, which means you can edit them without having to reinstall.

python/setup.py
jenniebrown: Hm, so I don't think you would be able to install just `python/<some_package>` but my thought was since we are copying the code into the docker image, if we only copy the packages we need then this would install them all properly. I think the name 'ingestion' should probably be changed to something like 'health-equity-tracker'. But now that I write it down it seems a bit hacky. WDYT?

run_gcs_to_bq/Dockerfile
jenniebrown: With find_packages() we could install all the required packages and only copy over the packages we need. See comment above

README.md
aaronsn: Wooooah that is cool, I didn't know that

python/setup.py
aaronsn: To make sure I'm understanding correctly, if you need `package1` but not `package2` the idea would be you copy `python/package1` but not `python/package2`, and then when you do pip install `./python` it will only install `package` because only `package` is actually there? You'd also need to manually copy over the `setup.py` file. So the dockerfile would be
```
COPY ./python/package1 ./python/package1
COPY ./python/setup.py ./python/setup.py
RUN pip install ./python // automatically installs package1 but not package2
```

Interesting idea. I think I might not be understanding the motivation. You still need to modify the Dockerfile to specify which packages to copy over. It seems like the main benefit is it makes local dev easier because you can run `pip install -e python/` just once? But I don't think it's that bothersome to do it once per directory, especially with the `-e` option.

If there's something I'm missing here let me know! I don't think this is *too* different from what we had before, but all else being equal I prefer the explicitness of the Dockerfile directly installing the deps it needs, and the isolation of having each package be separate and control its own setup. I don't have any super concrete reasons, just that it feels better to me haha. Like if someone didn't realize that the COPY command was also controlling what was being installed and just decided to change the individual COPY lines to just copy the whole `./python` package, and then it breaks because it's missing external deps.

jenniebrown: Yeah I think you're right. There could be some merit still in only copying over the packages we need but having the setup.py file at the top level is more confusing than helpful. I moved setup.py back into the ingestion/ folder and reverted the Docker files.

aaronsn: Yeah definitely valid to just copy over the relevant package code, that's an easy 1-line change to the dockerfile and doesn't affect package structure, you're welcome to do that if you think that's better

aaronsn (approved):

PR #80, opened 2020-10-19
Use a remote backend (GCS) to store terraform state information.

mayaspivak:


jenniebrown (approved):

slagathorr (approved):
So this will work, but in the future may want to consider passing the variable of the bucket name into the backend.tf file directly if that's possible, because it's usually a better practice to have it all of it contained in one place instead of splitting the configuration between the tf script and the shell command. Not a dealbreaker, but it's just more common.

mayaspivak:
> So this will work, but in the future may want to consider passing the variable of the bucket name into the backend.tf file directly if that's possible, because it's usually a better practice to have it all of it contained in one place instead of splitting the configuration between the tf script and the shell command. Not a dealbreaker, but it's just more common.

I tried this first but it's not allowed.

PR #82, opened 2020-10-22
Add data server Cloud Run service to terraform config

mayaspivak:


config/test/test.tfvars
```diff
@@ -14,3 +14,7 @@ notify_data_ingested_subscription_name = "notify-data-ingested-sub"
 gcs_to_bq_invoker_identity_id = "gcs-to-bq-invoker"
 gcs_to_bq_runner_identity_id = "gcs-to-bq-runner"
 gcs_to_bq_runner_role_id = "test_gcs_to_bq_runner"
+data_server_service_name = "data-server-service"
+data_server_image_name = "data-serving"
```
jenniebrown, line 18: nit: We don't use "serving" anywhere else so it might be more consistent to use "data-server" here instead

jenniebrown (approved):
LGTM, just one nit about naming before you submit

PR #90, opened 2020-10-27
Refactor data source ingestion code

mayaspivak:
- Create a datasources python package
- Create an abstract base class for datasources with default implementations of ingestion methods
- Create an extending class for each data source from the prototype (upload to gcs and write to bq code is copied from the ingestion package)
- Create a dictionary of data source id to data source class

Note that this new package is not used yet. In a subsequent PR, I will refactor the ingestion code to use the datasource classes and remove duplicate code.


python/datasources/data_source.py
```diff
@@ -0,0 +1,37 @@
+from abc import ABC
+from ingestion import di_url_file_to_gcs, gcs_to_bq_util
+
+
+# Abstract base class for all data sources ingested by the Health Equity Tracker.
+# This also includes default implementations for each of the ingestion methods.
+class DataSource(ABC):
```
ldbnr4, line 7: Should we add a @abstractmethod @staticmethod get_id()?

ldbnr4 (changes requested):
Does `python/ingestion/util.py` need to be updated?

python/datasources/data_source.py
mayaspivak: Done

mayaspivak:
> Does `python/ingestion/util.py` need to be updated?

Yes, but I am breaking up into two PRs instead of one giant PR. I have a note about this in the initial PR comment.

python/datasources/data_source.py
ldbnr4: Nice. I was thinking we could add the `@abstractmethod` decorator but not a strong opinion.

ldbnr4 (approved):

ldbnr4:
> > Does `python/ingestion/util.py` need to be updated?
> 
> Yes, but I am breaking up into two PRs instead of one giant PR. I have a note about this in the initial PR comment.

Sounds good. Thanks for cleaning this up!

python/datasources/data_source.py
```diff
@@ -0,0 +1,48 @@
...
+        # in column names lower case.
+        frame.rename(columns=lambda col: col.replace(' ', '_'), inplace=True)
+        frame.rename(columns=lambda col: col.replace(
+            '-', '_').lower(), inplace=True)
+
+        gcs_to_bq_util.append_dataframe_to_bq(frame, dataset, self.get_table_name())
+
+    def export_to_gcs(self):
+        # TODO: Implement
+        pass
```
jenniebrown, line 48: Have you given any thought to how we will handle exporting data that doesn't correspond 1:1 with an ingestion source? 

My other thought is that including the export code here necessitates that every service packages all the data sources and ingestion code. From a vulnerability perspective, it would be better if we could decouple them more so that e.g. the export service doesn't need to load all that code. What do you think?

mayaspivak: For data that includes more than one ingestion source we could potentially group by outcome? I am not sure whether we plan to display more than one health outcome on a single chart. 

For your second thought, I was thinking about the same thing but ultimately decided to trade off that for improved readability. Perhaps if the export code is pretty much the same for every source we don't need this here and can remove it in the future.

jenniebrown: For the first, I could imagine us wanting to join e.g. county adjacency with some outcome. Thinking about it more, we could probably export directly from a saved SQL query like Brian did in the prototype.

Second part sgtm

jenniebrown (approved):

Issue #84, opened 2020-10-22
Implement Data Server API


jenniebrown:
The Data Server will expose two API methods for the frontend: /getMetadata and /getDataset. Implement these two in a cloud run service. See https://docs.google.com/document/d/1viK-Z8J2Y92ZaM4SqxUiylQiwkVsTfarjVeTH4LkTww/edit for details.

mayaspivak:
I believe what is left here is to implement getDataset method.

PR #91, opened 2020-10-28
Data serving

jenniebrown:
For issue #84 
- Creates /getMetadata endpoint which serves a newline-delimited file to the client
- Creates a DatasetCache class which can hold datasets (and the metadata file) in memory with TTL per item
- Moves test requirements.txt into top-level requirements/. I plan to move all requirements.txt files here in the future for easier dependency management
- Updates the Dockerfile with necessary dependencies to run the server.


python/data_server/dataset_cache.py
```diff
@@ -0,0 +1,42 @@
+import threading
+
+import cachetools
+
+from data_server import gcs_utils
+
+
+class DatasetCache():
```
mayaspivak, line 8: Add a comment or docstring for the class

data_server/test_data_server.py
```diff
@@ -0,0 +1,82 @@
...
+{"label1":"value4","label2":["value5a","value2b","value2c"],"label3":"value12"}
+{"label1":"value5","label2":["value6a","value2b","value2c"],"label3":"value15"}
+{"label1":"value6","label2":["value7a","value2b","value2c"],"label3":"value18"}
+"""
+
+
+def get_test_data(gcs_bucket: str, filename: str):
+    """Returns the contents of filename as a bytes object. Meant to be used to
+    patch gcs_utils.download_blob_as_bytes."""
+    print('In get_test_data')
```
mayaspivak, line 27: Should this still be here? Or was it just used for debugging?

mayaspivak (approved):
This is great! Thanks for adding unit tests with the new code! 

Just a few minor comments.

python/data_server/dataset_cache.py
```diff
@@ -0,0 +1,42 @@
...
+        getDataset will return the dataset from memory if it exists in the
+        cache. Otherwise, it will request the file from GCS and update the
+        cache on success.
+
+        gcs_bucket: Name of GCS bucket where the dataset is stored.
+        table_id: Name of the data set file to access.
+
+        Returns: Bytes object containing the dataset if successful. Throws
+        NotFoundError on failure."""
+        with self.cache_lock:
```
aaronsn, line 30: It's been a while since I've written concurrent code so this may be a silly question, but do we need the lock? Setting a value in the cache seems like an atomic operation, and whether a reader gets the contents before or after it's updated doesn't really matter, right?

aaronsn (approved):
Looks good! I applaud your dedication to writing high quality tests :D

data_server/Dockerfile
```diff
@@ -9,9 +9,13 @@ ENV PYTHONUNBUFFERED True
 ENV APP_HOME /app
 WORKDIR $APP_HOME
 COPY ./data_server ./data_server
+COPY ./python/data_server ./python/data_server
 
 # Install production dependencies.
+RUN python -m pip install --upgrade pip
```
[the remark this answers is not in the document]
jenniebrown: I added it because when I was building my containers it was giving a warning that it wasn't using the latest version of pip. I think it would make sense to add it to the other Dockerfiles. I'll send that as a followup CL.

data_server/test_data_server.py
jenniebrown: My bad, this one slipped through

python/data_server/dataset_cache.py
```diff
@@ -0,0 +1,42 @@
+import threading
+
+import cachetools
+
+from data_server import gcs_utils
+
+
+class DatasetCache():
+    def __init__(self, max_cache_size=8, cache_ttl=2 * 3600):
```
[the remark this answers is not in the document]
jenniebrown: - 2 hrs is a gut-feeling number based on how often I thought it would make sense to refresh and how long I thought an individual instance might run. Right now that is much less than 2 hours, but this is tunable.
- I chose TTL because the data does get updated periodically and we don't want to be serving stale data indefinitely. Depending on how our performance is, we might choose to decrease the TTL to keep data fresher.
- I thought about this also, and think that it will make sense to count bytes rather than num elements in the future. I think that we might want to consider frequency-based solutions as well.

Basically this is a middle ground first attempt that we can iterate on based on how our needs shake out.

python/data_server/dataset_cache.py
jenniebrown: The documentation for cachetools states that they are not thread-safe. Since we modify the cache we don't want to operate on the bare object. As an example, we could have a situation where there is 1 spot left in the cache and 2 threads try to write to it at the same time. Since they both see that there are 7 items in the cache, we end up leaking memory because it doesn't properly evict.

python/data_server/dataset_cache.py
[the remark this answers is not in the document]
aaronsn: Oh whoops I forgot about the data being updated haha, it definitely needs a TTL

Yeah that all sounds good!

python/data_server/dataset_cache.py
aaronsn: Interesting, that makes sense. Does that apply for reading though? Could the lock just be around the writing?

I'm not sure that it will actually matter in practice because I'm guessing reading the cache will be fast, but my first thought when I saw this was I wonder if there would be performance issues of a bunch of connections trying to lock to read the cache at the same time.

Anyway, I leave this up to you

jenniebrown: Generally no, since it's still possible for a thread to try to read something that's being actively written. I'm not sure what the implementation of TTLCache is around this, but I don't think it's safe to assume that it is atomic.

I think that both reads and writes should be pretty fast, but this is something we will want to keep an eye on when we start receiving significant traffic.

PR #97, opened 2020-10-29
Update deployToTest Workflow

mayaspivak:
Now that manual testing is complete, run deployToTest workflow on PR merge into master branch.

Double check the workflow is running due to the correct action and repository/branch before running any steps that actually modify the test environment.


ldbnr4 (approved):
lgtm

PR #100, opened 2020-11-04
Adds gender and age data to ACS ingestion

ldbnr4:
This change also allows the lint check to run on dispatch


mayaspivak (commented):
It looks like the ACS is part of the GCP public BQ datasets. Is there a specific reason for ingesting the data this way instead of using what is already in BQ?

ldbnr4:
> It looks like the ACS is part of the GCP public BQ datasets. Is there a specific reason for ingesting the data this way instead of using what is already in BQ?

I was working from the notes in go/msm-data that says we still wanted to ingest this data using the API. Is this no longer valid?

mayaspivak:
> > It looks like the ACS is part of the GCP public BQ datasets. Is there a specific reason for ingesting the data this way instead of using what is already in BQ?
> 
> I was working from the notes in go/msm-data that says we still wanted to ingest this data using the API. Is this no longer valid?

I just didn't have that context. This looks good to me. Thanks!

mayaspivak:
Sorry. I meant to close the comment. Not the PR. Trying to figure out how to reopen it!

mayaspivak (approved):

slagathorr (approved):

Issue #103, opened 2020-11-10
BQ Schema Structure Design


slagathorr:
Design a dataset schema that's standardized for the final environment.

slagathorr:
Some environments to keep in mind:

- Dataset for sources where schemas have a detected diff #102 which will (likely) be in a separate project from the data one, possibly the infra project.
- Dataset which has DB views that the export process will pull from.
- Dataset for "cleaned" raw data.
- Dataset for pre-approved manual uploads (also likely in the infra project, not data).

slagathorr:
@ldbnr4 going to start ironing this out, one of the things that we need to do is pick a column for the update date for each row, and apply that across all tables. My suggestion would be a `last_update_ts` and have that as the TIMESTAMP datatype and have that in the UTC timezome. Wanted to pick something now because I'm working on other metadata things that require it, so want to make sure we set it so there's no variance. Thoughts? Yes/No?

ldbnr4:
Do we need to apply this to every row? I would imagine `last_update_ts` will have the same value for every row right?

slagathorr:
Yeah, just adding that column to all tables. Shouldn't add much more data. We can't guarantee that every row is going to be the same because of the possibility of late arriving changes. Even for things that change once every year or two, it'll still be helpful to have just so you can get the latest "batch" easily.

ldbnr4:
As discussed offline, sgtm!

slagathorr:
So it's documented, for that particular field we will be standardizing to having `ingest_ts` across tables with the `TIMESTAMP` data type set to UTC.

Issue #30, opened 2020-10-02
Design for handling data sources changing and avoiding duplicate/redundant data


aaronsn:
Currently the code is set up to ingest the same data source even if it hasn't changed. This causes duplicate data in BigQuery. This isn't a big deal for now because we have an ingestion_time timestamp so it's easy to just query all rows where `ingestion_time = MAX(ingestion_time)` to avoid duplicates/old data. The upside is it makes ingestion very simple. The downside is it complicates the data and uses unnecessary storage and query costs (time & money).

We should figure out whether to change this strategy. Some things that would need to be thought about if we do:
- how do we detect if the data has changed to prevent ingesting duplicates?
- how do we preserve historical data?

I think avoiding ingesting redundant data in the first place could get complicated as it may vary by data source. One idea I had is to keep the ingestion as is, and then periodically materialize new tables with just the latest data. Then the old data can be cleaned up with a periodic task or automatically with a TTL.

Eg:
`ingest data to GCS bucket  =>  upload from bucket to BigQuery  =>  BigQuery_historical_table  =>  BigQuery_current_data`
BigQuery_historical_table would have duplicate/redundant data and periodically get cleaned up, but BigQuery_current_data would always be the most recent.

aaronsn:
@mayaspivak @slagathorr If I remember correctly, both of you mentioned having opinions on this

jenniebrown:
I think that if we're not going to use ingestion time partitioned tables then my preference would be to use checksums at the GCS level to decide whether or not to continue to upload data. You can compute the md5 or crc32 checksum on data before uploading and check against the checksum of existing data in GCS (if any) to decide whether to upload at that point.

slagathorr:
Ooh. I do like the checksum idea, but might suggest doing that inside of BQ as a scheduled query. If the datasets get larger, it might be faster/cheaper/simpler to just use Dremel's horsepower to perform creating checksums and deduplications in bulk as a single atomic operation. Also means you get the benefits of BQ managing the schedules, capacity, and also provides a built in 7 day window to recover data in case something went wrong.

mayaspivak:
We could do a combination of the two. I think Jenny's suggestion to do it in GCS is best for the datasets we know are updated rarely (e.g. once a year). Pulling those into BQ daily or weekly would be excessive. But we could do the checksum in BQ for datasets we know are updated more frequently.

aaronsn:
[...]

PR #106, opened 2020-11-12
Creates an Airflow pipeline for CDC covid deaths

ldbnr4:


airflow/dev/docker-compose-data-ingestion.yml
```diff
@@ -15,6 +21,12 @@ services:
             - webserver
         environment:
             - PORT=8080
+            # Ensure that the project set with PROJECT_ID has a BigQuery dataset named 'data_sources'
+            - DATASET_NAME=data_sources
```
ldbnr4, line 25: @slagathorr @mayaspivak this is related to #103 

Are we ok with this being the standard BQ schema across all projects?
<project_id>
- data_sources
-- <some_table_1>
-- <some_table_2>
--<...>
- <future_data_sources_dataset>
-- <more_tables_1>
-- <more_tables_2>
--<...>

airflow/dags/ingestion.py
```diff
@@ -13,19 +13,45 @@
...
+        resp = requests.post(url, json=data)
+        resp.raise_for_status()
+    except requests.exceptions.HTTPError as err:
+        raise Exception('Failed response code: {}'.format(err))
+
+
+# CDC Covid Deaths
+cdc_covid_deaths_payload = {"message": {"is_airflow_run": True,
+                                        "filename": "cdc_deaths",
+                                        "gcs_bucket": Variable.get("CDC_COVID_DEATH_BUCKET"),
```
jenniebrown, line 48: I have a slight preference for using the variable name of the bucket directly instead of creating a variable name for the bucket for each source since it makes fewer things to keep track of. This assuming that each source will be sharing the same bucket.

airflow/dev/.env
```diff
@@ -5,4 +5,6 @@ AIRFLOW_CONN_METADATA_DB=postgres+psycopg2://airflow:airflow@postgres:5432/airfl
 AIRFLOW_VAR__METADATA_DB_SCHEMA=airflow
 AIRFLOW__SCHEDULER__SCHEDULER_HEARTBEAT_SEC=10
 AIRFLOW_VAR_INGEST_TO_GCS_SERVICE_ENDPOINT=http://ingest_to_gcs:8080
-AIRFLOW_VAR_GCS_TO_BQ_SERVICE_ENDPOINT=http://gcs_to_bq:8080
\ No newline at end of file
+AIRFLOW_VAR_GCS_TO_BQ_SERVICE_ENDPOINT=http://gcs_to_bq:8080
+# Ensure that the you set CDC_COVID_DEATH_BUCKET with a Cloud Storage bucket name in your executing project
+AIRFLOW_VAR_CDC_COVID_DEATH_BUCKET=${CDC_COVID_DEATH_BUCKET}
```
jenniebrown, line 10: nit: please add a newline at the end of the file

airflow/dev/docker-compose-data-ingestion.yml
mayaspivak: That works. If we don't already, we should set up a dataset called data_sources in the terraform config.

airflow/dags/ingestion.py
```diff
@@ -13,19 +13,45 @@
...
+                                        "id": "CDC_COVID_DEATHS",
+                                        "url": "https://data.cdc.gov/api/views/k8wy-p9cg/rows.csv?accessType=DOWNLOAD"}}
+cdc_covid_deaths_gcs_operator = create_gcs_ingest_operator(
+    'cdc_covid_deaths_to_gcs', cdc_covid_deaths_payload)
+cdc_covid_deaths_bq_operator = create_bq_ingest_operator(
+    'cdc_covid_deaths_to_bq', cdc_covid_deaths_payload)
```
mayaspivak, line 54: I think the expected payloads are slightly different. The gcs to bq one requires a dataset name. There may be others.

mayaspivak (changes requested):

airflow/dev/docker-compose-data-ingestion.yml
jenniebrown, line 25: This name isn't compatible with the TEST environment, which is called 'test_data_sources'

mayaspivak: Whoops. sorry I forgot that I actually named it somewhere. The schema from the above comment is fine. We should just have the names here and in terraform align.

slagathorr: I'd say we should probably have the datasets correspond to each major source. CDC, CDC BRFSS, Census ACS, etc., all major groups (where it makes sense) ideally would. I'd avoid just having one "data" dataset that has everything as it tends ot make discoverability more difficult down the road. The PDP does a good job at showing what that looks like.

image

airflow/dags/ingestion.py
```diff
@@ -13,19 +13,45 @@
 data_ingestion_dag = DAG(
     'data_ingestion_dag',
     default_args=default_args,
+    schedule_interval="@weekly",  # Run once a week at midnight on Sunday morning
```
slagathorr, line 16: Double checked the update frequency (https://www.cdc.gov/nchs/covid19/covid-19-mortality-data-files.htm) and given that, I'd change this to run once at around 1PM Friday, and once around 6PM Wednesday to get as close as we can to their updates, unless I'm looking at the wrong data source (decent chance I could be, in which case, ignore this).

slagathorr: Doh, I was looking at the wrong one. How often does this dataset update, and on which days of the week/time?

airflow/dags/ingestion.py
ldbnr4: So you're thinking we should use a single bucket for all our data?

jenniebrown: Oh, I was under the impression that there would be one shared landing bucket for all the data sources since that was the model of the prototype. We don't necessarily need to do it like that, but just so I'm up to speed, could you explain the reasoning behind having a different landing bucket per data source?

airflow/dags/ingestion.py
mayaspivak: I did the ingestion for the prototype. If we run it daily we will need logic to avoid appending duplicate data. Doable but still needs to be implemented.

airflow/dev/docker-compose-data-ingestion.yml
ldbnr4: No objection to separating these out into per source datasets but we'll need to refactor our ingestion server to support this.

For now i'll use the dataset name we defined in terraform so that we can get something working end to end with what we have.

ldbnr4: Good catch. I'll use the dataset name we defined in terraform. 

FYI: From the thread above on the BQ structure, we'll probably need to refactor this and the terraform file.

airflow/dev/.env
ldbnr4: done

airflow/dags/ingestion.py
ldbnr4: Good point. The url field is only need to do gcs ingestion but the dataset name is pulled from an environment variable.

https://github.com/SatcherInstitute/health-equity-tracker/blob/master/run_gcs_to_bq/main.py#L63

airflow/dags/ingestion.py
ldbnr4: I see. I think I misinterpreted how we implemented the prototype. Until a need arises to separate them, I don't see a reason why not to keep them all in a single bucket.

airflow/dags/ingestion.py
ldbnr4: Sounds good. I'll leave this part commented out until we address #30.

airflow/dags/ingestion.py
ldbnr4: Updated this to use the same `gcs_landing_bucket` name as terraform.

mayaspivak (approved):

jenniebrown (approved):

airflow/dev/docker-compose-data-ingestion.yml
slagathorr: Yup in the interim that SGTM.

slagathorr (approved):

PR #107, opened 2020-11-16
Create new Cloud Run service to export tables from BQ to GCS

mayaspivak:
This exports all tables in the given dataset.

Next steps:
- Add the new service to the terraform config.
- Add the new service to the github deployToTest workflow.
- Make calls to the service from the airflow pipeline.

Potential future improvements:
- Run the table extraction jobs in parallel instead of sequentially.
- Only export table if the data has changed.


mayaspivak:
All tests pass when I run pytest from the exporter directory, but they fail with "405: method not allowed" when pytest is run from the main repo directory. I am still investigating.

mayaspivak:
> All tests pass when I run pytest from the exporter directory, but they fail with "405: method not allowed" when pytest is run from the main repo directory. I am still investigating.

And actually the gcsToBqTest also has three failing test cases when I run locally both in the main directory and under python/tests. 

I thought at first it might have to do with the Flask test client but that doesnt explain the gcsToBqTest failures.
Has anyone else run into this?

jenniebrown:
> And actually the gcsToBqTest also has three failing test cases when I run locally both in the main directory and under python/tests.
> 
> I thought at first it might have to do with the Flask test client but that doesnt explain the gcsToBqTest failures.
> Has anyone else run into this?

Have you installed all the test dependencies? It may be that your workspace doesn't have all the required packages. The requirements are in requirements/tests.txt

mayaspivak:
> > And actually the gcsToBqTest also has three failing test cases when I run locally both in the main directory and under python/tests.
> > I thought at first it might have to do with the Flask test client but that doesnt explain the gcsToBqTest failures.
> > Has anyone else run into this?
> 
> Have you installed all the test dependencies? It may be that your workspace doesn't have all the required packages. The requirements are in requirements/tests.txt

Yea I do have all the dependencies installed. 

It looks like pytest needs to be run from each cloud run service directory because there doesn't seem to be a way to distinguish between which main module to import the app from since the services are not in separate python packages. What was happening somehow is that test_exporter was importing app from data_sever/main.py instead of from exporter/main.py when pytest was run from the main repo directory. 

Alternatively we could name the main file for each service something different so that there are no conflicts. 

I think I prefer the first solution, since the services are built and packaged separately for production. So we could run pytest under python/ to get all the package tests and then under data_server/ and exporter/
Thoughts?

jenniebrown:
> > > And actually the gcsToBqTest also has three failing test cases when I run locally both in the main directory and under python/tests.
> > > I thought at first it might have to do with the Flask test client but that doesnt explain the gcsToBqTest failures.
> > > Has anyone else run into this?
> > 
> > 
> > Have you installed all the test dependencies? It may be that your workspace doesn't have all the required packages. The requirements are in requirements/tests.txt
> 
> Yea I do have all the dependencies installed.
> 
> It looks like pytest needs to be run from each cloud run service directory because there doesn't seem to be a way to distinguish between which main module to import the app from since the services are not in separate python packages. What was happening somehow is that test_exporter was importing app from data_sever/main.py instead of from exporter/main.py when pytest was run from the main repo directory.
> 
> Alternatively we could name the main file for each service something different so that there are no conflicts.
> 
> I think I prefer the first solution, since the services are built and packaged separately for production. So we could run pytest under python/ to get all the package tests and then under data_server/ and exporter/
> Thoughts?

Ah yep that's annoying. My only concern with doing it the first way is that it makes it trickier for devs to make sure all the tests are run and the github workflow slightly more brittle. But having rules about not naming any files main.py is also weird and easy to trip over so I'm fine with your approach.

mayaspivak:
> > > > And actually the gcsToBqTest also has three failing test cases when I run locally both in the main directory and under python/tests.
> > > > I thought at first it might have to do with the Flask test client but that doesnt explain the gcsToBqTest failures.
> > > > Has anyone else run into this?
> > > 
> > > 
> > > Have you installed all the test dependencies? It may be that your workspace doesn't have all the required packages. The requirements are in requirements/tests.txt
> > 
> > 
> > Yea I do have all the dependencies installed.
> > It looks like pytest needs to be run from each cloud run service directory because there doesn't seem to be a way to distinguish between which main module to import the app from since the services are not in separate python packages. What was happening somehow is that test_exporter was importing app from data_sever/main.py instead of from exporter/main.py when pytest was run from the main repo directory.
> > Alternatively we could name the main file for each service something different so that there are no conflicts.
> > I think I prefer the first solution, since the services are built and packaged separately for production. So we could run pytest under python/ to get all the package tests and then under data_server/ and exporter/
> > Thoughts?
> 
> Ah yep that's annoying. My only concern with doing it the first way is that it makes it trickier for devs to make sure all the tests are run and the github workflow slightly more brittle. But having rules about not naming any files main.py is also weird and easy to trip over so I'm fine with your approach.

Yea. I think in the future we could scan the subdirectories of the main repo directory for a dockerfile and run pytest in ones that have one, but for now I'll keep this in a list of things I need to do to add this new service. So that it will at least be documented.

jenniebrown:
> > > > > And actually the gcsToBqTest also has three failing test cases when I run locally both in the main directory and under python/tests.
> > > > > I thought at first it might have to do with the Flask test client but that doesnt explain the gcsToBqTest failures.
> > > > > Has anyone else run into this?
> > > > 
> > > > 
> > > > Have you installed all the test dependencies? It may be that your workspace doesn't have all the required packages. The requirements are in requirements/tests.txt
> > > 
> > > 
> > > Yea I do have all the dependencies installed.
> > > It looks like pytest needs to be run from each cloud run service directory because there doesn't seem to be a way to distinguish between which main module to import the app from since the services are not in separate python packages. What was happening somehow is that test_exporter was importing app from data_sever/main.py instead of from exporter/main.py when pytest was run from the main repo directory.
> > > Alternatively we could name the main file for each service something different so that there are no conflicts.
> > > I think I prefer the first solution, since the services are built and packaged separately for production. So we could run pytest under python/ to get all the package tests and then under data_server/ and exporter/
> > > Thoughts?
> > 
> > 
> > Ah yep that's annoying. My only concern with doing it the first way is that it makes it trickier for devs to make sure all the tests are run and the github workflow slightly more brittle. But having rules about not naming any files main.py is also weird and easy to trip over so I'm fine with your approach.
> 
> Yea. I think in the future we could scan the subdirectories of the main repo directory for a dockerfile and run pytest in ones that have one, but for now I'll keep this in a list of things I need to do to add this new service. So that it will at least be documented.

Can you edit pytest calls to the pytest github action so that they pass for now?

jenniebrown:
> Can you edit pytest calls to the pytest github action so that they pass for now?

Ah nevermind, I just saw your other PR

jenniebrown (approved):

PR #109, opened 2020-11-18
Allows the Cloud Run service to ingest manually uploaded data

ldbnr4:


config/test/test.tfvars
```diff
@@ -1,15 +1,16 @@
...
-notify_data_ingested_topic             = "notify-data-ingested-topic"
-ingestion_service_name                 = "data-ingestion-service"
-ingestion_image_name                   = "data-ingestion"
-ingestion_subscription_name            = "data-ingestion-sub"
-ingestion_invoker_identity_id          = "ingestion-invoker"
-ingestion_runner_identity_id           = "ingestion-runner"
-ingestion_runner_role_id               = "test_ingestion_runner"
-gcs_to_bq_service_name                 = "gcs-to-bq-service"
-gcs_to_bq_image_name                   = "gcs-to-bq"
+gcs_landing_bucket = "msm-test-landing-bucket"
```
slagathorr, line 1: Can we add something to configure/override the project ID as well? It's not immediately relevant since everything is going into the same project, but in the HET environment we would want the manual uploads to go into the INFRA project, whereas "approved" data will go into the data project, so being able to configure/override the project ID will help us then.

ldbnr4: We store the project ID in an environment variable `project_id` that can be overridden based on the deployment. We do do this for the test deployment workflow.
https://github.com/SatcherInstitute/health-equity-tracker/blob/master/.github/workflows/deployToTest.yml

You think this will work for the infra project/manual upload case?

slagathorr: Hm. Wondering if I'm reading this right.

The gotcha I think I'm seeing is that `project_id` is global for the entire deployment. This will include external facing data which needs to go into the `het-data` project, as well as manual uploaded data into the `het-infra` project, so it feels like everything soup to nuts goes into one project. Or are we able to specify the target project as a parameter into the GCR service that loads to BQ? That may be a parameter there I am not aware of.

image

ldbnr4: I see. How about we deploy our code in each project `het-data` and `het-infra`? 

We could configure the current code to write to a separate project but there will be some overhead to refactor and setup the appropriate permissions to allow this.

slagathorr: Hm. Is there no way to add a parameter to the BQ load service for a project ID? Duplicating the entire infrastructure seems a bit heavy since all the other project needs to hold are BQ datasets/GCS buckets, and nothing else.

ldbnr4: Yeah we can do that, let me update this. When we actually run the pipeline I believe the executing service account will need permission to write to both the infra and data projects but let me test it out

slagathorr: Yeah, that's true. If you need help figuring out IAM stuff, lmk.

ldbnr4: Thanks. Added the project override

slagathorr (approved):
lgtm

PR #112, opened 2020-11-23
Adds /dataset route to dataserver

jenniebrown:
This is a very basic implementation that simply requests the given dataset name from GCS on every invocation. Note that it doesn't use the cache because I have reservations about the viability of that implementation, though I think it is okay for metadata.

### DIFF
--- a/data_server/main.py
+++ b/data_server/main.py
@@ -1,10 +1,12 @@
 import logging
 import os
 
-from flask import Flask, Response
+from flask import Flask, Response, request
+import google.cloud.exceptions
 from werkzeug.datastructures import Headers
 
 from data_server.dataset_cache import DatasetCache
+import data_server.gcs_utils as gcs_utils
 
 app = Flask(__name__)
 cache = DatasetCache()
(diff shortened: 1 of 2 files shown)


data_server/main.py
```diff
@@ -35,5 +37,27 @@ def generate_response(data: bytes):
...
+        dataset = gcs_utils.download_blob_as_bytes(
+            os.environ.get('GCS_BUCKET'), dataset_name)
+    except google.cloud.exceptions.NotFound:
+        return 'Dataset {} not found'.format(dataset_name), 404
+
+    def generate_response(data: bytes):
+        for row in dataset.splitlines():
+            yield row + b'\n'
+    headers = Headers()
+    headers.add('Content-Disposition', 'attachment', filename=dataset_name)
```
aaronsn, line 57: You'll need to add
```
headers.add('Access-Control-Allow-Origin', '*')
headers.add('Vary', 'Origin ')
```
to allow the frontend to request the data

aaronsn (changes requested):
Nice! Glad to see this coming together :)

data_server/main.py
```diff
@@ -35,5 +37,27 @@ def generate_response(data: bytes):
...
+
+    try:
+        dataset = gcs_utils.download_blob_as_bytes(
+            os.environ.get('GCS_BUCKET'), dataset_name)
+    except google.cloud.exceptions.NotFound:
+        return 'Dataset {} not found'.format(dataset_name), 404
+
+    def generate_response(data: bytes):
+        for row in dataset.splitlines():
+            yield row + b'\n'
```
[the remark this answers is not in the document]
jenniebrown: That's fine, I can reformat it on the server if it's better not to do it on the front end. The files exported from BQ will be in newline-delimited json regardless.

data_server/main.py
jenniebrown: Done. Do you have a link that explains these headers? I'm not very familiar with them.

aaronsn: Yep, I'd start here: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

TL;DR is that because the data server and our frontend are on different origins, the frontend cannot read the contents of responses from the data server unless the data server provides the "Access-Control-Allow-Origin" header with a value that matches the frontend's origin. Passing the "*" value says that any origin can read it (which I think is fine per offline discussions about not using CORS as a privacy mechanism and that it's fine if the API is public). The "Vary" header specifies how intermediate caches (like CDNs, or I think even local browser http cache) uniquely identify a resource. So specifying "Origin" means that separate origins will not share a cache. This is important to make sure the access control is respected even from cache. I'm guessing flask doesn't add caching headers by default so the Vary header probably doesn't matter, but it's good practice to be safe.

Actually, I forgot - the Vary header value should be `Accept-Encoding, Origin`. The Accept-Encoding piece makes sure intermediate caches don't serve the wrong encoding. Here's a good explanation: https://blog.stackpath.com/accept-encoding-vary-important/

aaronsn (approved):

data_server/main.py
jenniebrown: Thanks! These are really helpful

PR #121, opened 2020-12-02
Adds a DAG to trigger manual ingestion

ldbnr4:


airflow/dags/manual_ingestion.py
```diff
@@ -0,0 +1,22 @@
...
+    'manual_ingestion_dag',
+    default_args=default_args,
+    description='Triggering for manual uploads.')
+
+# Manual Uplaods
+manual_uploads_payload = {'message': {'is_airflow_run': True,
+                                      'gcs_bucket': Variable.get('GCS_MANUAL_UPLOADS_BUCKET'),
+                                      'id': 'MANUAL_UPLOADS'}}
+manual_uploads_bq_operator = create_bq_ingest_operator(
+    'manual_uploads_task', manual_uploads_payload, manual_ingestion_dag)
```
mayaspivak, line 22: We will still need the exporter for manual uploads. Were you planning on adding this in a separate PR?

slagathorr: Ah so theoretically the BQ project that the exporter should pull from should be separate. The manual ingest stuff has an internal loop because we need human eyes on it, so it should be a somewhat separate loop.

https://github.com/SatcherInstitute/health-equity-tracker/pull/109#discussion_r529041940

So in the drawing there, this is geared towards the "bottom track" of manual loads, and the "manual landing" dataset would not be something the exporter is sourcing.

Lorenzo, I think this is what this script is for, but correct me if I'm mis-interpreting.

ldbnr4: Yeah my plan was to do that in a separate PR. I think Brian had some idea that would make this process different from the other sources.

slagathorr: Yeah, gotcha. The missing step is some sort of "yes, this dataset looks right, let's promote it to the public one" but we can do that in a separate issue/PR.

mayaspivak (approved):

ldbnr4:
Added a few missing pieces for the test deployment workflow.

mayaspivak (approved):

PR #122, opened 2020-12-04
Prefixes Airflow variables to be recognized by the webserver

ldbnr4:
As of Airflow 1.10.10, the naming convention requires the environment variable be prefixed with "AIRFLOW_VAR_" 

https://airflow.apache.org/docs/apache-airflow/stable/concepts.html#storing-variables-in-environment-variables


mayaspivak (approved):
Awesome! Thank you for figuring this out!!

PR #123, opened 2020-12-04
Give the default compute service account the run.invoker IAM role

mayaspivak:
This allows the Cloud Composer pipeline to make requests to the Cloud Run services.


.github/workflows/deployToTest.yml
```diff
@@ -69,6 +69,7 @@ jobs:
           terraform apply -auto-approve -var-file=test/test.tfvars \
           -var 'gcp_credentials=${{ secrets.TEST_DEPLOYER_SA_KEY }}' \
           -var 'project_id=${{ secrets.TEST_PROJECT_ID }}' \
```
ldbnr4, line 71: These values (project_id and project_number) are different?

ldbnr4 (approved):

.github/workflows/deployToTest.yml
mayaspivak: Yea the project id is a user readable string and the number is numerical digits represented as a string.

PR #127, opened 2020-12-08
Get token for compute service account to make request to cloud run services

mayaspivak:
This change also fixes how the exporter handles request input.

Additionally, remove the invoker role binding for the default compute service account. It turns out this isn't necessary because the service account already has the project editor role which inherits the invoker role.


ldbnr4 (approved):
lgtm!

PR #138, opened 2020-12-11
Create aggregator that runs stored routines for the given dataset.

mayaspivak:
The routines will be configured and stored through terraform.

This Cloud Run service is not yet configured through terraform either. That will happen in a subsequent PR.


aggregator/main.py
```diff
@@ -0,0 +1,47 @@
...
+    project_id = os.environ.get('PROJECT_ID')
+    dataset_id = "{}.{}".format(project_id, dataset_name)
+
+    logging.info("Running aggregation queries on dataset: %s", dataset_id)
+    # List and call routines for the dataset. (These have been uploaded to BQ 
+    # through terraform)
+    bq_client = bigquery.Client() 
+    routines = bq_client.list_routines(dataset_id)
+
+    for routine in routines:
```
ldbnr4, line 32: Instead of calling all the routines in a dataset that may or may not be an aggregator, what are your thoughts of prefixing aggregating routines with something like "agg_*" and only executing routine names that match that pattern?

jenniebrown: Do we expect to have other types of routines? Are the dedupe routines going to be attached to the same datasets?

jenniebrown (approved):
LGTM pending Lorenzo's comment. Thanks for adding tests also!

aggregator/main.py
mayaspivak: I had thought the dedupe routines were more generic and not tied to specific datasets. But I added a required AGG_ prefix to be safe.

ldbnr4: Thanks for updating this! From my understanding, BQ routines have to be tied to a specific dataset

ldbnr4 (approved):

PR #142, opened 2020-12-15
Use the 'latest' tag when deploying to cloud run through terraform

mayaspivak:
I originally used the digest in hopes that terraform would recognize when a container had not changed and thus not redeploy the service. However, terraform seems to redeploy the service anyway. This simplifies the configuration without changing behavior.

### DIFF
--- a/.github/actions/buildAndPush/action.yml
+++ b/.github/actions/buildAndPush/action.yml
@@ -7,17 +7,10 @@ inputs:
   image-path:
     description: 'Location to where image should be pushed'
     required: true
-outputs:
-  image-digest:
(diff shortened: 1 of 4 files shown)


jenniebrown (approved):
LGTM, but is the reason that Terraform was always redeploying because we always create a new image regardless of whether the code has changed?

mayaspivak:
> LGTM, but is the reason that Terraform was always redeploying because we always create a new image regardless of whether the code has changed?

I had thought that if the code didnt change the hash wouldnt change but I guess if the build time is used in the hash that wouldnt be true. As a future update we could try to condition the builds on what code paths change but it gets complicated with the shared python directory.

PR #143, opened 2020-12-15
Allows for ingestion of large files

ldbnr4:
This change also renames the ingestion_time column as discussed offline to fit the column type


airflow/dags/util.py
```diff
@@ -19,19 +20,22 @@ def create_exporter_operator(task_id: str, payload: dict, dag: DAG) -> PythonOpe
...
-    token_request_url = token_url + url
-    token_request_headers = {'Metadata-Flavor': 'Google'}
-
-    # Fetch the token for the default compute service account
-    token_response = requests.get(token_request_url, headers=token_request_headers)
-    jwt = token_response.content.decode("utf-8")
-
-    # Provide the token in the request to the receiving service
-    receiving_service_headers = {'Authorization': f'bearer {jwt}'}
+    receiving_service_headers = {}
```
mayaspivak, line 23: Thanks for improving this!

mayaspivak (approved):

python/datasources/data_source.py
```diff
@@ -42,27 +42,28 @@ def write_to_bq_table(self, dataset: str, gcs_bucket: str, filename: str, table_
         gcs_bucket: The name of the gcs bucket to read the data from
         filename: The name of the file in the gcs bucket to read from
         table_name: The name of the BigQuery table to write to"""
-        frame = gcs_to_bq_util.load_csv_as_dataframe(gcs_bucket, filename)
+        chunked_frame = gcs_to_bq_util.load_csv_as_dataframe(
```
mayaspivak, line 45: Actually wait. For sources that override this method, some of them may be larger than the chunk size added to the util file but wont iterate over the chunks to append them all to bq.

mayaspivak (changes requested):

python/datasources/data_source.py
ldbnr4: Oh wow good catch. Added an additional parameter so that data frames don't come back unexpectedly chunked.

mayaspivak (approved):

slagathorr (approved):

PR #156, opened 2020-12-17
Add configuration to enable ingestion of Covid Tracking Project data

jenniebrown:
Enables ingestion up to GCS stage. Standardization and writing to BQ to follow. 

- Adds CovidTrackingProject data source class
- Adds Terraform config for BQ dataset
- Adds Airflow DAG for Covid Tracking Project that triggers ingest_to_gcs, plus a small refactor


airflow/dags/covid_tracking_project.py
```diff
@@ -0,0 +1,28 @@
...
+data_ingestion_dag = DAG(
+    'covid_tracking_project_ingestion_dag',
+    default_args=default_args,
+    schedule_interval='@daily',  # Run once a day at midnight
+    description='Ingestion configuration for Covid Tracking Project')
+
+# Covid Tracking Project
+ctp_dataset_name = 'covid_tracking_project'
+ctp_gcs_payload = util.generate_gcs_payload(
+    'covid_tracking_project', 'COVID_TRACKING_PROJECT', _CTP_DOWNLOAD_URL)
```
mayaspivak, line 22: nit: make these strings constants like the file below.

airflow/dags/ingestion.py
```diff
@@ -19,21 +21,19 @@
...
-cdc_covid_deaths_gcs_payload['message']['url'] = 'https://data.cdc.gov/api/views/k8wy-p9cg/rows.csv?accessType=DOWNLOAD'
-cdc_covid_deaths_gcs_operator = create_gcs_ingest_operator(
+cdc_covid_deaths_gcs_payload = util.generate_gcs_payload(
+    'cdc_deaths', 'CDC_COVID_DEATHS', _CDC_COVID_DEATHS_DOWNLOAD_URL)
+cdc_covid_deaths_gcs_operator = util.create_gcs_ingest_operator(
     'cdc_covid_deaths_to_gcs', cdc_covid_deaths_gcs_payload, data_ingestion_dag)
-cdc_covid_deaths_bq_payload['message']['dataset'] = cdc_dataset_name
-cdc_covid_deaths_bq_operator = create_bq_ingest_operator(
+cdc_covid_deaths_bq_payload = util.generate_bq_payload(
+    'cdc_deaths', 'CDC_COVID_DEATHS', _CDC_DATASET_NAME)
```
mayaspivak, line 29: Make this a constant as well. 

Sorry I know this is the worst type of comment... "since you are already here...". So feel free to ignore if you want and I can change in a separate PR.

config/covid_tracking_project.tf
```diff
@@ -0,0 +1,7 @@
+# Resources and routines for Covid Tracking Project ingestion.
```
mayaspivak, line 1: I think the data sources should be under a separate directory since we will be adding more of them. This means they will need to be in a separate terraform module and then called from the main module (https://www.terraform.io/docs/configuration/blocks/modules/syntax.html#calling-a-child-module).

In terms of input variables, I was thinking dataset names (and in the future routine names) dont necessarily need to be variables because they are qualified by project so if you are deploying the tf config in a new project there will be no conflict. I feel like our variable list is getting a bit unwieldy. What are your thoughts on this?

mayaspivak (changes requested):

config/covid_tracking_project.tf
jenniebrown: I'm all for removing variables that we don't ever need to change. The only issue I can forsee is if someone wanted to have 2 deployments in their test project, which I have done to accomodate for running terraform both locally and through github actions (but that is very niche and hacky anyway). I'll get rid of the variables. We can always add them back later if necessary.

airflow/dags/ingestion.py
jenniebrown: np, done! :)

config/.terraform/modules/modules.json
```diff
@@ -0,0 +1 @@
+{"Modules":[{"Key":"","Source":"","Dir":"."},{"Key":"data_sources","Source":"./data_sources","Dir":"data_sources"}]}
```
mayaspivak, line 1: Nit: add line

mayaspivak (approved):

PR #157, opened 2020-12-17
Writes all manual uploads to BQ with string values

ldbnr4:


python/datasources/data_source.py
```diff
@@ -45,26 +45,33 @@ def write_to_bq_table(self, dataset: str, gcs_bucket: str, filename: str, table_
...
-                .replace(')', '_')
-                .replace('=', 'eq')
-            ), inplace=True)
+            self.clean_frame_column_names(chunk)
             gcs_to_bq_util.append_dataframe_to_bq(
                 chunk, dataset, table_name, project=project)
 
+    def clean_frame_column_names(self, frame):
+        """ Replaces unfitting BigQuery column
+        characters and makes them all coumn names
```
jenniebrown, line 55: nit: "makes all column names"

super nit: I think this comment could fit on 2 lines

python/datasources/manual_uploads.py
```diff
@@ -21,5 +21,9 @@ def write_to_bq(self, dataset, gcs_bucket, _):
         bucket_files = gcs_to_bq_util.list_bucket_files(gcs_bucket)
```
jenniebrown, line 21: Do the manual uploads go to a separate gcs bucket from the automated ones? And are the manual uploads all in the same bucket? If so, do we want to be processing all of them every time?

python/ingestion/gcs_to_bq_util.py
```diff
@@ -30,22 +33,62 @@ def append_dataframe_to_bq(
...
+       column_types: Optional dict of column name to BigQuery data type. If
+                     present, the column names must match the columns in the
+                     DataFrame. Otherwise, table schema is inferred.
+       col_modes: Optional dict of modes for each field. Possible values include
+                  NULLABLE, REQUIRED, and REPEATED. Must also specify
+                  column_types to specify col_modes."""
+    json_data = __convert_frame_to_json(frame, column_types)
+    for sub in json_data:
+        for key in sub:
+            sub[key] = str(sub[key])
```
jenniebrown, line 67: Can you use DataFrame.convert_dtypes(convert_string=True) before converting to json instead? Then it could be taken care of in the ManualUploads class instead. Another option could be DataFrame.applymap(str)

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.convert_dtypes.html

python/ingestion/gcs_to_bq_util.py
```diff
@@ -30,22 +33,62 @@ def append_dataframe_to_bq(
...
+
+    __append_dataframe_to_bq(frame, dataset, table_name,
+                             None, None, project, json_data)
```
jenniebrown, line 70: I think this should use the column_types and col_modes variables here instead of None

jenniebrown (changes requested):

python/ingestion/gcs_to_bq_util.py
ldbnr4: Yeahp, thanks.

python/ingestion/gcs_to_bq_util.py
ldbnr4: Nice find, I gave this a try but no luck. I've tried similar approaches but the only way that this works is to do the string casting after converting to json.

python/datasources/manual_uploads.py
ldbnr4: Yeah we separate all manual uploads into their own bucket but good point. We probably don't want to upload all the files at the same time every time.

I opened [an issue](#159 ) to separate these out so we can do one file at a time.

python/datasources/data_source.py
ldbnr4: Done

python/ingestion/gcs_to_bq_util.py
```diff
@@ -7,20 +7,23 @@
 from google.cloud import bigquery, storage
 
 
-def append_dataframe_to_bq(
-        frame, dataset, table_name, column_types=None, col_modes=None, project=None):
-    """Appends the provided DataFrame to the table specified by
-       `dataset.table_name`. Automatically adds an ingestion time column.
+def __convert_frame_to_json(frame, column_types):
```
jenniebrown, line 10: Sorry to get picky, but could you add a docstring that indicates this function modifies column_types?

python/ingestion/gcs_to_bq_util.py
jenniebrown: Bummer :/ thanks for trying it!

jenniebrown (approved):
looks good! just one minor request before you submit

python/ingestion/gcs_to_bq_util.py
ldbnr4: No problem. It was a bit odd for this function to add the ingestion timestamp. Moved this logic into a separate function and placed it where the method description gives details on what's going on here.

PR #158, opened 2020-12-22
Iterate over Airflow DAGs to import.

mayaspivak:
The workflow no longer needs to be updated for every new source/DAGs.


jenniebrown (changes requested):
Can you add a note to the README or the wiki that all DAGS from this directory will be uploaded and a warning to not add files that aren't dags to airflow/dags?

mayaspivak:
> Can you add a note to the README or the wiki that all DAGS from this directory will be uploaded and a warning to not add files that aren't dags to airflow/dags?

Done.

jenniebrown (approved):

PR #160, opened 2020-12-23
Put Cloud Run image digests back into terraform config

mayaspivak:
It turns out terraform relies on the image path changing to know when to redeploy a service.


jenniebrown (approved):
Thanks!

PR #161, opened 2020-12-23
Remove ACS population subclasses and run both state and county level from the same DataSource. This simplifies ingestion configuration

aaronsn:


python/datasources/acs_population.py
```diff
@@ -129,39 +129,29 @@ def update_col_types(frame):
...
+class ACSPopulationIngester():
     """American Community Survey population data in the United States from the
        US Census."""
 
-    def __init__(self, county_level):
-        # TODO pass this in from message data.
-        self.base_acs_url = "https://api.census.gov/data/2019/acs/acs5"
+    def __init__(self, county_level, base_acs_url):
+        self.base_acs_url = base_acs_url
         """The base ACS url to use for API calls."""
```
jenniebrown, line 138: I don't think it's common practice to write docstring-style comments underneath member variables. Could you move the comments to above their variables and use '#' style comments instead?

python/datasources/acs_population.py
```diff
@@ -360,21 +348,32 @@ def get_sex_by_age_and_race(self, var_map, sex_by_age_frames):
         return result
 
 
-class ACSStatePopulation(ACSPopulationBase):
-    def __init__(self):
-        super().__init__(False)
+# TODO pass this in from message data.
+BASE_ACS_URL = "https://api.census.gov/data/2019/acs/acs5"
```
jenniebrown, line 352: nit: could you move this to the top of the file?

jenniebrown (changes requested):
Is there ever a situation where we might want to ingest one and not the other?

python/datasources/acs_population.py
aaronsn: Huh I thought I read this somewhere. Oh well, updated to # style

aaronsn:
> Is there ever a situation where we might want to ingest one and not the other?

In theory, maybe, but I can't think of any reason it would be that important to do so. ACS releases state/county data together so we can do them on the same schedule. I realized this introduces additional annoyances with the airflow configuration and I don't think it buys us that much.

jenniebrown (approved):

PR #164, opened 2020-12-23
Fixes bug where duplicate attributes are being passed through

aaronsn:


mayaspivak (approved):
For my understanding, can you explain what was happening?

aaronsn:
> For my understanding, can you explain what was happening?

Yeah so the dictionary that was passed in contained the values that were extracted from it, which had name conflicts with the explicitly passed params. I.e.:
```
attrs = {"gcs_bucket": "bucket", "extra_key": "whatever"}
gcs_bucket = attrs["gcs_bucket"]
# This fails, because **attrs includes gcs_bucket but gcs_bucket is already provided
upload_to_gcs(gcs_bucket, **attrs)
```

Using dict.pop() removes and returns the value, so it won't be included in the additional attributes

PR #174, opened 2020-12-29
Add Cloud Run service that serves the frontend, and add terraform config to deploy it

aaronsn:
- Also adds basic authentication

### DIFF
--- a/frontend_server/package.json
+++ b/frontend_server/package.json
@@ -1,8 +1,7 @@
 {
-  "name": "docker_web_app",
+  "name": "frontend_server",
   "version": "1.0.0",
-  "description": "Node.js on Docker",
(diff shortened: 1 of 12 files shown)


frontend_server/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "docker_web_app",
+  "version": "1.0.0",
+  "description": "Node.js on Docker",
+  "author": "First Last <first.last@example.com>",
```
kkatzen, line 5: delete/replace this one?

kkatzen (approved):

frontend_server/package.json
aaronsn: Good call! I also updated to make it private, matching how our frontend package.json looks

PR #175, opened 2020-12-30
Move card keys into components

kkatzen:
Fixes https://github.com/SatcherInstitute/het-frontend/issues/179


aaronsn (approved):
Thanks for doing this! Might be worth explaining why we're using keys for these components in the comments, I think it's not particularly obvious

PR #177, opened 2020-12-30
Adding Storybook

kkatzen:
Adding Storybook library, setting up stories for MapCard and ChoroplethMap (which is renamed)

Also migrating some styles to MaterialTheme so they are rendered through React and shown in Storybook


frontend/.storybook/main.js
```diff
@@ -0,0 +1,11 @@
...
+  "stories": [
+    "../src/**/*.stories.mdx",
+    "../src/**/*.stories.@(js|jsx|ts|tsx)"
+  ],
+  "addons": [
+    "@storybook/addon-links",
+    "@storybook/addon-essentials",
+    "@storybook/preset-create-react-app"
+  ]
+}
```
aaronsn, line 11: nit: newline at end of file

frontend/.storybook/preview.js
```diff
@@ -0,0 +1,4 @@
+
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+}
```
aaronsn, line 4: nit: newline at end of file

frontend/.storybook/preview.js
```diff
@@ -0,0 +1,4 @@
+
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
```
aaronsn, line 3: For my understanding, what's going on here? Maybe add a comment?

frontend/src/storybook/StoryWrapper.tsx
```diff
@@ -0,0 +1,29 @@
...
+} from "../data/useDatasetStore";
+import Logger from "../utils/Logger";
+import { initGlobals } from "../utils/globals";
+import DataFetcher from "../data/DataFetcher";
+import CssBaseline from "@material-ui/core/CssBaseline";
+import { ThemeProvider } from "@material-ui/styles";
+import MaterialTheme from "../styles/MaterialTheme";
+
+export const StoryWrapper = (storyFn: any) => {
+  initGlobals(new Logger(false), new DataFetcher());
```
aaronsn, line 16: Is storybook only intended for local development or will it be deployed somewhere? If the former, add a TODO to make sure these dependencies are set up so they won't make real API calls.

frontend/src/storybook/StoryWrapper.tsx
```diff
@@ -0,0 +1,29 @@
...
+  startMetadataLoad();
```
aaronsn, line 17: I'd move the initGlobals and startMetadataLoad calls outside of the component so they aren't repeatedly called. (I should probably update the initGlobals method to throw an exception if it's called multiple times actually...)

frontend/src/App.tsx
```diff
@@ -135,6 +136,7 @@ function App() {
   return (
     <ThemeProvider theme={MaterialTheme}>
       <ReactTooltip />
+      <CssBaseline />
```
aaronsn, line 139: what's going on with the CSS changes in this PR? Is it related to storybook?

aaronsn (approved):
This is awesome, thanks for getting this set up!

frontend/.storybook/preview.js
kkatzen: This is actually part of the boiler plate but we're not using it so I just deleted the file

frontend/src/storybook/StoryWrapper.tsx
kkatzen: oh good call!

frontend/src/App.tsx
kkatzen: yes- some of the styles were happening outside of the React framework so Storybook wasn't picking up on them and so the components looked different on storybook than the website. Since almost all of our styles were already within the React framework it seemed best to move all of them inside the framework to reduce complication and ensure the storybook renderings were the same as the actual site

aaronsn: Ooh gotcha, good call!

PR #183, opened 2021-01-05
Refactor Variable Provider to not throw error

kkatzen:
Resolving https://github.com/SatcherInstitute/het-frontend/issues/181


frontend/src/data/MetricQuery.ts
```diff
@@ -48,6 +48,10 @@ export class MetricQueryResponse {
       this.error = input as Error;
       this.data = [];
       this.invalidValues = {};
+    } else if (input.length <= 0) {
```
aaronsn, line 51: This comment isn't anchored properly because github doesn't allow me to comment on the right line...

Should we rename `shouldShowError` to `shouldShowDataWarning`?

frontend/src/data/useDatasetStore.test.tsx
```diff
@@ -1,50 +1,92 @@
...
+    return (
+      <WithMetrics queries={[props.query]}>
+        {() => {
+          const response = datasetStore.getMetrics(props.query);
+          return (
+            <div data-testid="MetricQueryResponseReturned">
+              {response.isError() && <>Error: {response.error!.message}</>}
+              {!response.isError() && (
+                <>
+                  Loaded {response.data!.length} rows.{" "}
```
aaronsn, line 75: Data is non-nullable so this `!` shouldn't be necessary

aaronsn (approved):

frontend/src/data/MetricQuery.ts
```diff
@@ -48,6 +48,10 @@ export class MetricQueryResponse {
...
+      this.error = new ExpectedError("No rows returned");
+      this.data = [];
+      this.invalidValues = {};
```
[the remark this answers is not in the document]
kkatzen: I'm not sure if failure is the best term. To me, it implies success was potentially possible. But the situations that would result in this lack of data are: 
1. unsupported breakdown 2. dataset rows are empty (for example, user requests a state for which the dataset doesn't have data)
as opposed to situations situations that throw instead of returning MetricQueryResponse:
1. invalid dataset id 2. server failures

Maybe "unavailable"or "invalid" or "bad request" would be more accurate than "error" or "failure"?

I've changed "isError" to "dataIsUnavailable" and the "showErrorMessage" to "shouldShowMissingDataMessage" since the actual text of the alert we're showing in this situation does say "data is missing".

What do you think?

I'm also not sure we need isFailure as a constructor input... couldn't this be derived based on absence/presence of failureMessage/dataUnavailableMessage ?

Also, do you have a specific opposition to having the input just be either message or the data?

frontend/src/data/MetricQuery.ts
kkatzen: I've got it as shouldShowMissingDataMessage now to match the message content but we can do shouldShowDataWarning if you prefer.

aaronsn: What you have works too!

frontend/src/data/MetricQuery.ts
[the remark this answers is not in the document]
aaronsn: Your reasoning makes sense. I'm not specifically attached to "failure" terminology, I just don't want to use "error" and failure was the best alternative I could think of because I thought others were potentially too specific. But data unavailable works too, I think it covers all the use cases.

re: input as (Row[] | string) and deriving failure based on presence of the message, I can see pros/cons to both approaches. The only downsides to doing it the way you suggested are that it doesn't scale as well if you need more granular/structured information about the failure, and it's a little less explicit with what's going on. But it works for the current use case, and it's not hard to change later if we find ourselves needing more structured information. I'd suggest adding a comment though indicating that when the data provided is a string, it is treated as an error. I think it's kind of subtle. I also still think a factory method that only accepts a string and is named clearly, like `createMissingDataResponse(message: string)` helps improve readability.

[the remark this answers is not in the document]
kkatzen: i added createMissingDataResponse and made all names refer to "missing data" per slack chat

PR #190, opened 2021-01-08
Add ACS aggregations and hook up to airflow

aaronsn:
Maya - is there a better place I can put the aggregation SQL file? Seems like it doesn't belong in the config folder, but I'm not sure where it should go and how to make sure the terraform config can find the file if it's in a parent or sibling directory

### DIFF
--- a/airflow/dags/acs_population.py
+++ b/airflow/dags/acs_population.py
@@ -28,9 +28,13 @@ acs_pop_bq_payload = util.generate_bq_payload(
 acs_pop_bq_operator = util.create_bq_ingest_operator(
     'acs_population_to_bq', acs_pop_bq_payload, data_ingestion_dag)
 
+acs_pop_aggregator_payload = {'dataset_name': _ACS_DATASET_NAME}
+acs_pop_aggregator_operator = util.create_aggregator_operator(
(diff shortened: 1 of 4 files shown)


mayaspivak:
> Maya - is there a better place I can put the aggregation SQL file? Seems like it doesn't belong in the config folder, but I'm not sure where it should go and how to make sure the terraform config can find the file if it's in a parent or sibling directory

Yea. My concern about moving it out is that we then depend on the relative paths of config and whatever new sql dir remaining the same. We could have a separate sql dir or one per source if thats more useful for organization. Either or leaving as is seems fine to me.

mayaspivak (approved):

aaronsn:
> > Maya - is there a better place I can put the aggregation SQL file? Seems like it doesn't belong in the config folder, but I'm not sure where it should go and how to make sure the terraform config can find the file if it's in a parent or sibling directory
> 
> Yea. My concern about moving it out is that we then depend on the relative paths of config and whatever new sql dir remaining the same. We could have a separate sql dir or one per source if thats more useful for organization. Either or leaving as is seems fine to me.

I'll leave for now, we can re-evaluate if this gets out of hand

PR #191, opened 2021-01-08
Adds short circuiting pipeline logic after the GCS download step

ldbnr4:


airflow/dags/util.py
```diff
@@ -100,14 +102,34 @@ def service_request(url: str, data: dict):
     try:
         resp = requests.post(url, json=data, headers=receiving_service_headers)
         resp.raise_for_status()
+        kwargs['ti'].xcom_push(key='response_status', value=resp.status_code)
```
mayaspivak, line 105: Add comment that you are saving the response status for checking later.

What will happen if a step runs twice in a row? Will it save both values?

airflow/dags/util.py
```diff
@@ -100,14 +102,34 @@ def service_request(url: str, data: dict):
...
         task_id=task_id,
+        provide_context=provide_context,
         python_callable=service_request,
         op_kwargs={'url': url, 'data': payload},
+        xcom_push=xcom_push,
+        dag=dag,
+    )
+
+
+def did_gcs_file_download(gcs_download_task_id: str, **kwargs):
```
mayaspivak, line 122: Add comment.

mayaspivak (approved):

airflow/dags/util.py
ldbnr4: In that scenario I believe the only the latest value will be saved for retrieval. If this becomes a problem I think we can use a combination of xcom keys and task ids to differentiate multiple calls.

https://airflow.apache.org/docs/apache-airflow/stable/concepts.html?highlight=xcom#xcoms

airflow/dags/util.py
ldbnr4: done

PR #258, opened 2021-01-11
Add flask_cors integration to data server

### LINKED ISSUE #238: Data server doesn't add CORS headers to all requests
The metadata request is missing CORS headers. Also, if the dataset isn't found it's missing CORS headers.

The data server is simple enough that we could definitely manually add the headers for every response, but using a Flask extension like https://flask-cors.readthedocs.io/en/latest/ is probably the better way to do it so it is guaranteed to happen for all requests.

jenniebrown:
Resolves #238


data_server/test_data_server.py
```diff
@@ -71,6 +71,7 @@ def testGetMetadata(mock_func: mock.MagicMock, client: FlaskClient):
     assert response.status_code == 200
     assert (response.headers.get('Content-Disposition') ==
             'attachment; filename=test_data.ndjson')
+    assert response.headers.get('Access-Control-Allow-Origin') == '*'
```
aaronsn, line 74: Add asserts that the Vary header is present too

aaronsn (approved):
Thanks for the quick fix!

data_server/test_data_server.py
jenniebrown: This doesn't add the Vary header, but according to w3 the Vary header is only necessary if the CORS requirements are more complicated than setting `Access-Control-Allow-Origin` to *. flask_cors follows the w3 implementation. Is there another reason why Vary: Origin is required here? 

https://fetch.spec.whatwg.org/#cors-protocol-and-http-caches

aaronsn: I think the "Accept-Encoding" part is actually the more important piece. Though I think that only matters for very old browsers so it probably doesn't matter, I just think it's better to be safe. While unnecessary, it's certainly not harmful, and makes it future-proof if we decide to restrict access to specific origins.

I'd say if the Flask CORS extension you added makes it easy to add then do it, otherwise probably fine to drop

aaronsn (approved):

data_server/test_data_server.py
jenniebrown: I don't think there's a way to force it to add the Vary header for this configuration. The w3 docs actually specify not to use Vary: Origin with "Access-Control-Allow-Origin" to *. flask_cors already has the setting CORS_VARY_HEADER True by default. If we end up restricting access to specific origins, then it will inject it automatically.

I'm not sure what you mean by "Accept-Encoding" here.

aaronsn: Oh sorry you're right, Accept-Encoding has to be added separately. That's good to know about the cors extension automatically adding Vary if you restrict it to specific origins

PR #259, opened 2021-01-11
Use service account creds to make e2e test requests

mayaspivak:


jenniebrown (commented):
LGTM, but were we going to send the e2e requests to the frontend server instead of the data server?

mayaspivak:
> LGTM, but were we going to send the e2e requests to the frontend server instead of the data server?

Couldn't get that working right away. I want to get the prod deployment and documentation out. Making a request with creds will be needed for calling the pipeline anyway so this will be useful even if we do switch. 

If you are ok with it, I will talk to Aaron about how to call the frontend correctly and make the change separately.

jenniebrown (approved):

jenniebrown:
Sounds good, thanks for looking into it

PR #260, opened 2021-01-12
Set up frontend to use environment variables for different contexts

aaronsn:


frontend/src/data/DataFetcher.ts
```diff
@@ -54,29 +66,18 @@ class DataFetcher {
...
-      process.env.NODE_ENV === "test" ||
-      (process.env.NODE_ENV === "development" && !this.apiUrl);
+  baseApiUrl: string;
+
+  constructor(baseApiUrl: string, deployContext: DeployContext) {
+    this.baseApiUrl = baseApiUrl;
+
+    // Use the static file directory for development environments unless the API
+    // url is provided
+    this.forceStaticFile = deployContext === "development" && !this.baseApiUrl;
```
kkatzen, line 76: I've noticed Storybook is actually kinda slow (DatasetStore isn't shared between stories) so I'm inclined to have Storybook also read from static file. We could add a TODO to follow up

kkatzen (approved):

frontend/src/data/DataFetcher.ts
aaronsn: Oh interesting. I think the tricky part about this is in the long-term we don't want to be keeping any static data files in github, which is the only way to deploy these. I don't think there's a way around that unfortunately.

PR #265, opened 2021-01-13
Set up frontend environments

aaronsn:
- Create environment files for each deploy context and configure them appropriately
- Update deployToTest to pass deploy context, which is used to determine which environment file to run
- Update deployToTest to read basic auth username and password from github secrets and pass to docker build. These are the only environment variables that need to be passed in this way since they are read from github secrets. Other variables can just be added to the appropriate environment files
- Add README instructions for running the frontend in various ways


.github/actions/buildAndPush/action.yml
```diff
@@ -17,7 +26,16 @@ runs:
     - name: Set Up Docker to Use gcloud Credentials
       run: gcloud auth configure-docker -q
       shell: bash
-    - run: docker build -t ${{ inputs.image-path }} -f ${{ inputs.dockerfile }} .
+    # Note: it is not generally recommended to pass secrets as build args, but
```
mayaspivak, line 29: For the prod workflow, instead of building containers, they are copied to the prod project's repo. 

That won't work with this change. 

I'd like to keep copying at least the backend containers with the idea that the containers from the test environment have been verified and tested already. 

Should I change the prod workflow to build the frontend container with the new build arguments instead of just copying from test?

mayaspivak (approved):

.github/actions/buildAndPush/action.yml
aaronsn: Hmm. An alternative is to provide these as runtime environment variables. That would work fine for the username/password (and arguably it's better to do it that way because it avoid this not-recommended way of passing sensitive data to a container).

However, the build-context argument is needed at build-time to determine which environment file to use for the frontend. I don't think there's a good way around that unfortunately. We could in theory build support for runtime environment variables in the frontend but I don't think that's a good idea because it introduces a bunch of complexities like concerns around XSS, caching, etc.

I guess I'd say the frontend should just be built separately, unless you have another idea?

mayaspivak: I don't know enough about the js container runtime to have any other suggestions. Building the frontend for prod should work fine. Just want to make sure it's good with you.

aaronsn: Yeah it's fine with me. The frontend container is simple enough, we'll just need to make sure that prod releases are validated immediately after pushing. But I think we'd need to do that anyway since environment variables would change the behavior regardless of whether they're passed at build time or runtime.

mayaspivak: Yep, the e2e tests will still be run right after the prod resources are updated.

README.md
```diff
@@ -237,13 +240,48 @@ If you encounter errors during install that mention `gyp`, that refers to a Node
 
 #### Develop
 
-To run the app in development mode, start a local web server, and watch for changes do:
+Since the frontend is a static site that just connects to an API for data requests, most frontend development happens independently of server-side changes. The simplest way to do this is to connect the frontend to the test website server. First, copy `frontend/.env.example` into `frontend/.env.development`. This file is already set up to point to the test website server.
```
kkatzen, line 243: What if we start with definitions of the larger elements and how they are related:
* React App
* Cloud server hosting of the React App
* Data server
* Storybook

Then something like "We use configuration files (frontend/.env.prod, frontend/.env.staging, etc)" to control settings in different environments. These include things like the data server URL and logging settings. These can be overridden."

(Maybe this is in a ###Overview section between ###Install and ###Develop)

And then have the sections be of the format:
1. Running just the React App
2. Running the Frontend Server locally
3. Running the Frontend Server with Docker locally
4. Running frontend in your own GCP project
5. Running Storybook (moving it from below up here)

And for each section
-> When/why you would want to do this
-> How to do it (commands)
-> Default behavior, and and any available overrides

What do you think?

kkatzen (approved):

README.md
aaronsn: Did my best to match that structure, with some minor changes. Let me know what you think and we can iterate going forward!

PR #271, opened 2021-01-14
Additional test for Variable Providers

kkatzen:
Adding national tests to ACS
Adding CovidProvider test
Adding BrfssProvider test


aaronsn (approved):
Thank you for adding tests!

PR #277, opened 2021-01-15
Add build directory optional input

mayaspivak:
This allows this action to be used from the cloud platform setup repository.

### DIFF
--- a/.github/actions/buildAndPush/action.yml
+++ b/.github/actions/buildAndPush/action.yml
@@ -7,6 +7,10 @@ inputs:
   image-path:
     description: 'Location to where image should be pushed'
     required: true
+  build-directory:
+    description: 'Directory from where docker build should be run'
+    required: false
+    default: '.'
   deploy-context:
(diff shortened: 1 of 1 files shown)


aaronsn (approved):
For my understanding, why is this required to run from the setup repository?

mayaspivak:
> For my understanding, why is this required to run from the setup repository?

The setup repo checks out both its own code and the het repo code into two different subdirectories. The action, before this change, assumes you are already in the het repo subdirectory. Specifying a working-directory isn't allowed in workflow steps that use actions, so it needs to be an input instead.

aaronsn:
> > For my understanding, why is this required to run from the setup repository?
> 
> The setup repo checks out both its own code and the het repo code into two different subdirectories. The action, before this change, assumes you are already in the het repo subdirectory. Specifying a working-directory isn't allowed in workflow steps that use actions, so it needs to be an input instead.

Ah makes sense, thanks!

Issue #255, opened 2021-01-11
Add "total" sum as option on MetricQuery


kkatzen:

kkatzen:
To date, we only have race data and "Total" is included as its own race. Other breakdowns will not have "Total". The total is the same across breakdowns.

Do we use the "Total" from race data? Do we add "Total" to all breakdowns? Do we create another way to get "Total" separate from breakdowns as they are not related.

Currently, if we're only looking at age data, we will not query for the race dataset and I do not think we should.

kkatzen:
Update MetricQuery to take a boolean indicating if response should include a "Total" as a value for the specified breakdown field (race_and_ethnicity, age, gender)

MapCard will just repeat the "Total" in each of the breakdown options in the Filter dropdown. This should be less weird when we have subset menus for each breakdown

PR #280, opened 2021-01-15
Add "include total" option in Breakdowns

kkatzen:
#255


frontend/src/data/Breakdowns.ts
```diff
@@ -53,16 +53,21 @@ export class Breakdowns {
 
   filterFips?: string;
 
+  // Whether or not a we should include a "Total" value in response which sums total population
```
aaronsn, line 56: Replace "sums total population" with "sums the total value for the metric across the breakdown"

frontend/src/data/variables/CovidProvider.ts
```diff
@@ -77,6 +77,7 @@ class CovidProvider extends VariableProvider {
     // TODO How to handle territories?
     const acsBreakdowns = breakdowns.copy();
     acsBreakdowns.time = false;
+    acsBreakdowns.includeTotal = true;
```
aaronsn, line 80: Use the setter here?

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -46,6 +46,23 @@ class BrfssProvider extends VariableProvider {
       });
     }
 
+    if (breakdowns.includeTotal && breakdowns.demographic) {
+      const breakdownName: string =
+        breakdowns.demographic === "race"
+          ? "race_and_ethnicity"
+          : breakdowns.demographic;
```
aaronsn, line 53: Use `demographicBreakdownToCol` from the `Breakdowns` file?

frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -61,6 +61,30 @@ class AcsPopulationProvider extends VariableProvider {
...
+    ) {
+      df = df.where((row) => row["race_and_ethnicity"] !== "Total");
+    }
+
+    // Age and sex data comes from backend without Total, if it's wanted, we add it.
+    if (
+      breakdowns.includeTotal &&
+      (breakdowns.demographic === "age" || breakdowns.demographic === "sex")
+    ) {
+      df = df.concat(
```
aaronsn, line 79: resetIndex after concat to be safe since the indices overlap when you concat frames

frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -61,6 +61,30 @@ class AcsPopulationProvider extends VariableProvider {
...
+    // Race data comes from backend with Total, if it's not wanted, we remove it.
+    if (
+      !breakdowns.includeTotal &&
+      (breakdowns.demographic === "race_nonstandard" ||
+        breakdowns.demographic === "race")
+    ) {
+      df = df.where((row) => row["race_and_ethnicity"] !== "Total");
+    }
+
+    // Age and sex data comes from backend without Total, if it's wanted, we add it.
```
aaronsn, line 74: I think it might be better to move this above line 51, so that you don't have to special-case the population_pct logic anymore. Also makes it easier to drop when the server gives us Totals. Ie something like this:

```
if (breakdowns.age || breakdowns.sex) {
  // Add "Total" for those
}

df = applyToGroups((group) => {
  // Calculate population_pct, using "Total" regardless of breakdown
});

if (!breakdowns.includeTotal) {
  df = df.where(...) // filter out "Total"
}
```

frontend/src/cards/TableCard.tsx
```diff
@@ -55,10 +54,7 @@ export function TableCard(props: TableCardProps) {
       {() => {
         const queryResponse = datasetStore.getMetrics(query);
         const dataset = queryResponse.data.filter(
-          (row) =>
-            !["Not Hispanic or Latino", "Total"].includes(
-              row.race_and_ethnicity
-            )
+          (row) => "Not Hispanic or Latino" !== row.race_and_ethnicity
```
aaronsn, line 57: nit: reverse order of these

aaronsn (changes requested):

kkatzen (commented):
big changes! let me know what you think!

frontend/src/data/Breakdowns.ts
```diff
@@ -24,81 +24,80 @@ export const BREAKDOWN_VAR_DISPLAY_NAMES: Record<BreakdownVar, string> = {
...
 export class Breakdowns {
   geography: GeographicBreakdown;
   // We may want to extend this to an explicit type to support variants for
   // day/week/month/year.
   time: boolean;
-  race: boolean;
-  race_nonstandard: boolean;
-  age: boolean;
-  sex: boolean;
+  demographicBreakdowns: Record<string, DemographicBreakdown>;
```
aaronsn, line 60: Given the number of places were we directly access "race"/"age"/etc on this, could we make this a structured type? I think you can do that while still making it a map that is iterable:
```
type DemographicBreakdowns = {
  race: DemographicBreakdown;
  age: DemographicBreakdown;
  ...
}
```

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -46,39 +48,37 @@ class BrfssProvider extends VariableProvider {
...
+
+    return new MetricQueryResponse(df.toArray(), ["brfss"]);
   }
 
   allowsBreakdowns(breakdowns: Breakdowns): boolean {
     const validDemographicBreakdownRequest =
       breakdowns.demographicBreakdownCount() === 0 ||
-      (breakdowns.demographicBreakdownCount() === 1 && breakdowns.race);
+      (breakdowns.demographicBreakdownCount() === 1 &&
+        breakdowns.demographicBreakdowns.race.enabled);
```
aaronsn, line 81: optional improvement: I feel like many of these checks would be more readable if we had convenience methods like:
```
breakdowns.hasExactlyOneDemographic()
breakdowns.hasOnlyRace()
breakdowns.hasOnlyAge()
...
```

aaronsn (approved):
Nice! This does a good job of generalizing things and reducing hard-coding

frontend/src/data/Breakdowns.ts
kkatzen: I added a type as the key. We could always change this later

kkatzen: I made a type for the key... make the whole thing a structured type was introducing a lot of typescript complaints... we can always revisit

PR #284, opened 2021-01-16
Set DATA_SERVER_URL env var in terraform config

mayaspivak:


config/run.tf
```diff
@@ -159,7 +159,12 @@ resource "google_cloud_run_service" "frontend_service" {
   template {
     spec {
       containers {
-        image = format("gcr.io/%s/%s@%s", var.project_id, var.frontend_image_name, var.frontend_image_digest)
+        image = format("gcr.io/%s/%s@%s", var.project_id, var.frontend_image_name, var.frontend_image_digest)   
+        env {
+          # URL of the Data Server Cloud Run service.
+          name = DATA_SERVICE_URL
```
aaronsn, line 165: This should be DATA_SERVER_URL

aaronsn (approved):

config/run.tf
mayaspivak: Fixed. Thank you!

Issue #283, opened 2021-01-15
Handle multiple data sources per metric


aaronsn:
Scenario: a single metric comes from multiple data sources. For example, covid cases at the state level come from CTP, covid cases at the county level come from the CDC. Right now, that will result in the card footer showing both of the sources, when only one of them is being used for the current data. This is confusing because the user has no way of telling which data source was actually used for the data they're looking at.

To solve this, we will add a `getDependentDatasets(query: MetricQuery)` method on the VariableProviders that specifies which datasets are required to serve a particular query, and change `variableProviders.getDependentDatasets()` to use that rather than referencing the static list of datasets.

kkatzen:
A reminder regarding the footer:
For MapCard, we request all demographic breakdowns if the "all" toggle (race/age/sex/all) is selected which results in multiple ACS datasets listed. This issue does not address that. #264 should add a static map of consolidate unique dataset IDs (acs_race, acs_age) that point to the same data source (ACS)

kkatzen:
This is no longer needed for the MVP so we'll de-prioritize. It is something we may want in the future, however.

benhammondmusic:
@joshzarrabi was this addressed in your recent vaccine update? Or do we still need to fix the cited sources?

PR #285, opened 2021-01-19
Get dataset IDs based on MetricQuery, not MetricId

kkatzen:
#283 
Any given Variable Provider may provide different metrics using different  datasets.
Previously, we assumed all dataset IDs in the Variable Provider were used for any given metric it provided, this PR refactors so that we can determine exactly which datasets were used based on the MetricQuery.

- VariableProviders have a new method that takes a MetricQuery and returns which dataset IDs are needed to calculate the metric.
- A helper function which previously got dataset based on metricId now takes MetricQuery
- Instead of taking dataset IDs directly, CardWrapper uses the helper function above, passing in its queries prop.


frontend/src/data/variableProviders.ts
```diff
@@ -67,16 +68,21 @@ export function getProvider(variableId: MetricId): VariableProvider {
...
   const dedupedIds = Array.from(new Set(providerIds));
   return dedupedIds.map((id) => providersById[id]);
 }
 
-/** Returns the list of dataset ids that the provided variables depend on. */
-export function getDependentDatasets(variableIds: MetricId[]): string[] {
-  const providers = variableIds.map((id) => getProvider(id));
-  return VariableProvider.getUniqueDatasetIds(providers);
+/** Returns the list of dataset ids that the provided MetricQuerys depend on. */
+export function getDependentDatasets(queries: MetricQuery[]): string[] {
```
aaronsn, line 78: I wonder if this should be named the same as the VariableProvider method (getRequiredDatasetIds)?

frontend/src/data/variableProviders.ts
```diff
@@ -67,16 +68,21 @@ export function getProvider(variableId: MetricId): VariableProvider {
...
+  let datasetIds = new Set<string>();
+  queries.forEach((query) =>
+    query.metricIds.forEach((id) =>
+      getProvider(id)
+        .getRequiredDatasetIds(query)
+        .forEach((id) => datasetIds.add(id))
+    )
+  );
+  return Array.from(datasetIds);
```
aaronsn, line 87: nit: I prefer a slightly more functional style for this kind of logic since it's more concise, similar to the way `getUniqueDatasetIds` works. But up to you, I think this is more of a style preference

frontend/src/data/variables/CovidProvider.ts
```diff
@@ -36,6 +36,19 @@ class CovidProvider extends VariableProvider {
     this.acsProvider = acsProvider;
   }
 
+  // TODO- perhaps add checks to verify that MetricQuery is supported at all
```
aaronsn, line 39: Could you do this in the superclass method using `allowsBreakdowns`? I think this would be nice to do in this PR because otherwise it will be very easy for it to go out of date as we add more datasets

frontend/src/data/variables/CovidProvider.ts
```diff
@@ -36,6 +36,19 @@ class CovidProvider extends VariableProvider {
...
+  getRequiredDatasetIds(metricQuery: MetricQuery) {
+    if (
+      metricQuery.breakdowns.demographic === "race_nonstandard" ||
+      metricQuery.breakdowns.demographic === "race"
+    ) {
+      return ["acs_population-by_race_state_std", "covid_by_state_and_race"];
+    } else if (metricQuery.breakdowns.demographic === "age") {
+      return ["acs_population-by_age_state", "covid_by_state_and_race"];
```
aaronsn, line 47: Covid doesn't support age right now

aaronsn (approved):
Nice! This is a good improvement

frontend/src/data/variableProviders.ts
kkatzen: dicsussed on slack

frontend/src/data/variables/CovidProvider.ts
kkatzen: outdated

frontend/src/App.module.scss
```diff
@@ -16,9 +16,10 @@
   button {
     color: $white;
   }
-  &:link, &:visited {
-  color: $white;
-  text-decoration: none;
+  &:link,
```
kkatzen, line 19: think my automatic linter may have done something here when I merged... just whitespace changes

frontend/src/App.tsx
```diff
@@ -91,11 +91,9 @@ function AppToolbar() {
       </Typography>
       {[EXPLORE_DATA_PAGE_LINK, DATA_CATALOG_PAGE_LINK, ABOUT_US_PAGE_LINK].map(
         (pageUrl, i) => (
-            <LinkWithStickyParams to={pageUrl} class={styles.NavLink}>
-              <Button key={i}>
-                {PAGE_URL_TO_NAMES[pageUrl]}
-              </Button>
-            </LinkWithStickyParams>
+          <LinkWithStickyParams to={pageUrl} class={styles.NavLink}>
```
kkatzen, line 94: think my automatic linter may have done something here when I merged... just whitespace changes

frontend/src/cards/BarChartCard.tsx
```diff
@@ -58,17 +58,21 @@ function BarChartCardWithKey(props: BarChartCardProps) {
...
-
   // TODO - what if there are no valid types at all? What do we show?
   const validDisplayMetricConfigs: MetricConfig[] = Object.values(
     props.variableConfig.metrics
   ).filter((metricConfig) => VALID_METRIC_TYPES.includes(metricConfig.type));
 
+  const metricIds = Object.values(props.variableConfig.metrics).map(
+    (metricConfig: MetricConfig) => metricConfig.metricId
+  );
+  let metrics: MetricId[] = [...metricIds];
```
aaronsn, line 69: Why do you need to copy the metric ids here?

frontend/src/cards/CardWrapper.tsx
```diff
@@ -73,15 +73,24 @@ function CardWrapper(props: {
       }
     >
       {() => {
+        const queryResponses = queries.map((query) =>
+          datasetStore.getMetrics(query)
+        );
+        const consumedDataset = queryResponses.reduce(
```
aaronsn, line 79: consumedDatasetIds?

aaronsn (approved):
Very nice!

Issue #282, opened 2021-01-15
Support multiple demographic breakdowns in frontend


kkatzen:
In Breakdowns, migrate "demographic?: DemographicBreakdown" to be 3 boolean fields, one for each potential breakdown. Update providers accordingly, currently we shouldn't have any datasets supporting multiple breakdowns so they can all just err

PR #286, opened 2021-01-19
Breakdowns can support multiple demographics

kkatzen:
#282 
Previously, Breakdowns only supported requests for data broken down by no more than one of age, sex, race.
This PR splits a one-of field into distinct fields.
This PR should not change existing functionality. All VariableProviders should still only support one breakdown for now but can be refactored later to support multiple.


frontend/src/data/Breakdowns.ts
```diff
@@ -44,23 +44,29 @@ function demographicBreakdownToCol(
 
 export class Breakdowns {
   geography: GeographicBreakdown;
-  // Note: this assumes only one demographic breakdown at a time. If we want to
-  // support more later we can refactor this to multiple boolean fields.
-  demographic?: DemographicBreakdown;
```
aaronsn, line 49: Is the DemographicBreakdown type still used anywhere or can we delete it?

frontend/src/data/Breakdowns.ts
```diff
@@ -149,6 +161,18 @@ export class Breakdowns {
...
+  demographicBreakdownCount() {
+    let demographicBreakdownCount = 0;
+    [this.age, this.sex, this.race, this.race_nonstandard].forEach(
+      (demographic) => {
+        if (demographic) {
+          demographicBreakdownCount++;
+        }
+      }
+    );
+    return demographicBreakdownCount;
```
aaronsn, line 173: you can change this whole block to:
`[this.age, this.sex, this.race, this.race_nonstandard].filter((demo) => !!demo).length`

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -63,7 +63,8 @@ class BrfssProvider extends VariableProvider {
       !breakdowns.time &&
       (breakdowns.geography === "state" ||
         breakdowns.geography === "national") &&
-      (!breakdowns.demographic || breakdowns.demographic === "race")
+      breakdowns.demographicBreakdownCount() === 1 &&
+      !!breakdowns.race
```
aaronsn, line 67: Previously this allowed no demographic breakdown - don't you want both `0` and `1` to be valid counts?
```
breakdowns.demographicBreakdownCount() === 0 || (breakdowns.demographicBreakdownCount() === 1 && !!breakdowns.race)
```

frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -39,6 +39,18 @@ class AcsPopulationProvider extends VariableProvider {
...
+      !!metricQuery.breakdowns.race_nonstandard ||
+      !!metricQuery.breakdowns.race
+    ) {
+      return ["acs_population-by_race_state_std"];
+    } else if (!!metricQuery.breakdowns.age) {
+      return ["acs_population-by_age_state"];
+    }
+    return [];
+  }
+
```
aaronsn, line 53: I think this snuck in from a different PR?

frontend/src/data/Breakdowns.ts
```diff
@@ -44,23 +44,29 @@ function demographicBreakdownToCol(
...
-  // support more later we can refactor this to multiple boolean fields.
-  demographic?: DemographicBreakdown;
   // We may want to extend this to an explicit type to support variants for
   // day/week/month/year.
   time?: boolean;
-
+  race?: boolean;
+  race_nonstandard?: boolean;
+  age?: boolean;
+  sex?: boolean;
```
aaronsn, line 53: What do you think about making these booleans non-optional and then defaulting them to false in the constructor? Avoids the need to add `!!` everywhere they're used

aaronsn (approved):

frontend/src/data/Breakdowns.ts
kkatzen: It was used in one other place but I don't think it was really needed- just removed!

frontend/src/data/Breakdowns.ts
kkatzen: done- I added a new method on Breakdowns to keep the string in unsupported breakdown error the same - I think it's easier to read when we don't list which are set to false.

Issue #116, opened 2020-11-30
Covid Tracking Project Race Data Ingestion and Standardization


aaronsn:
This only covers the Covid Tracking Project race data (https://covidtracking.com/race), not their general data

jenniebrown:
This issue is also tracking standardization work for CTP data.

jenniebrown:
Closing this since ingestion/standardization steps are complete. I'll file separate issues for cleanup items.

PR #288, opened 2021-01-19
Add standardization/write to BQ step for CTP

jenniebrown:
- Implements write_to_bq for the CTP data source
- Adds two standardized race categories, API and INDIGENOUS
- Adds unit tests

#116


python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,85 @@ def get_id():
     def get_table_name():
         return 'covid_tracking_project'
 
+    @staticmethod
+    def get_standard_columns():
```
mayaspivak, line 21: Add method comment here and to other helper methods below.

mayaspivak (approved):

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+        df = df.melt(id_vars=['date', 'state'])
+        df[['variable_type', 'race_and_ethnicity']] = df.variable.str.split(
+            "_", 1, expand=True)
+        df.drop('variable', axis=1, inplace=True)
+        df.rename(columns={'state': 'state_postal_abbreviation'}, inplace=True)
+        df.replace({'race_and_ethnicity': self.get_standard_columns()},
+                   inplace=True)
+
+        # Get the metadata table
+        metadata = self._download_metadata(metadata_table_id)
```
aaronsn, line 59: Per a comment in some google doc thread, don't we need a `reports_ethnicity` column too? I'm forgetting the exact details, but I think we discussed needing that information in order to know how to join with ACS?

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+        # Rename combined race categories
+        self._rename_race_category(merged, 'reports_api', col_std.Race.ASIAN,
+                                   col_std.Race.API)
+        self._rename_race_category(merged, 'reports_ind', col_std.Race.AIAN,
+                                   col_std.Race.INDIGENOUS)
+
+        merged.drop(columns=['reports_api', 'reports_ind'], inplace=True)
+
+        # Write to BQ
+        gcs_to_bq_util.append_dataframe_to_bq(merged, dataset, table_name)
```
aaronsn, line 78: Some general data cleanup stuff:
- Could we convert the dates to `yyyy-mm-dd` format?
- Not sure if CTP fixed this issue, but in the past when I was running a script on their data, some of the int values were actually strings with commas in them, so I had to [fix those](https://github.com/aaronsn/health-equity-tracker/blob/566860cf99fe1fd3be65ca930db1c677d9efcdbf/python/datasources/covid_tracking_project.py#L103) before uploading to BQ

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+        """Returns a dict containing conversions from Covid Tracking Project's
+           race categories to their standardized values."""
+        return {
+            'aian': col_std.Race.AIAN.value,
+            'asian': col_std.Race.ASIAN.value,
+            'black': col_std.Race.BLACK.value,
+            'nhpi': col_std.Race.NHPI.value,
+            'white': col_std.Race.WHITE.value,
+            'multiracial': col_std.Race.MULTI.value,
+            'other': col_std.Race.OTHER.value,
```
aaronsn, line 31: I think maybe we should split OTHER into two (not sure on the exact naming):
1. The same "Other" that matches up with ACS "Other"
2. A different "Other" that can't be joined with ACS

If a dataset uses the exact same race categories as ACS then we can join the "other" group, otherwise we can't. Even if we mostly just use the non-joinable one, I think being explicit would be better

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+
+        # Massage the data into the standard format.
+        df.drop(columns=['cases_latinx', 'deaths_latinx',
+                         'hosp_latinx', 'tests_latinx'], inplace=True)
+        df = df.melt(id_vars=['date', 'state'])
+        df[['variable_type', 'race_and_ethnicity']] = df.variable.str.split(
+            "_", 1, expand=True)
+        df.drop('variable', axis=1, inplace=True)
+        df.rename(columns={'state': 'state_postal_abbreviation'}, inplace=True)
+        df.replace({'race_and_ethnicity': self.get_standard_columns()},
```
aaronsn, line 55: Use the constant from the standardized columns file?

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
     def get_table_name():
         return 'covid_tracking_project'
 
+    @staticmethod
+    def get_standard_columns():
```
aaronsn, line 21: Are you still planning on splitting `race_and_ethnicity` into two columns (`race` and `includes_hispanic`)? I wasn't sure if that was intended as a followup or part of this

aaronsn (changes requested):
I really like how you're doing the metadata joining. We should be able to reuse this pattern to get fips codes into tables during the standardization step in the future

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+
+        # Merge the tables
+        merged = pd.merge(
+            df, metadata, how='left',
+            left_on=['state_postal_abbreviation', 'variable_type'],
+            right_on=['state_postal_abbreviation', 'variable_type'])
+        # Rename combined race categories
+        self._rename_race_category(merged, 'reports_api', col_std.Race.ASIAN,
+                                   col_std.Race.API)
+        self._rename_race_category(merged, 'reports_ind', col_std.Race.AIAN,
```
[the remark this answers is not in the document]
jenniebrown: There's only one state that does this right now, and they do report under AIAN. If other states start to do this (I think it's unlikely) then we'll have to double check.

python/ingestion/gcs_to_bq_util.py
```diff
@@ -173,6 +173,10 @@ def load_csv_as_dataframe(gcs_bucket, filename, dtype=None, chunksize=None):
     blob.download_to_filename(local_path)
     frame = pandas.read_csv(local_path, dtype=dtype, chunksize=chunksize)
 
+    # Warning: os.remove() will remove the directory entry but will not release
+    # the file's storage until the file is no longer being used by |frame|.
+    # Double warning: This will cause an exception on Windows. See
+    # https://docs.python.org/3/library/os.html#os.remove for details.
```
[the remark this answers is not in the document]
jenniebrown: The warning about windows is just for dev environments, but the first comment flags a potential race condition/memory concern for all environments. I added this comment a while ago since the call to os.remove() has behavior that could be unexpected (I originally thought that the memory referenced by the resulting dataframe could be corrupted when the file was removed, but that is not the case in Unix). I'm not sure when this comment was removed but it seems like something that is being added back in due to a desync in my workspace. I'd like to keep it in if that's all right. 

I personally think that this is kind of a code smell and would prefer to fix it, but I think it might require a bit of a refactor.

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+
+        df = gcs_to_bq_util.load_csv_as_dataframe(gcs_bucket, filename)
+        self.clean_frame_column_names(df)
+
+        # Massage the data into the standard format.
+        df.drop(columns=['cases_latinx', 'deaths_latinx',
+                         'hosp_latinx', 'tests_latinx'], inplace=True)
+        df = df.melt(id_vars=['date', 'state'])
+        df[['variable_type', 'race_and_ethnicity']] = df.variable.str.split(
+            "_", 1, expand=True)
```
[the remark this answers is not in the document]
jenniebrown: :)

python/datasources/covid_tracking_project.py
jenniebrown: Yes, that information is included in the metadata table.

python/datasources/covid_tracking_project.py
jenniebrown: Yes, when the metadata is joined the includes_hispanic column is added. I used the column name race_and_ethnicity still because it is a standard across multiple data sets, and it will still include ethnicity alone as its own "race category" (Hispanic and Non Hispanic), if I understood correctly.

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+                   inplace=True)
+
+        # Get the metadata table
+        metadata = self._download_metadata(metadata_table_id)
+        if len(metadata.index) == 0:
+            raise RuntimeError(
+                'BigQuery call to {} returned 0 rows'.format(metadata_table_id))
+
+        # Merge the tables
+        merged = pd.merge(
```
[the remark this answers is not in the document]
jenniebrown: No reason, they are equivalent

python/datasources/covid_tracking_project.py
jenniebrown: Yeah, I'll format the dates.

I haven't been able to repro the ints as strings issue you describe. The CTP data is a CSV, so the data is untyped and pandas has been importing them as ints

python/datasources/covid_tracking_project.py
```diff
@@ -14,5 +17,112 @@ def get_id():
...
+            df, metadata, how='left',
+            left_on=['state_postal_abbreviation', 'variable_type'],
+            right_on=['state_postal_abbreviation', 'variable_type'])
+        # Rename combined race categories
+        self._rename_race_category(merged, 'reports_api', col_std.Race.ASIAN,
+                                   col_std.Race.API)
+        self._rename_race_category(merged, 'reports_ind', col_std.Race.AIAN,
+                                   col_std.Race.INDIGENOUS)
+
+        merged.drop(columns=['reports_api', 'reports_ind'], inplace=True)
```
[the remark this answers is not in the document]
jenniebrown: Yeah, this is the big one that I wanted to get your input on. Technically any of the metadata columns can vary between cases/deaths/hospitalizations/tests, and several of them do (e.g. race_ethnicity_combined, reports_ethnicity, reports_race, uses_api...) This will result in a proliferation of columns if we use a pivoted table. My reasoning for using cases/deaths/etc as a breakdown column instead was that it makes it easy to look up the data you want without needing to do string parsing on column names. It also makes joining with the ACS data more straightforward.

I haven't worked on the frontend at all, but based on Krista's talk it seems like we would need to add a new Breakdown class? I'm not sure how much work that would be but in my head this format seems easier to work with. If it's not feasible, I think that creating multiple tables might be slightly preferable

python/datasources/covid_tracking_project.py
jenniebrown: Oooh okay I figured out the strings issue. I found if you set thousands=',' in the call to pd.read_csv, it will parse the numbers correctly. Woo!!

python/ingestion/gcs_to_bq_util.py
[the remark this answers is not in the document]
aaronsn: Oh yeah definitely fine to keep the comment, I just meant it would be nice to add something like "This shouldn't affect production environments but may be an issue for local dev on Windows"

python/datasources/covid_tracking_project.py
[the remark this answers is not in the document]
aaronsn: I agree with the part about not having separate columns because joining with ACS gets complicated and I don't want to do string-parsing on column names.

A 4th alternative I just thought of is using nulls and add additional rows where they differ:
```
reports_combined | cases | deaths
true             | 400   | null
false            | null  | 20
```
This would allow joining with ACS smoothly, and we just filter out nulls first. Thoughts on that?

@kkatzen what do you think? To clarify, the format Jennie is suggesting here is:
```
state | race_and_ethnicity | variable | number
NY    | white              | cases    | 50
NY    | white              | deaths   | 50
NY    | white              | hosp     | 50
```
This would require some massaging on the frontend. I don't think it'd be terribly complicated to do that in the variable providers. But I'm hesitant to add more custom logic on the frontend than already exists.

My preference is this 4th alternative using additional rows, and my second preference is separate files. What do y'all think?

python/datasources/covid_tracking_project.py
aaronsn: wooooah. That's pretty neat, nice find! Of course pandas can do this automagically

python/datasources/covid_tracking_project.py
aaronsn: Sure, fine to not block on this. If there's a possibility you don't get to this before the end of the fellowship, do you mind filing a github issue in case you don't finish?

I think the "defines_other" doesn't necessarily give all the information we need to know? Even if "other" is defined, it still needs to be defined in the exact same way as the ACS "other" category, which may not always be the case?

python/datasources/covid_tracking_project.py
aaronsn: Oh, I thought the way you were suggesting was that Hispanic/Non-hispanic would be part of the "race" column? I may have forgotten the details here...

aaronsn (approved):
Approving to unblock you - up to you how to chunk the followups we discussed as different PRs

python/datasources/covid_tracking_project.py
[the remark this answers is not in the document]
jenniebrown: As discussed offline, we will go for option 2, splitting into multiple tables. I opened issue #308

python/datasources/covid_tracking_project.py
jenniebrown: Created #309

python/datasources/covid_tracking_project.py
jenniebrown: Renamed to "race"

jenniebrown:
Thanks for the reviews! I created issues for the items we discussed

Issue #215, opened 2020-11-10
Bug: Fix nav bar tabs


kkatzen:
Clicking on button but not the text doesn't change the tab

aaronsn:
Per offline discussion: I think swapping the link to be outside the button may fix this: https://github.com/SatcherInstitute/health-equity-tracker/blob/e4a1e35737134e1f8bcde1409fd23419dad3fd3e/frontend/src/App.tsx#L93

The buttons here work as intended: https://github.com/SatcherInstitute/health-equity-tracker/blob/dcf8468e9e4f024953f6739f94a3293b7cca2163/frontend/src/pages/Landing/LandingPage.tsx#L35

PR #290, opened 2021-01-19
Fixing nav bar tabs

efregoso:
Making the button itself clickable rather than just link, per Issue #215

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -91,11 +91,11 @@ function AppToolbar() {
       </Typography>
       {[EXPLORE_DATA_PAGE_LINK, DATA_CATALOG_PAGE_LINK, ABOUT_US_PAGE_LINK].map(
         (pageUrl, i) => (
-          <Button className={styles.NavButton} key={i}>
-            <LinkWithStickyParams to={pageUrl}>
-              {PAGE_URL_TO_NAMES[pageUrl]}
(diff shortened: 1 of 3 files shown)


frontend/src/App.tsx
```diff
@@ -91,11 +91,11 @@ function AppToolbar() {
       </Typography>
       {[EXPLORE_DATA_PAGE_LINK, DATA_CATALOG_PAGE_LINK, ABOUT_US_PAGE_LINK].map(
         (pageUrl, i) => (
-          <Button className={styles.NavButton} key={i}>
-            <LinkWithStickyParams to={pageUrl}>
-              {PAGE_URL_TO_NAMES[pageUrl]}
+            <LinkWithStickyParams to={pageUrl} class={styles.NavLink}>
```
kkatzen, line 94: What if you put the NavLink class on the Button directly? then you wouldn't have to add the prop to LinkWithStickyParams or nest the CSS

efregoso: Unfortunately, this did not seem to work.  :(  I haven't been able to sleuth it out, but I tried adding the NavLink class to the button and altering CSS with a few variations on:
```.NavLink {
  color: $white;
  &:visited, &:link {
    color: white;
    text-decoration: none;
  }
}```
I've pushed a more condensed version of the previous CSS just to clean up the code a little bit regardless

kkatzen (approved):

PR #293, opened 2021-01-19
Refactor useDatasetStore into hooks that use local load state, and a global cache outside the React lifecycle 

aaronsn:
This fixes a bug where resources were loaded multiple times, and prepares for more complex caching scenarios. It also makes it so the UI correctness is not dependent on the details of the cache.


frontend/src/data/DataManager.ts
```diff
@@ -0,0 +1,256 @@
...
+import { DataFrame, IDataFrame } from "data-forge";
+import { MetricQuery, MetricQueryResponse } from "./MetricQuery";
+import { getDataFetcher, getDataManager, getLogger } from "../utils/globals";
+import { useEffect } from "react";
+
+async function onMetadataLoaded(callback: (metadata: MetadataMap) => void) {
+  try {
+    const metadata = await getDataManager().loadMetadata();
+    callback(metadata);
+    // Swallow errors - they are logged in the DataManager
```
kkatzen, line 14: isn't this the DataManager? not sure if you want to be more specific

kkatzen (approved):
ooo the performance gains on the storybook site are great :D :D 

Are you thinking you'll try to do these TODOs for 20%? If not it may be worth talking this over with the next cohort

frontend/src/data/DataManager.ts
aaronsn: Moved this to its own file, it was kinda awkward that it was here anyway since this is a React hook but the rest of this file is a data loader/cache that isn't React-specific

aaronsn:
> ooo the performance gains on the storybook site are great :D :D
> 
> Are you thinking you'll try to do these TODOs for 20%? If not it may be worth talking this over with the next cohort

I'm planning on doing the caching/performance/data loading related ones. Not sure which specific ones you're referring to, let me know if there's any in particular you think I should make sure to do.

PR #295, opened 2021-01-20
Update airflow image in docker compose

vanshkumar:
Updating airflow image in docker-compose file as the more recent version of airflow prompts login during local dev


ldbnr4 (approved):
Thanks for updating this!

Issue #279, opened 2021-01-15
Handle County level breakdowns


kkatzen:
In each VariableProvider, where we currently match breakdowns.filterFips for equivalency to state_fips we should be checking if the filterFips is a state or county and geographic breakdown.

If filterFips == state and geographic breakdown == state, do what currently is in code (equivalent on state_name)
If filterFips == state and geographic breakdown == county, do prefix match on county_name
If filterFips == county and geographic breakdown == county, do equivalent on county_name
If filterFips == county and geographic breakdown == state, throw invalid breakdown error

After this, let's keep whichever of county_fips and state_fips match geographic breakdown and rename it to just fips

Examples of what we would want returned for some sample Breakdowns:

Breakdowns.byState().forFips(NC_FIPS)
-> one row, NC
Breakdowns.byCounty().forFips(NC_FIPS)
-> one row per county for all NC counties
Breakdowns.byCounty().forFips(DURHAM_COUNTY)
-> one row, durham county

PR #310, opened 2021-01-25
Support County level breakdowns in AcsProvider

kkatzen:
Starting support for county level for #279 

Will follow up with renaming of the state_fips and county_fips fields


frontend/src/data/variables/CovidProvider.ts
```diff
@@ -71,8 +71,9 @@ class CovidProvider extends VariableProvider {
             })
             .resetIndex();
 
-    if (breakdowns.filterFips) {
-      df = df.where((row) => row.state_fips === breakdowns.filterFips);
+    if (breakdowns.filterFips !== undefined) {
+      const fips = breakdowns.filterFips as Fips;
```
aaronsn, line 75: Does this need the cast? Usually typescript can figure out from the if() check

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -52,6 +53,7 @@ class BrfssProvider extends VariableProvider {
       breakdowns.demographicBreakdowns.race.enabled &&
       breakdowns.demographicBreakdowns.race.includeTotal
     ) {
+      // TODO totals
```
aaronsn, line 56: What's the TODO here for specifically?

frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -144,15 +173,19 @@ class AcsPopulationProvider extends VariableProvider {
   }
 
   allowsBreakdowns(breakdowns: Breakdowns): boolean {
-    const validDemographicBreakdownRequest: boolean =
+    const validGeographicBreakdown =
+      breakdowns.geography === "county"
+        ? breakdowns.demographicBreakdowns.race_nonstandard.enabled ||
+          breakdowns.demographicBreakdowns.race.enabled
```
aaronsn, line 179: Use helper methods `hasOnlyRaceNonStandard` and `hasOnlyRace`?

frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -35,17 +35,49 @@ class AcsPopulationProvider extends VariableProvider {
...
+    }
+    if (
+      breakdowns.demographicBreakdowns.race_nonstandard.enabled ||
+      breakdowns.demographicBreakdowns.race.enabled
+    ) {
+      return breakdowns.geography === "county"
+        ? "acs_population-by_race_county_std"
+        : "acs_population-by_race_state_std";
+    }
+    return "";
```
aaronsn, line 58: I'd probably throw an error here - otherwise it'll just throw an error anyway when you try to use an invalid id, but the error will be clearer here

frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -73,7 +105,7 @@ class AcsPopulationProvider extends VariableProvider {
     const enabledBreakdown = Object.values(
       breakdowns.demographicBreakdowns
     ).find((breakdown) => breakdown.enabled === true)!;
-    df = applyToGroups(df, ["state_name"], (group) => {
+    df = applyToGroups(df, [geoNameColumn], (group) => {
```
aaronsn, line 108: Hmm, I think this should be fipsColumn? I think it was done this way because early datasets might not have fips columns but I think they all do now

frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -35,17 +35,49 @@ class AcsPopulationProvider extends VariableProvider {
...
+      breakdowns.geography === "county"
+        ? ["county_fips", "county_name"]
+        : ["state_fips", "state_name"];
+
+    // If requested, filter geography by state or county level
+    if (breakdowns.filterFips !== undefined) {
+      const fips = breakdowns.filterFips as Fips;
+      if (fips.isState() && breakdowns.geography === "county") {
+        df = df.where(
+          (row) => row["county_fips"].substring(0, 2) === fips.code
```
aaronsn, line 76: I'd prefer to keep the substring logic on the Fips class. One way to do that is to create a Fips instance to do the comparison, but in order to avoid creating so many objects I think a better way is to add a method `isParentOf(fipsCode: string)`

aaronsn (approved):
Nice! This is great

frontend/public/tmp/acs_population-by_race_county_std.json
```diff
@@ -0,0 +1,52 @@
+[{"county_fips":"37037","county_name":"Chatam","race_and_ethnicity":"American Indian and Alaska Native","population":2576},
```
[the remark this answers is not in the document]
kkatzen: Didn't realize it was available! oops!

frontend/src/data/Breakdowns.ts
```diff
@@ -114,6 +114,11 @@ export class Breakdowns {
       Object.assign({}, this.demographicBreakdowns),
       this.time,
       this.filterFips
+        ? Object.assign(
+            Object.create(Object.getPrototypeOf(this.filterFips)),
+            this.filterFips
+          )
```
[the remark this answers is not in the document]
kkatzen: o good call

frontend/src/data/variables/AcsPopulationProvider.ts
kkatzen: o nice idea

frontend/src/data/variables/AcsPopulationProvider.ts
kkatzen: oh ya! little merge artifact...

frontend/src/data/variables/BrfssProvider.ts
kkatzen: oh oops I think this was not meant to stay- deleted it

frontend/src/data/variables/CovidProvider.ts
kkatzen: it is actually throwing typescript errors without casting. :/ 
it seems like the first if statement is okay but inside the if/else it's throwing typescript errors :(

Issue #311, opened 2021-01-25
Create library for VariableProvider use


kkatzen:
We could extract some of the reusable aspects of the VariableProviders into a util/library for reuse

kkatzen:
these could be extracted as functions on the VariableProvider interface

PR #312, opened 2021-01-26
CovidProvider supports county with small sample fake data

kkatzen:
Refactoring CovidProvider to support a few counties in NC with fake data to identify and remove points of hard coding. #279 

Also starts in on #311 by moving some common functions onto VariableProvider


frontend/src/data/variables/VariableProvider.ts
```diff
@@ -42,6 +44,56 @@ abstract class VariableProvider {
...
   }
 
+  filterByGeo(df: IDataFrame, breakdowns: Breakdowns): IDataFrame {
+    const fipsColumn: string =
+      breakdowns.geography === "county" ? "county_fips" : "state_fips";
+
+    if (breakdowns.filterFips !== undefined) {
+      const fips = breakdowns.filterFips as Fips;
+      if (fips.isState() && breakdowns.geography === "county") {
+        df = df.where((row) => fips.isParentOf(row["county_fips"]));
```
aaronsn, line 54: It's generally good practice to avoid re-assigning function arguments (https://medium.com/the-andela-way/its-bad-practice-to-reassign-function-params-c37f289a5dc4). Usually doesn't matter but probably best to use a new variable. (same below)

frontend/src/data/variables/VariableProvider.ts
```diff
@@ -42,6 +44,56 @@ abstract class VariableProvider {
...
+      } else {
+        df = df.where((row) => row[fipsColumn] === fips.code);
+      }
+    }
+    return df;
```
aaronsn, line 59: resetIndex just to be safe? lol

frontend/src/charts/ChoroplethMap.tsx
```diff
@@ -16,8 +16,7 @@ const GEO_DATASET = "GEO_DATASET";
 const GEO_ID = "id";
 
 const VAR_DATASET = "VAR_DATASET";
-const VAR_STATE_FIPS = "state_fips";
-const VAR_COUNTY_FIPS = "COUNTY_FIPS";
+const VAR_FIPS = "fips";
```
aaronsn, line 19: I wonder if we should make a file of constants with standard column names? Not blocking this PR, just a thought to make refactoring easier because right now it's tricky to search for all usages of a particular column name

aaronsn (approved):
This is so cool!

frontend/src/charts/ChoroplethMap.tsx
kkatzen: added a TODO :)

PR #313, opened 2021-01-26
Implement CTP metadata standardization code

jenniebrown:
- Adds standardization and write to BQ code for CTP metadata
- Adds terraform config for the CTP metadata dataset
- Minor change to have CTP use class method to get table name instead of passing as an argument

#116


config/data_sources/variables.tf
```diff
@@ -0,0 +1,13 @@
+# Covid Tracking Project vars
```
aaronsn, line 1: are these comments accurate? They're suggesting that these are CTP-specific but the file name and variable names don't say anything about CTP

config/data_sources/covid_tracking_project.tf
```diff
@@ -5,3 +5,20 @@ resource "google_bigquery_dataset" "bq_covid_tracking_project" {
...
 }
+
+# We need to define the metadata table in Terraform so that we can grant
+# access to the gcs_to_bq runner to make queries to it.
+resource "google_bigquery_table" "covid_tracking_project_metadata" {
+  dataset_id = google_bigquery_dataset.bq_covid_tracking_project.dataset_id
+  table_id = var.ctp_metadata_table_name
+}
+
+resource "google_bigquery_table_iam_binding" "covid_tracking_project_metadata_iam_binding" {
```
aaronsn, line 16: For my understanding - why is this necessary?

If this is a pattern specifically required for some datasets based on whether they have to use a metadata table, maybe add this to the doc explaining how to add a new dataset?

python/datasources/covid_tracking_project_metadata.py
```diff
@@ -10,3 +11,48 @@ def get_id():
...
+        gcs_to_bq_util.append_dataframe_to_bq(df, dataset, self.get_table_name())
+
+    @staticmethod
+    def _metadata_columns_map():
+        """Returns a dict for renaming raw column names."""
+        return {
+            'api': 'reports_api',
+            'combined_category_other_than_api': 'reports_ind',
+            'ethnicity': 'reports_ethnicity',
+            'race': 'reports_race'
```
aaronsn, line 57: Not sure what the ideal place to document these columns is, but doesn't hurt to briefly summarize them in the code

python/datasources/covid_tracking_project_metadata.py
```diff
@@ -10,3 +11,48 @@ def get_id():
...
+        df = gcs_to_bq_util.load_csv_as_dataframe(gcs_bucket, gcs_file)
+        self.clean_frame_column_names(df)
+
+        # Standardize the data
+        keep_cols = [
+            'state_postal_abbreviation', 'api_death',
+            'defines_other_death', 'race_ethnicity_separately_death',
+            'race_ethnicity_combined_death', 'race_mutually_exclusive_death',
+            'combined_category_other_than_api_death', 'race_death',
+            'ethnicity_death', 'api_cases', 'defines_other_cases',
```
aaronsn, line 32: They only have these for deaths and cases I take it? Maybe add a comment about why there isn't anything for hospitalizations or tests

aaronsn (approved):
Sweet, this is very cool

config/data_sources/covid_tracking_project.tf
```diff
@@ -5,3 +5,20 @@ resource "google_bigquery_dataset" "bq_covid_tracking_project" {
   dataset_id = "covid_tracking_project"
   location   = "US"
 }
+
+# We need to define the metadata table in Terraform so that we can grant
```
mayaspivak, line 9: Isn't it generally recommended to define all resources in tf? It seems like this comment should go with the binding bellow.

mayaspivak (approved):

config/data_sources/covid_tracking_project.tf
jenniebrown, line 9: IIUC we don't actually define any other BQ tables, just the datasets. Our other tables are created when writing to BQ or through BQ routines.

config/data_sources/covid_tracking_project.tf
jenniebrown: In order for a user (or service account) to run query jobs, they need to have read access to the table. The way our project is set up, our tables are locked down on creation, so the service accounts running our Cloud Run services don't have access. So we need to grant them access by defining the table and its permissions in terraform. When testing on my cloud project, I wasn't able to get the gcs_to_bq service to query the metadata table unless this iam binding was defined.

This will be applicable to any of our Cloud Run services that wish to query any BQ table (e.g. if we wanted to join with county/state fips codes during standardization in a similar way). I'm currently writing documentation for this :) Also thinking about how we might allow new dependencies without having to re-create the table in terraform...

config/data_sources/variables.tf
jenniebrown: These variables are currently all specific to CTP, but if any other datasource needs to define variables, they would do it in this file. I added a comment to help clarify.

python/datasources/covid_tracking_project_metadata.py
jenniebrown: Yeah, they didn't provide us with any info on tests or hospitalizations, unfortunately.

python/datasources/covid_tracking_project_metadata.py
jenniebrown: Detailed definitions for these names are in the Structured Metadata for Covid Tracking Project doc. Ideally, that doc could get turned into a wiki and we could link to it from here.

config/data_sources/covid_tracking_project.tf
jenniebrown: Whoops, somehow this response was created in its own thread: 

IIUC we don't actually define any other BQ tables, just the datasets. Our other tables are created when writing to BQ or through BQ routines.

config/data_sources/covid_tracking_project.tf
aaronsn: Hmm, I think I ran into the a similar issue with the aggregator, and got it working by adding the `bigquery.tables.getData` role. It looks like the gcs_to_bq service doesn't have that permission - maybe that would fix it?

PR #314, opened 2021-01-26
Airflow config for covid tracking project and covid tracking project metadata

jenniebrown:
#116

### DIFF
--- a/.github/linters/mypy.ini
+++ b/.github/linters/mypy.ini
@@ -3,7 +3,7 @@
 python_version = 3.8
 
 # Per-module options. These 3rd-party libraries don't export type hints, so suppress errors from them.
-[mypy-airflow]
+[mypy-airflow.*]
 ignore_missing_imports = True
 
 [mypy-flask_cors]
--- a/.github/linters/mypy.ini
+++ b/.github/linters/mypy.ini
@@ -3,6 +3,9 @@
 python_version = 3.8
 
 # Per-module options. These 3rd-party libraries don't export type hints, so suppress errors from them.
+[mypy-airflow]
+ignore_missing_imports = True
+
 [mypy-flask_cors]
 ignore_missing_imports = True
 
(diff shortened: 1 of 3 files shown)


.github/linters/mypy.ini
```diff
@@ -3,6 +3,9 @@
 python_version = 3.8
 
 # Per-module options. These 3rd-party libraries don't export type hints, so suppress errors from them.
+[mypy-airflow.*]
+ignore_missing_imports = True
```
ldbnr4, line 7: Nice!

ldbnr4 (approved):
lgtm

ldbnr4:
Also, what were your thoughts on joining the metadata and covid tracking project tables? Did you consider putting the ingest to bq operator for the metadata table in the covid tracking project dag?

jenniebrown:
> Also, what were your thoughts on joining the metadata and covid tracking project tables? Did you consider putting the ingest to bq operator for the metadata table in the covid tracking project dag?

For joining with the metadata, the covid tracking project is going to perform a query to BQ to get the standardized metadata table during its write_to_bq stage. I thought about putting the metadata operator in the CTP dag, but since the metadata is manually uploaded, I thought it better to keep it simple and have it be its own dag. The metadata won't change frequently, so it will only need to be run when we obtain a new file. Since CTP is scheduled daily, doing it this way avoids any accidents if a new metadata file is uploaded but not approved before the CTP scheduled run kicks off. 

The joining logic is in this PR #313 if you're interested

ldbnr4:
> > Also, what were your thoughts on joining the metadata and covid tracking project tables? Did you consider putting the ingest to bq operator for the metadata table in the covid tracking project dag?
> 
> For joining with the metadata, the covid tracking project is going to perform a query to BQ to get the standardized metadata table during its write_to_bq stage. I thought about putting the metadata operator in the CTP dag, but since the metadata is manually uploaded, I thought it better to keep it simple and have it be its own dag. The metadata won't change frequently, so it will only need to be run when we obtain a new file. Since CTP is scheduled daily, doing it this way avoids any accidents if a new metadata file is uploaded but not approved before the CTP scheduled run kicks off.
> 
> The joining logic is in this PR #313 if you're interested

Sounds good, thanks!

PR #320, opened 2021-01-27
Make VariableProviders directly load the datasets they need which improves performance by not requiring unnecessary dataset loads

aaronsn:


frontend/src/data/variables/VariableProvider.ts
```diff
@@ -1,28 +1,17 @@
...
     this.providesMetrics = providesMetrics;
-    this.datasetIds = datasetIds;
   }
 
-  // TODO change return type to MetricQueryResponse instead of Row[]
-  getData(
-    datasets: Record<string, Dataset>,
-    breakdowns: Breakdowns
-  ): MetricQueryResponse {
+  async getData(breakdowns: Breakdowns): Promise<MetricQueryResponse> {
```
kkatzen, line 14: Do the tests all pass without any changes now that this is async?

aaronsn: Sorry, was still working on tests. They should be passing now! (Also the variable provider tests are somewhat simpler)

aaronsn:
Fixed merge conflicts and tests, please take a look!

kkatzen (approved):

Issue #307, opened 2021-01-22
Frontend changes to support backend race changes


kkatzen:
ACS & Covid Tracking project are going to split race_and_ethnicity into two columns:
something like:
race_or_ethnicity & reports_ethnicity

In the frontend, we'll use these two columns to create the race_and_ethnicity field (as we currently have it). (API should ultimately be returning same values to the cards)

Additionally, instead of one table for cases, deaths and hospitalizations we will have 3 different tables. Each variable may contain different values for various metadata fields (such as reports_ethnicity)

So the current single dataset feeding into CovidProvider is of the format:
FIPS/race_and_ethnicity/cases/hospitalizations/deaths

And we will now have it turn into 3 datasets of the format:
FIPS/race_or_ethnicity/reports_ethnicity/some_metadata/some_other_metadata/cases
FIPS/race_or_ethnicity/reports_ethnicity/some_metadata/some_other_metadata/deaths
FIPS/race_or_ethnicity/reports_ethnicity/some_metadata/some_other_metadata/hospitalizations

CovidVariable provider will now need to take a MetricQuery in getData() now so we know which datasetId to get since each Metric will now require a different one.
Additionally, we want MetricQueryResponse to have a new attribute returning those metadata fields (reports_ethnicity, some_metadata, some_other_metadata). Metadata fields in MetricQueryResponse can be used as warnings or footnotes on the card. For instance: "Races for NC covid cases does not include hispanic" and "Races for NC deaths does include hispanic"

kkatzen:
Given that we aren't using CTP anymore we're pivoting what we're doing here pre-launch.

Frontend will expect backwards compatibility as far as using the existing "race_and_ethnicity" column. ACS data from backend will now also send "race_includes_hispanic" boolean (https://github.com/SatcherInstitute/health-equity-tracker/issues/339#issuecomment-786762651). This boolean should be added to MetricQueryResponse and available for use in CardWrapper. For datasets that don't provide this value from backend, we'll hardcode it's value for now in each provider.

I'll sync with Chi to figure out how we are displaying this information in the Cards.

kkatzen:
![footnote_v1](https://user-images.githubusercontent.com/6402808/110867189-3a8f7300-827b-11eb-9bba-90c6ef0366ac.png)

vanshkumar:
The BE changes to add separate race and ethnicity columns are done. Over time post-launch, we should move the FE to start using these new columns. Moving to backlog

[...]

PR #329, opened 2021-01-28
VariableProvider.getData takes MetricQuery instead of breakdowns

kkatzen:
#307


frontend/src/data/variables/AcsPopulationProvider.ts
```diff
@@ -54,7 +54,10 @@ class AcsPopulationProvider extends VariableProvider {
     throw new Error("Not implemented");
   }
 
-  async getDataInternal(breakdowns: Breakdowns): Promise<MetricQueryResponse> {
+  async getDataInternal(
+    metricQuery: MetricQuery
+  ): Promise<MetricQueryResponse> {
+    const breakdowns = metricQuery.breakdowns;
```
aaronsn, line 60: I forgot, where did we land on the "additional columns" aspect? Are we okay with returning additional columns that were not requested in the query or should we strip them out? Maybe we should add a method spec for getData() that says how it behaves

frontend/src/data/variables/VariableProvider.ts
```diff
@@ -14,17 +18,17 @@ abstract class VariableProvider {
...
+  async getData(metricQuery: MetricQuery): Promise<MetricQueryResponse> {
+    if (!this.allowsBreakdowns(metricQuery.breakdowns)) {
       return createMissingDataResponse(
         "Breakdowns not supported for provider " +
           this.providerId +
           ": " +
-          breakdowns.getUniqueKey()
+          metricQuery.breakdowns.getUniqueKey()
       );
     }
```
aaronsn, line 29: Now that this accepts a MetricQuery, I think it would be nice to throw an error if the query contains a metric that is not in `this.providesMetrics`. I think it should be a normal error rather than a missing data response, because that would indicate a bug in our variable mapping logic

aaronsn (approved):

frontend/src/data/variables/AcsPopulationProvider.ts
kkatzen: I'm inclined to remove them. I think I'll add a TODO and follow up :)

PR #331, opened 2021-01-28
Swap service account token for basicAuth

mayaspivak:
Using basicAuth overrides the service account token in the "Authorization" header. So put the service account token in a temporary header until right before the proxy request is sent.

### DIFF
--- a/frontend_server/server.js
+++ b/frontend_server/server.js
@@ -81,16 +81,16 @@ app.use('/api', apiProxy);
 // auth middleware must be installed before setting up routes so it applies
 // to the whole site.
 if (!getBooleanEnvVar("DISABLE_BASIC_AUTH")) {
-    const username = assertEnvVar("BASIC_AUTH_USERNAME");
-    const password = assertEnvVar("BASIC_AUTH_PASSWORD");
-    app.use(basicAuth({
-      // Temporary values until we can use Github Secrets. Also needs to be set up
-      // so that it's disabled for production but enabled for the test site.
-      users: { [username]: password },
-      challenge: true,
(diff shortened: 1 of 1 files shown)


frontend_server/server.js
```diff
@@ -69,6 +69,11 @@ const apiProxyOptions = {
   target: assertEnvVar("DATA_SERVER_URL"),
   changeOrigin: true, // needed for virtual hosted sites
   pathRewrite: { '^/api': '' },
+  onProxyReq: (proxyReq, req, res) => {
```
mayaspivak, line 72: @aaronsn This feels pretty hacky. Is there a better convention for storing the token until the onProxyReq event?

Cant actually fetch the token during the event because the proxy request will be sent before the token is fetched.

aaronsn: Hmm, I see. Did you test it out and it works this way?

I'm guessing there's probably a way to create a custom middleware function that waits for the header and then calls the proxy directly with the header as an argument..... buuut we're definitely reaching the limit of my understanding of express.

TBH this doesn't seem that bad to me. It's probably less complex than anything else I can think of. I'd maybe rename `AuthorizationTemp` to something like `Authorization_DataServer` and expand on the comment with why it's set up this way.

aaronsn (approved):

frontend_server/server.js
mayaspivak: Yea, this is the only way I could get it to work. I tried writing additional middleware to only use basic auth on routes that didn't start with /api but no matter what I did the basic auth header still ended up there.

PR #334, opened 2021-01-28
Tidying frontend VariableProvider tests

kkatzen:


frontend/src/data/variables/TestUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+export const ASIAN = "Asian (Non-Hispanic)";
+export const TOTAL = "Total";
+export const RACE = "race_and_ethnicity";
+export const AGE = "age";
+export const SEX = "sex";
+
+export interface FipsSpec {
+  code: string;
+  name: string;
+}
```
aaronsn, line 18: Optional nit: potentially some of these could be in a common non-test file?

frontend/src/data/variables/TestUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+export const MARIN: FipsSpec = {
+  code: "06041",
+  name: "Marin County",
+};
+export const USA: FipsSpec = {
+  code: USA_FIPS,
+  name: USA_DISPLAY_NAME,
+};
+
+export async function evaluateWithAndWithoutTotalInternal(
```
aaronsn, line 44: Optional nit: I usually don't start filename with a capital letter unless they export something with the same name. Just a personal preference.

aaronsn, line 44: I'd maybe recommend making this function `createWithAndWithoutTotalEvaluator` that returns a function

aaronsn (approved):

frontend/src/data/variables/TestUtils.ts
kkatzen: oo neat idea

PR #336, opened 2021-01-29
Adding unique population MetricId to MetricConfig

kkatzen:
woo


frontend/src/data/variables/CovidProvider.ts
```diff
@@ -141,6 +148,38 @@ class CovidProvider extends VariableProvider {
...
+      if (metricQuery.metricIds.includes(reportingPopulation)) {
+        df = df
+          .generateSeries({
+            [reportingPopulation]: (row) => row["population_pct"],
+          })
+          .resetIndex();
+      }
+    });
+    df = df.dropSeries(["population"]).resetIndex();
+    df = df.dropSeries(["population_pct"]).resetIndex();
```
aaronsn, line 181: Do this in one call, passing both columns

aaronsn (approved):
Thanks for the quick fix!

PR #337, opened 2021-01-29
Rename BRFSS race categories and add diabetes and copd percent share metrics

aaronsn:


frontend/src/data/MetricConfig.ts
```diff
@@ -77,6 +79,10 @@ export function formatFieldValue(metricType: MetricType, value: any): string {
 
 // TODO - strongly type key
 // TODO - count and pct_share metric types should require populationComparisonMetric
+
+// Note: metrics must be declared in a consistent order becuase the UI relies
```
kkatzen, line 83: oh good catch!

kkatzen (approved):
exciting stuff :D

PR #343, opened 2021-01-29
Remove nonstandard race categories by generalizing query to specify filters

### LINKED ISSUE #256: Remove non standardized race option
Breakdown should no longer depend on the non standardized race config

aaronsn:
Closes #256

### DIFF
--- a/frontend/src/data/variables/AcsPopulationProvider.test.ts
+++ b/frontend/src/data/variables/AcsPopulationProvider.test.ts
@@ -24,8 +24,8 @@ import {
   USA,
 } from "./TestUtils";
 import {
-  WHITE,
-  ASIAN,
+  WHITE_NH,
+  ASIAN_NH,
   TOTAL,
   RACE,
   AGE,
@@ -33,6 +33,7 @@ import {
   MALE,
   FEMALE,
   NON_HISPANIC,
+  WHITE,
 } from "../Constants";
 import { onlyIncludeStandardRaces } from "../query/BreakdownFilter";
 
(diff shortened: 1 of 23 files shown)


frontend/src/data/variables/AcsPopulationProvider.test.ts
```diff
@@ -187,6 +192,56 @@ describe("AcsPopulationProvider", () => {
...
 
+  test("State and Race Breakdown with standard race filter", async () => {
+    const rawData = [
+      stateRow(AL, RACE, TOTAL, 2),
+      stateRow(AL, RACE, ASIAN, 2),
+      stateRow(NC, RACE, TOTAL, 20),
+      stateRow(NC, RACE, ASIAN, 5),
+      stateRow(NC, RACE, WHITE, 15),
+      // Non-standard, will be excluded from the non-standard filter
+      stateRow(NC, RACE, "White", 17),
```
kkatzen, line 203: Maybe we should rename WHITE to WHITE_NH and create WHITE as "White"?

frontend/src/data/variables/AcsPopulationProvider.test.ts
```diff
@@ -187,6 +192,56 @@ describe("AcsPopulationProvider", () => {
...
+      stateRow(NC, RACE, NON_HISPANIC, 13),
+    ];
+
+    const datasetId = "acs_population-by_race_state_std";
+    dataFetcher.setFakeDatasetLoaded(datasetId, rawData);
+
+    let breakdowns = Breakdowns.forFips(new Fips(NC.code)).andRace();
```
kkatzen, line 210: Could you just put the breakdowns directly in the MetricQuery? I don't think we need to reuse the same variable

frontend/src/data/variables/AcsPopulationProvider.test.ts
```diff
@@ -187,6 +192,56 @@ describe("AcsPopulationProvider", () => {
...
+      )
+    );
+
+    breakdowns = Breakdowns.forFips(new Fips(NC.code)).andRace(
+      onlyIncludeStandardRaces()
+    );
+    response = await new AcsPopulationProvider().getData(
+      new MetricQuery("population", breakdowns)
+    );
+    expect(response).toEqual(
```
kkatzen, line 233: I wonder if we should update createWithAndWithoutTotalEvaluator to return a function that takes an Array of BreakdownFilters and forEach them instead of hard coding the total/not total dichotomy

kkatzen (approved):
oooo good stuff very satisfying :D such flexibility

frontend/src/data/variables/AcsPopulationProvider.test.ts
aaronsn: Done, and added a TODO to add constants for other values

frontend/src/data/variables/AcsPopulationProvider.test.ts
aaronsn: You say that now, wait till our formatter gets its hands on it :P 

Anyway I moved it directly to the MetricQuery

frontend/src/data/variables/AcsPopulationProvider.test.ts
aaronsn: Yeah I thought about that, but I worried about over-generalizing a test utility that becomes hard to maintain. It's easy to have those things get out of control (I say this from personal failures lol). I think it can be valuable if done carefully but requires some thought with the best way to design

PR #346, opened 2021-01-29
Make sub-folders of data/ to better organize different types

aaronsn:
Also move FIPS to data folder


kkatzen (approved):
ooo so organized!

PR #353, opened 2021-02-04
BUG - TableChartDataTable keeps re-rendering for some reason

efregoso:
I cannot figure out why TableCardDataTable component keeps re-rendering.  The page loads fine for a few seconds, and the table appears normal, but then it goes to a page with a "Maximum Depth Exceeded" error.  This doesn't happen when I switch out TableChartDataTable with DataTable + fake data, so it has to be something in TableChartDataTable.


frontend/src/dataset_explorer/TableChartDataTable.tsx
```diff
@@ -0,0 +1,184 @@
...
+    };
+  });
+  columns = [{
+    Header: BREAKDOWN_VAR_DISPLAY_NAMES[breakdownVar],
+    accessor: "0",
+  }].concat(columns);
+
+  const demoData = [
+    {
+      0: "Hi",
```
Sethvg, line 72: Cant help but feel this was a missed oppertunity to use:

0: Hell0
1: He11o

Sethvg (approved):

Issue #221, opened 2020-12-21
Wrap text on Vega labels


kkatzen:
instead of truncating the vega label text, it would be nice to wrap it onto a new line

Issue #220, opened 2020-12-17
Bar Chart ordering should be consistent


kkatzen:
Rather than sorted by any value, let's sort the bars by a consistent ordering of the categories (race, age, gender)

kkatzen:
For age based bar charts, we want to make sure they are ordered from youngest to oldest. Alphabetically may or may not work, we should double check before closing this bug

kkatzen:
I'm not seeing any bar charts that aren't ordered so I'll close this

PR #354, opened 2021-02-04
Updating Disparity Bar Chart labels

kkatzen:
Adding multi-line label support based on hardcoded delimiters for long breakdown value names.

Screen Shot 2021-02-04 at 5 51 35 PM

#220 
#221


frontend/src/charts/DisparityBarChart.tsx
```diff
@@ -203,11 +234,26 @@ export function DisparityBarChart(props: DisparityBarChartProps) {
   const [ref, width] = useResponsiveWidth(
     100 /* default width during intialization */
   );
+
+  let data = props.data;
+  // Some standarized race names should be annotated with delimiters to indicate linebreaks
+  if (props.breakdownVar === RACE) {
+    data = props.data.map((data) => {
+      let copy = Object.assign({ ...data });
+      if (Object.keys(MULTILINE_RACE_MAP).includes(data.race_and_ethnicity)) {
```
aaronsn, line 243: nit: pull the Object.keys() call outside the loop so it's only done once

frontend/src/charts/DisparityBarChart.tsx
```diff
@@ -203,11 +234,26 @@ export function DisparityBarChart(props: DisparityBarChartProps) {
...
+        copy["race_and_ethnicity"] =
+          MULTILINE_RACE_MAP[data.race_and_ethnicity];
+      }
+      return copy;
+    });
+  }
+  data.sort((a, b) => (a[props.breakdownVar] > b[props.breakdownVar] ? 1 : 0));
```
aaronsn, line 250: You'd want 1, -1, and 0 instead of just 1 and 0. But you can just use the built-in `localeCompare` function for sorting which I think is better

aaronsn (approved):

frontend/src/charts/DisparityBarChart.tsx
```diff
@@ -7,6 +7,23 @@ import {
   BREAKDOWN_VAR_DISPLAY_NAMES,
 } from "../data/query/Breakdowns";
 import { MetricConfig } from "../data/config/MetricConfig";
+import { RACE } from "../data/utils/Constants";
+
+const DELIMITER = "*~*";
+// TODO - it may be worth calculating where delimiters should go based on line length if we have a lot more breakdown values in future
```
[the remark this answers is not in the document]
kkatzen: just let me lazy, aaron!! hehe jk you're right I should just do it now :P

PR #356, opened 2021-02-05
Disparity Bar chart chooses domain of larger metric

### LINKED ISSUE #216: Disparity Bar Chart axis issues
The disparity bar chart y Axis is based on the range of the thicker measure bar, it should be based on the range of whichever of the thick or thin bar is larger

https://github.com/SatcherInstitute/het-frontend/blob/50f1cfca97e368d6105b4089d4120cfef02894e9/app/src/features/charts/DisparityBarChart.tsx#L133

kkatzen:
Closes #216

See North Carolina Covid Cases for Durham county for an example


frontend/src/charts/DisparityBarChart.tsx
```diff
@@ -33,6 +33,19 @@ function getSpec(
   // We use nested ternerys to determine the label's y axis delta based on the number of lines in the label to vertically align
   const AXIS_LABEL_Y_DELTA = `length(${MULTILINE_LABEL}) == 2 ? -3 : length(${MULTILINE_LABEL}) > 2 ? -7 : 5`;
 
+  function maxValueInField(field: string) {
+    return Math.max(
+      ...data
+        .map((row) => row[field])
+        .filter((value: number | undefined) => !!value)
```
aaronsn, line 40: `!!value` will cause you to filter out `0` which is a valid number (though in practice this will probably never be an issue). Better to do `!== undefined`

aaronsn (approved):
nice fix!

PR #358, opened 2021-02-09
Support sorting for TableChart

efregoso:
Creation of TableChartDataTable for sorting of info displayed in TableChart
One pending TODO


frontend/src/dataset_explorer/TableChartDataTable.tsx
```diff
@@ -0,0 +1,150 @@
...
+      </TableCell>
+  );
+}
+
+/** Corollary for TableHeaderGroup **/
+function TableHeaderGroup({ group }: { group: HeaderGroup<any> }) {
+  return (
+      <TableRow {...group.getHeaderGroupProps()}>
+        {group.headers.map((col, index) => (
+            <TableHeader column={col} key={index} />
```
kkatzen, line 38: I'm inclined to actually just inline TableHeader rather than having its own component

frontend/src/dataset_explorer/TableChartDataTable.tsx
```diff
@@ -0,0 +1,150 @@
...
+      return <TableCell {...cell.getCellProps()}>{cell.render("Cell")}</TableCell>;
+    }
+  }
+
+  return (
+      <TableContainer component={Paper} style={{maxHeight: "500px"}}>
+        <Table stickyHeader {...getTableProps()}>
+          <TableHead>
+            {headerGroups.map((group, index) => (
+                <TableHeaderGroup group={group} key={index} />
```
kkatzen, line 121: Just a style preference, but I might rename this and CustomTableRow such that their names echo their distinct and mutually exclusive roles more explicitly: TableHeaderRow and TableDataRow

frontend/src/dataset_explorer/TableChartDataTable.tsx
```diff
@@ -0,0 +1,150 @@
...
+      {
+        columns: memoCols,
+        data: memoData,
+        initialState: {pageSize: 5},
+      },
+      useSortBy,
+      usePagination
+  );
+
+  function CustomTableRow({ row, }: { row: Row<any> }) {
```
kkatzen, line 85: It took me a little while to figure out why the data row components were inside the main component and the header components were not. I see now that it's because you're using prepareRow here.

You might consider moving the header row inside as well just for consistency?

kkatzen (commented):
Looking good! :) I think there's a build issue though so I wasn't able to see the netlify preview demo. Could you take a look? Thanks!

frontend/src/charts/TableChart.tsx
```diff
@@ -1,70 +1,135 @@
...
+      Header: metricConfig.fullCardTitleName,
+      accessor: metricConfig.metricId,
+    };
+  });
+  columns = [{
+    Header: BREAKDOWN_VAR_DISPLAY_NAMES[breakdownVar],
+    accessor: breakdownVar as MetricId,
+  }].concat(columns);
+
+  // eslint-disable-next-line
```
kkatzen, line 37: I may add a note what this is escaping. If someone else modifies the line later it would be easier for them to know if the lint escape is still necessary

frontend/src/charts/TableChart.tsx
```diff
@@ -1,70 +1,135 @@
 import React from "react";
```
kkatzen, line 1: You could also import useMemo here so you don't have to do React.useMemo below:
import React, {useMemo}

just a style preference. up to you

frontend/src/charts/TableChart.tsx
```diff
@@ -1,70 +1,135 @@
...
+                    direction={col.isSortedDesc ? "desc" : "asc"}
+                />
+              </TableCell>
+          ))}
+        </TableRow>
+    );
+  }
+
+  /** Component for the table's data rows **/
+  function TableDataRow({row,}: { row: Row<any> }) {
```
kkatzen, line 78: delete extra comma?

kkatzen (approved):
Looks good! :D I did notice that sometimes the columns appear to be changing size when you toggle them... I'm okay if you fix it now or if you want to file a bug to fix later. Your call!

PR #360, opened 2021-02-10
Fixes map breakdown filter bug

### LINKED ISSUE #338: Map chart filter and legend sometimes break
image

Not sure what is causing it, but steps to repro:
1. Select a state
2. Select "COPD"
3. Select "the United States"

For some reason doing it in this order causes the filter and legend to break, but it doesn't happen when you just select COPD/US right away.

kkatzen:
Fixes #338

Issue was that when you view a map combination where there are no breakdown values breakdownFilter was set to undefined so we needed to update logic to consider undefined in addition to default of empty string.


Sethvg (approved):

Sethvg:
Love your commit names ;p

PR #382, opened 2021-02-23
Splitting Data Source and Datasets for Data Catalog page

### LINKED ISSUE #264: Frontend: Refactor DatasetMetadata for download and data catalog requirements
acs_population-by_race_state_std and acs_population-by_age_state are unique dataset IDs.

This means they would show up as separate items in the data catalog. We would like them to be in the same card instead

see FakeMetadataMap

### LINKED ISSUE #345: Frontend: Implement datasource download popup in UI
As per #264:
"Data catalog will have one card per data source and clicking download will open a pop up dialog that will, for MVP, list all files available for download. Downloads only happen from this popup."

kkatzen:
closes #264
closes #345

Data Catalog will now display Data Sources per the Config file. Each listing has a download button that opens a dialog that lists the latest update time based on the dataset metadata from DataFetcher (which is currently fake). A lot of static metic on Datasets is moved to static Data Source config.

See various download states here:
Screen Shot 2021-02-23 at 3 36 21 PM


frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.tsx
```diff
@@ -0,0 +1,173 @@
...
+      case "loaded":
+        return <CheckCircleIcon />;
+      default:
+        return "";
+    }
+  };
+
+  if (props.datasetMetadata === undefined) {
+    return <></>;
+  }
```
aaronsn, line 56: Why allow undefined?

frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const getIcon = () => {
+    switch (downloadStatus) {
+      case "unloaded":
+        return <GetAppIcon />;
+      case "loading":
+        return <CircularProgress className={styles.DownloadIcon} />;
+      case "loaded":
+        return <CheckCircleIcon />;
+      default:
+        return "";
```
aaronsn, line 50: Typescript should be able to detect that you've covered all the cases, so you can switch default to `case "error":` to be more explicit

frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.tsx
```diff
@@ -0,0 +1,173 @@
...
+import ListItemIcon from "@material-ui/core/ListItemIcon";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import Alert from "@material-ui/lab/Alert";
+import ListItemText from "@material-ui/core/ListItemText";
+import CheckCircleIcon from "@material-ui/icons/CheckCircle";
+import GetAppIcon from "@material-ui/icons/GetApp";
+import { IconButton } from "@material-ui/core";
+import CloseIcon from "@material-ui/icons/Close";
+
+export type LoadStatus = "loading" | "unloaded" | "error" | "loaded";
```
aaronsn, line 27: Does this need to be exported?

frontend/src/data/config/MetadataMap.ts
```diff
@@ -375,8 +306,65 @@ const fakeMetadata: DatasetMetadata[] = [
...
+> = Object.fromEntries(dataSourceMetadataList.map((m) => [m.id, m]));
 
-export default FakeMetadataMap;
+export const FakeDatasetMetadataMap: Record<
+  string,
+  DatasetMetadata
+> = Object.fromEntries(
+  datasetMetadataList.map((m) => {
+    let metadataWithSource = m;
+    let test = Object.values(DataSourceMetadataMap).find((metadata) =>
```
aaronsn, line 364: nit: use const

frontend/src/data/config/MetadataMap.ts
```diff
@@ -1,18 +1,10 @@
-import { DatasetMetadata, MetadataMap } from "../utils/DatasetTypes";
+import { DatasetMetadata, DataSourceMetadata } from "../utils/DatasetTypes";
 
-const fakeMetadata: DatasetMetadata[] = [
+const datasetMetadataList: DatasetMetadata[] = [
```
aaronsn, line 3: To improve the distinction between what's fake vs permanent, could we move this to another file and name it FakeDatasetMetadata?

frontend/src/pages/DataCatalog/dataset_explorer/DatasetExplorer.tsx
```diff
@@ -88,29 +75,37 @@ function DatasetExplorer(props: { preFilterDatasetIds: string[] }) {
...
         <WithMetadata>
-          {(metadata) => {
-            const viewingSubsetOfDatasets =
-              getFilteredDatasetIds(metadata, activeFilter).length === 1;
+          {(datasetMetadata) => {
+            const filteredDatasets = getFilteredSources(
+              DataSourceMetadataMap,
+              activeFilter
+            );
+            // Check if more than the default filters are enabled to see if you're viewing subset of sources
```
aaronsn, line 83: nit: line too long

frontend/src/pages/DataCatalog/dataset_explorer/DatasetExplorer.tsx
```diff
@@ -43,33 +46,17 @@ function DatasetExplorer(props: { preFilterDatasetIds: string[] }) {
...
-    );
-    setActiveFilter((prevFilter) => {
-      const names = prevFilter[NAME_FILTER_ID];
-      const newNames = names.filter((name) => validIds.includes(name));
-      return {
-        ...prevFilter,
-        [NAME_FILTER_ID]: newNames,
-      };
-    });
-  });
```
aaronsn, line 60: What's the reason for getting rid of this?

frontend/src/pages/DataCatalog/dataset_explorer/DatasetExplorer.tsx
```diff
@@ -88,29 +75,37 @@ function DatasetExplorer(props: { preFilterDatasetIds: string[] }) {
...
+            );
+            // Check if more than the default filters are enabled to see if you're viewing subset of sources
+            const viewingSubsetOfSources =
+              Object.keys(activeFilter).length > 1 ||
+              activeFilter[NAME_FILTER_ID].length > 0;
 
             const defaultDatasetNames = props.preFilterDatasetIds
-              .filter((datasetId) => !!metadata[datasetId])
-              .map((datasetId) => metadata[datasetId].name);
+              .filter((datasetId) => !!FakeDatasetMetadataMap[datasetId])
```
aaronsn, line 89: I don't know if we should be using this directly. Eventually we will need to be getting last-updated metadata from the server because it's dynamic, and this will make it harder to do that.

Alternatively, maybe we can split this into two: a DatasetMetadataConfig and DynamicDatasetMetadata (names need improvement lol). The former can be static config, the latter can be only the dynamic stuff that needs to come from the server.

frontend/src/pages/DataCatalog/dataset_explorer/downloadDataset.ts
```diff
@@ -1,5 +1,3 @@
-import Button from "@material-ui/core/Button";
-import React from "react";
 import { getDataManager } from "../../../utils/globals";
 
 function download(filename: string, content: string) {
```
aaronsn, line 3: nit: add comment explaining return type

aaronsn (changes requested):
Love the download dialog! Looks really great

frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.tsx
kkatzen: In case there is a misconfiguration and there isn't available metadata for a dataset I thought it was better to fail gracefully by just hiding that dataset.

aaronsn: Maybe add `getLogger().logError()`? I'm still holding onto hope that one day we'll hook up the logger to real reports that we can monitor on the server

kkatzen: good call!

frontend/src/pages/DataCatalog/dataset_explorer/DatasetExplorer.tsx
kkatzen: Sorry I actually changed preFilterDatasetIds to be source IDs but forgot to rename. I didn't think we needed to check against the dataset metadata anymore since all the data sources are in static config.

frontend/src/pages/DataCatalog/dataset_explorer/DatasetExplorer.tsx
kkatzen: oh wait i actually had gotten the data properly dynamically I just forgot to use it :P

kkatzen (commented):
Could you took another look, Aaron? Thanks :)

frontend/src/pages/DataCatalog/dataset_explorer/DatasetExplorer.tsx
aaronsn: haha nice

frontend/src/pages/DataCatalog/dataset_explorer/DatasetExplorer.tsx
aaronsn: Oh okay! I think you can delete useOnMetadataLoaded in that case, since I'm pretty sure it's only used here

aaronsn (approved):

PR #385, opened 2021-02-24
Aggregating, standardizing, and ingesting the CDC restricted data

vanshkumar:
Tested locally that it produces the appropriate CSVs


vanshkumar:
Updated PR to also include logic for ingesting CDC restricted data. Note that this is all at the county-level, state-level will be another PR.

vanshkumar:
Added Jennie and Aaron as reviewers, mostly as FYI

python/datasources/cdc_restricted_local.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    # Add a columns of all ones, for counting the # of cases / records.
+    df[std_col.COVID_CASES] = np.ones(df.shape[0], dtype=int)
+
+    # Add columns for hospitalization yes/no/unknown and death yes/no/unknown,
+    # as we aggregate and count these individually. Do a sanity check that we
+    # covered all the data and drop the original hospitalization/death columns.
+    df[std_col.COVID_HOSP_Y] = (df['hosp_yn'] == 'Yes')
+    df[std_col.COVID_HOSP_N] = (df['hosp_yn'] == 'No')
+    df[std_col.COVID_HOSP_UNKNOWN] = ((df['hosp_yn'] == 'Unknown') |
```
Sethvg, line 107: Might better logic just be != 'Yes and != 'No'.  Makes it more future proof.  For instance if they change 'hosp_yn' to 'NA'

python/datasources/cdc_restricted_local.py
```diff
@@ -0,0 +1,251 @@
...
+                                      (df['hosp_yn'] == 'Missing') |
+                                      (df['hosp_yn'] == 'OTH'))
+    df[std_col.COVID_DEATH_Y] = (df['death_yn'] == 'Yes')
+    df[std_col.COVID_DEATH_N] = (df['death_yn'] == 'No')
+    df[std_col.COVID_DEATH_UNKNOWN] = ((df['death_yn'] == 'Unknown') |
+                                       (df['death_yn'] == 'Missing'))
+
+    assert (df[std_col.COVID_HOSP_Y] | df[std_col.COVID_HOSP_N] |
```
Sethvg, line 115: Are these asserts still needed?  If so does the error message make this easy to debug?

python/datasources/cdc_restricted_local.py
```diff
@@ -0,0 +1,251 @@
...
+            df[std_col.COVID_HOSP_UNKNOWN]).all()
+    assert (df[std_col.COVID_DEATH_Y] | df[std_col.COVID_DEATH_N] |
+            df[std_col.COVID_DEATH_UNKNOWN]).all()
+
+    df = df.drop(columns='hosp_yn')
```
Sethvg, line 120: you can combine these into one call.

python/datasources/cdc_restricted_local.py
```diff
@@ -0,0 +1,251 @@
...
+    "50 - 59 Years": "50-59",
+    "60 - 69 Years": "60-69",
+    "70 - 79 Years": "70-79",
+    "80+ Years": "80+",
+    "NA": "Unknown",
+    "Missing": "Unknown",
+}
+
+
+def accumulate_data(df, groupby_cols, overall_df, demog_col,
```
Sethvg, line 87: nit: spell out demog.  Took me a minute to understand demog was demographic.

python/datasources/cdc_restricted_local.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def standardize_data(df):
+    """Standardizes the data by cleaning string values and standardizing column
+    names.
+
+    df: Pandas dataframe to standardize.
+    """
+    # Clean string values in the dataframe.
+    def _clean_str(x):
```
Sethvg, line 151: Is there a way in Python for extension methods?  Like can this be added to the string python proto or something rather than a local func?

python/datasources/data_sources.py
```diff
@@ -16,6 +17,7 @@
 DATA_SOURCES_DICT = {
     ACSPopulation.get_id(): ACSPopulation(),
     CDCCovidDeaths.get_id(): CDCCovidDeaths(),
+    CdcRestrictedData.get_id(): CdcRestrictedData(),
```
Sethvg, line 20: Nit: Keep consistent Capitalization with other CDC file

Sethvg (approved):

python/datasources/cdc_restricted_local.py
vanshkumar: done

python/datasources/cdc_restricted_local.py
vanshkumar: this one actually i had an extra "strip" in there so i removed that and made it into an inline lambda

python/datasources/data_sources.py
vanshkumar: done

Issue #370, opened 2021-02-16
Data Dashboard: Tooltip label lists everything as "State:"


efregoso:
Mousing over the state in the "Per 100,000 people" card, the tooltip label has “State:” listed for everything, including counties and cities.
We want to change this as appropriate.

PR #386, opened 2021-02-25
Add manual legend to MapCard

### LINKED ISSUE #218: MapChart - toggling between breakdown should have common legend
We should have an explicit legend passed into the map card if we're going to toggle the heat map based on a breakdown variable otherwise it's misleading

kkatzen:
Closes #218 
#370


frontend/src/data/query/MetricQuery.ts
```diff
@@ -65,10 +65,21 @@ export class MetricQueryResponse {
...
+    if (fieldValues.length === 0) {
+      return undefined;
     }
+    return {
+      min: Math.min.apply(Math, fieldValues),
+      max: Math.max.apply(Math, fieldValues),
+    };
+  }
+
+  getUniqueFieldValues(fieldName: string): string[] {
```
aaronsn, line 82: Did you mean to drop the `if (this.isFieldMissing(fieldName)) {` check?

frontend/src/charts/ChoroplethMap.tsx
```diff
@@ -14,6 +15,7 @@ const MISSING_DATASET = "MISSING_DATASET";
 const VALID_DATASET = "VALID_DATASET";
 const GEO_DATASET = "GEO_DATASET";
 const GEO_ID = "id";
+const COLOR_SCALE = "COLOR_SCALE";
```
aaronsn, line 18: Why did this change from "colorScale" to "COLOR_SCALE"? Does this work by referencing the below variable by the "name" property?

aaronsn (approved):

frontend/src/charts/ChoroplethMap.tsx
kkatzen: Yeah I was trying to change the arbitrary string names to the variables to reduce human error.

frontend/src/data/query/MetricQuery.ts
kkatzen: oops good catch

PR #398, opened 2021-02-26
MadLib Fips selector: hide autocomplete until user types 2 characters

### LINKED ISSUE #377: Data Dashboard: Only show dropdown options on type-in location field when user has typed 1+ characters
Currently, the type-in field for selecting a location presents a dropdown of every location available on the first user click.
We want to amend this so that this option dropdown only presents when a user has typed in at least 1 character as they are searching for their location. 

Another consideration for this could be subcategorizing the options in the location dropdown and alphabetizing them; this can be considered further after implementation.

kkatzen:
Closes #377


efregoso (approved):
Tbabm you for this!!!

PR #402, opened 2021-03-01
Data provider skeleton for Health Insurance

Sethvg:


frontend/src/data/variables/AcsHealthInsuranceProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+  async getDataInternal(
+    metricQuery: MetricQuery
+  ): Promise<MetricQueryResponse> {
+    const breakdowns = metricQuery.breakdowns;
+    let df = await this.getDataInternalWithoutPercents(breakdowns);
+
+    // Calculate totals where dataset doesn't provide it
+    // TODO: this should be removed when Totals come from the Data Server. Note
+    // that this assumes that the categories sum to exactly the total
+    const breakdownsToSum: DemographicBreakdownKey[] = ["age", "sex"];
```
kkatzen, line 61: Since this is the same as AcsPopulationProvider, we could make it a function in VariableProvider and call it in the children to reduce code reuse

PR #414, opened 2021-03-08
ACS Health Insurance File diff And -> Or

Sethvg:
The circuit was short circuiting if any of the diffs were the same

### DIFF
--- a/python/datasources/acs_health_insurance.py
+++ b/python/datasources/acs_health_insurance.py
@@ -235,19 +235,19 @@ class AcsHealhInsuranceIngestor:
 
         female_state_params = format_params(
             HEALTH_INSURANCE_BY_SEX_GROUPS_PREFIX, HEALTH_INSURANCE_BY_SEX_FEMALE_SUFFIXES)
-        file_diff = file_diff and url_file_to_gcs.url_file_to_gcs(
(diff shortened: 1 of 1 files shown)


python/datasources/acs_health_insurance.py
```diff
@@ -235,21 +235,21 @@ def upload_to_gcs(self, bucket):
 
         female_state_params = format_params(
             HEALTH_INSURANCE_BY_SEX_GROUPS_PREFIX, HEALTH_INSURANCE_BY_SEX_FEMALE_SUFFIXES)
-        file_diff = file_diff and url_file_to_gcs.url_file_to_gcs(
+        file_diff = url_file_to_gcs.url_file_to_gcs(
             self.base_url, female_state_params, bucket,
-            self.get_filename(Sex.FEMALE, None, False))
+            self.get_filename(Sex.FEMALE, None, False)) or file_diff
```
vanshkumar, line 240: Could you keep the structure of file_diff = file_diff or url_file_to_gcs... ? IMO it helps with readability

vanshkumar (approved):

python/datasources/acs_health_insurance.py
Sethvg: No.  This short circuits the code and if one file diff is different will not upload any of the other datasets to GCS

PR #416, opened 2021-03-08
Added auto find URLS and buckets to update env

Sethvg:


airflow/update-environment-variables.sh
```diff
@@ -1,31 +1,40 @@
 #!/usr/bin/env bash
 echo Input ingestion service url?
-read ingestionUrl
+INGESTION_URL=$(gcloud run services list --platform managed --filter ingestion-service | grep -o "https:.*\.app")
```
jenniebrown, line 3: You can use --format 'value(status.url)' instead of grepping.

I also prefer using gcloud run services describe instead of list since it is more direct, but I'll leave that up to you since it requires a region and I'm not sure how straightforward it is to get that here (I think we could hard code it but that might be annoying later)

jenniebrown (changes requested):

airflow/update-environment-variables.sh
Sethvg: I switched to using the format value :)  Ima leave the rest as is as its working and I dont wanna spend too much time on it.

airflow/update-environment-variables.sh
```diff
@@ -1,31 +1,40 @@
 #!/usr/bin/env bash
 echo Input ingestion service url?
```
vanshkumar, line 2: Are these "echo Input ..." lines still needed?

airflow/update-environment-variables.sh
```diff
@@ -1,31 +1,40 @@
...
 echo Input aggregator service url?
-read aggegatorServiceUrl
+AGGREGATOR_URL=$(gcloud run services list --platform managed --filter aggregator-service --format 'value(status.url)')
+echo "Using Fetched Data: ${AGGREGATOR_URL}"
 
 echo Input landing bucket?
-read landingBucket
+LANDING_BUCKET=$(gsutil ls | grep -oP ".*landing.*")
+LANDING_BUCKET=${LANDING_BUCKET:5:-1}
+echo $LANDING_BUCKET
```
vanshkumar, line 25: Could you echo a more descriptive message for LANDING_BUCKET and MANUAL_BUCKET, eg "Using ${LANDING_BUCKET} as landing bucket" and similar for manual bucket?

vanshkumar (approved):

PR #417, opened 2021-03-09
Adding dataset to config for health insurance

Sethvg:


config/main.tf
```diff
@@ -57,4 +57,10 @@ resource "google_bigquery_dataset" "bq_cdc_covid_deaths_dataset" {
   location   = "US"
 }
 
+# Create a BigQuery dataset for cdc covid deaths
```
vanshkumar, line 60: Can you instead add this as a separate tf file under config/data_sources (eg add acs_health_insurance.tf in that folder)?

Sethvg: Done

config/data_sources/acs_health_insurance.tf
```diff
@@ -0,0 +1,5 @@
+# Create a BigQuery dataset for cdc covid deaths
```
vanshkumar, line 1: should say acs health insurance

vanshkumar (approved):

PR #418, opened 2021-03-09
Add side-by-side option to disparity bar chart

### LINKED ISSUE #375: Data Dashboard: Separating the overlaid bars in the “Percent Share” bar chart and placing them side-by-side with equal width
Currently, the bars in the "Percent share" bar chart card are overlaid and of different widths. 
For ease of comparison and in response to user testing feedback, we want to place these bars side by side instead, and make them of equal width.

kkatzen:
Closes #375 

Note that this change leaves the old stacked version of the disparity bar chart in tact. You can toggle between them in storybook


efregoso (approved):
This looks awesome!  I don't see the ability to toggle between stacked and side-by-side in storybook, but I don't think we need it.

Issue #333, opened 2021-01-28
Move requirements files into top-level requirements/ directory


jenniebrown:
We currently have requirements.txt files scattered throughout our codebase. Originally, we decided to keep each requirements file it its related directory, but with the expansion of libraries and services, this has made requirements difficult to find and can hide relationships between different requirements files.

One way to address this is to move all requirements files to the same directory. This will make them easier to find and add consistency for running update commands. The files will live in the top-level directory requirements. Each service or library will put its .in and .txt files there, named after the service or library. For example, our testing requirements will have names requirements/tests.in and requirements/tests.txt, and the GCS to BQ service will have names requirements/run_gcs_to_bq.in and requirements/run_gcs_to_bq.txt. 

When updating, all commands can be run from the same directory, `pip-compile --output-file=requirements/<name>.txt requirements/<name>.in`.

This change will require us to update our .github workflows and Dockerfiles that install requirements files.

PR #419, opened 2021-03-09
Python install script

Sethvg:


install-all-python.sh
```diff
@@ -0,0 +1,15 @@
+# because why not
+pip install wheel
+
+for file in $(find . -name "requirements.txt")
```
jenniebrown, line 4: The test requirements are actually annoyingly called requirements/tests.txt. I have a bug (#333) where I intended to move all the requirements files into requirements/ so they were easier to find but never got around to finishing it.

But this script would work to install all the prod reqs

jenniebrown (approved):

PR #426, opened 2021-03-10
Add cdc_restricted dataset to TF, update script for new hosp_yn value in latest data

vanshkumar:


vanshkumar:
Also added a mapping for state abbreviations, so I can do state name -> fips code later while joining with the ACS population table

python/datasources/cdc_restricted_local.py
```diff
@@ -106,6 +107,7 @@ def accumulate_data(df, groupby_cols, overall_df, demographic_col,
     df[std_col.COVID_HOSP_N] = (df['hosp_yn'] == 'No')
     df[std_col.COVID_HOSP_UNKNOWN] = ((df['hosp_yn'] == 'Unknown') |
                                       (df['hosp_yn'] == 'Missing') |
+                                      (df['hosp_yn'] == 'nul') |
```
efregoso, line 110: Is this supposed to be "null"?

efregoso (approved):

python/datasources/cdc_restricted_local.py
efregoso: Offline conversation -- for some reason the data uses "nul" so this is not a typo.

Issue #389, opened 2021-02-25
Add small multiples to MapCard


kkatzen:

kkatzen:
Bare minimum functionality is submitted, but there's still some fit&finish to be done here with UX and copy regarding the math of the scales

kkatzen:
![Screen Shot 2021-03-15 at 10 21 36 AM](https://user-images.githubusercontent.com/6402808/111197598-e4247c00-857b-11eb-8689-45f38c3efd86.png)

Update filter to support age and sex breakdowns

efregoso:
Completion time estimate: < 1 week

achipham:
Change color scheme to #yellowgreen 
https://screenshot.googleplex.com/7oiH7eac5EsdtSh

achipham:
Add 16px spacing between Filter By: and Category
https://screenshot.googleplex.com/AHY9uirgg2oLm4Q

achipham:
Change info bubble: https://screenshot.googleplex.com/86BJmLuDW7sNEba

Text:
Note that the legend changes between different categories. To see all categories at once, click 'Show full breakdowns' link

Background color: #F8F9FA
Outline + info icon: #0B5240

<img width="603" alt="Screen Shot 2021-03-18 at 1 11 49 PM" src="https://user-images.githubusercontent.com/31259975/111683757-85653980-87eb-11eb-99d6-7e6d0d0683b6.png">

achipham:
Modal popup changes: https://screenshot.googleplex.com/7PHHDYshbczQYEL

PR #430, opened 2021-03-11
First draft of small multiples

kkatzen:
#389

The details may not be quite right but the functionality should all be there. We're still waiting for MSM/SHLI input so we'll be revisiting this card.


frontend/src/data/query/MetricQuery.ts
```diff
@@ -79,6 +79,12 @@ export class MetricQueryResponse {
     };
   }
 
+  getValidRowsForField(fieldName: string) {
```
aaronsn, line 82: Add comment?

frontend/src/cards/ui/MultiMapDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+  breakdown: string;
+  fips: Fips;
+  data: Row[];
+  fieldRange: FieldRange | undefined;
+  open: boolean;
+  handleClose: () => void;
+  breakdownValues: string[];
+}
+
+export function MultiMapDialog(props: MultiMapDialogProps) {
```
aaronsn, line 26: Add comment for what this is (I think this is a good habit to start getting into for components unless it's really obvious).

Same for the legend and two-level dropdown

Doesn't need a comment for every prop, but would be good to explain how the major behavior changing ones work. Eg breakdown & breakdownValues

frontend/src/cards/ui/MultiMapDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+      scroll="paper"
+      aria-labelledby="Dialog showing choropleth maps of each breakdown category with the same scale."
+    >
+      <DialogContent dividers={true}>
+        <Grid container justify="space-around">
+          <Grid item xs={6}>
+            <Alert severity="info">
+              This scale is a{" "}
+              <a href="https://vega.github.io/vega/docs/scales/#quantile">
+                quantile
```
aaronsn, line 41: I'd pull "quantile" into a variable and use that throughout this class so if it were to change it's easy to just update the variable value

frontend/src/cards/MapCard.tsx
```diff
@@ -55,21 +53,33 @@ function MapCardWithKey(props: MapCardProps) {
...
+
+  const dialogElementRef = useRef<HTMLElement>(null);
+  useEffect(() => {
+    if (smallMultiplesDialogOpen) {
+      const { current: dialogElement } = dialogElementRef;
+      if (dialogElement !== null) {
+        dialogElement.focus();
+      }
+    }
+  }, [smallMultiplesDialogOpen]);
```
aaronsn, line 71: You can probably extract this into a simple hook `[isOpen, setIsOpen] = useAutoFocusDialog()`

frontend/src/cards/MapCard.tsx
```diff
@@ -138,86 +174,75 @@ function MapCardWithKey(props: MapCardProps) {
...
-                <Alert severity="warning">
-                  No data available for filter: <b>{breakdownFilter}</b>
-                </Alert>
+            {!queryResponse.dataIsMissing() &&
+              dataForActiveBreakdownFilter.length !== 0 &&
+              props.metricConfig && (
+                <CardContent>
+                  <Alert severity="info">
+                    Note that legend changes between races. To see races with
+                    common legend, use show all breakdowns button.
```
aaronsn, line 225: Can you make the text on alerts left-aligned? It looks a bit odd center-aligned when the second text line is very short

frontend/src/cards/MapCard.tsx
```diff
@@ -138,86 +174,75 @@ function MapCardWithKey(props: MapCardProps) {
...
             <Divider />
-            <CardContent>
-              {queryResponse.dataIsMissing() && (
+            {queryResponse.dataIsMissing() && (
+              <CardContent>
                 <Alert severity="error">No data available</Alert>
+              </CardContent>
+            )}
+            {!queryResponse.dataIsMissing() &&
+              dataForActiveBreakdownFilter.length === 0 && (
```
aaronsn, line 212: What's the use case for this condition? Sorry, I'm having trouble following the different data-missing scenarios

frontend/src/cards/ui/TwoLevelDropDown.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </Button>
+
+      <ListPopover
+        popover={firstMenu}
+        currentSelection={firstMenuSelection}
+        items={oneLevelMenu ? Object.values(props.options)[0] : props.options}
+        onClick={(event: React.MouseEvent<HTMLElement>, value: string) => {
+          if (oneLevelMenu) {
+            props.onOptionUpdate(value);
+            firstMenu.close();
```
aaronsn, line 100: Does this case need to reset the firstMenuSelection state? Or no because it will never be used in a one-level menu?

aaronsn (approved):
This is pretty darn sweet.

It's also very complicated. I don't think I can really validate all of it for correctness, so please carefully test it manually!

frontend/src/cards/MapCard.tsx
kkatzen: If there is some data, but not for the currently selected filter. For instance if you select Unknown there is no field data.

frontend/src/cards/ui/TwoLevelDropDown.tsx
kkatzen: Well I was using it to highlight the currently selected item, but on second thought I think it's better if we don't highlight the current selections. So now it's fine to not reset it since it's only used to render the second menu, which you can only open by selecting a new first menu value.

frontend/src/cards/MapCard.tsx
kkatzen: I'm not even sure if we want this to be the final mock or if we just did this for UX testing... I have a bug already for these alerts so I'll add a note to address this too https://github.com/SatcherInstitute/health-equity-tracker/issues/431

kkatzen:
Thanks for all the comments Aaron :D

PR #434, opened 2021-03-12
Add aggregation for CDC restricted data

vanshkumar:
Add aggregation logic for joining CDC restricted data with the public state_fips dataset & the ACS population tables. Also fix a standardization issue in the cdc_restricted local script to make county_fips a string.

Note this also updates/uses the "_staging" tables, which is a change that will have to be updated to use "_latest" once dedupe work is complete

### DIFF
--- a/airflow/dags/cdc_restricted.py
+++ b/airflow/dags/cdc_restricted.py
@@ -27,11 +27,19 @@ cdc_bq_payload = util.generate_bq_payload(
     _CDC_RESTRICTED_WORKFLOW_ID,
     _CDC_RESTRICTED_DATASET,
     gcs_bucket=Variable.get('GCS_MANUAL_UPLOADS_BUCKET'),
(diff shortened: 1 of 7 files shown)


ldbnr4 (changes requested):
After looking this over again it might be better for us to revert the changes in acs_population.py and go back to using get_table_name_by_race and get_table_name_by_sex_age_race until all dedupe changes have been submitted.

https://github.com/SatcherInstitute/health-equity-tracker/blob/master/python/datasources/acs_population.py#L212-L217

vanshkumar:
> After looking this over again it might be better for us to revert the changes in acs_population.py and go back to using get_table_name_by_race and get_table_name_by_sex_age_race until all dedupe changes have been submitted.
> 
> https://github.com/SatcherInstitute/health-equity-tracker/blob/master/python/datasources/acs_population.py#L212-L217

PR #442, opened 2021-03-17
ACS Refactor and Income Ingestion

Sethvg:


python/ingestion/acs_utils.py
```diff
@@ -0,0 +1,106 @@
...
+
+def parseMetadata(raw_metadata_trimmed, required_keys, metadataInitializer):
+    parsed = {}
+    for k, v in raw_metadata_trimmed.items():
+        label = v['label']
+        parts = label.split("!!")
+        label_meta = metadataInitializer(k)
+
+        for part in parts:
+            if(part == 'Estimate'):
```
vanshkumar, line 52: You can combine these into

if part == 'Estimate' or part == 'Total:'

python/datasources/acs_income.py
```diff
@@ -0,0 +1,248 @@
...
+            "Population": "104",
+        },
+        ...
+    } Note: This can be debugged via the acs_income_internal_data.json file
+    '''
+
+    def accumulate_acs_data(self, data):
+        key_row = data[0]
+        for row in data[1::]:
+            data = {}
```
vanshkumar, line 149: It would be good to use a different variable name than "data" here as it's already an argument being passed into the function

vanshkumar (approved):
This is adding some non-trivial logic that IMO we should definitely add tests for, up to you if you want to do that in this PR or in a future PR. If you want to do it in a future PR, please open a github issue to remember to do so

python/ingestion/standardized_columns.py
```diff
@@ -11,6 +11,7 @@
 COUNTY_FIPS_COL = "county_fips"
 COUNTY_NAME_COL = "county_name"
 POPULATION_COL = "population"
+INCOME_COL = "income"
```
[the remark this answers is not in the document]
Sethvg: Ima leave this one as is, as the columns are basic headers and context can be infered from the table.

PR #444, opened 2021-03-18
Dedupe removal

ldbnr4:


airflow/dags/acs_health_insurance.py
```diff
@@ -41,5 +38,5 @@
     'acs_health_insurance_exporter', acs_hi_exporter_payload, data_ingestion_dag)
 
 # Ingestion DAG
-(acs_hi_gcs_operator >> acs_hi_gcs_short_op >>
```
vanshkumar, line 44: Don't we still want to not upload to GCS to there is no file-level diff?

run_ingestion/main.py
```diff
@@ -29,10 +29,8 @@ def ingest_data():
     logging.info(f"message: {event}")
 
     try:
-        if ingest_data_to_gcs(event):
-            return ('', HTTPStatus.CREATED)
-        else:
-            return ('', HTTPStatus.NO_CONTENT)
+        ingest_data_to_gcs(event)
+        return ('', HTTPStatus.NO_CONTENT)
```
vanshkumar, line 33: Should this be CREATED?

airflow/dags/acs_health_insurance.py
ldbnr4: As discussed offline, this will put us back in the pre-existing state where we overwrite what's in the GCS bucket. We can optimize in another change to keep the file level check, only overwrite when necessary and always continue with the rest of the pipeline.

run_ingestion/main.py
ldbnr4: Good call. Updated

Issue #376, opened 2021-02-16
Data Dashboard: "Total” / Alphabetical / “Unreported unknowns” ordering in bar charts


efregoso:
Currently, the ordering for bar chart categories is purely alphabetical, except for "Unknown" which is always at the top. 
We want to amend this ordering so that "Total" is always at the top and "Unreported unknowns" (vs. "Declined to report unknowns") are always at the bottom, with every other category in between and in alphabetical order.

efregoso:
Completion time estimate: < 1 week

PR #445, opened 2021-03-18
Re-order VariableProvider output

kkatzen:
Addresses #376


efregoso (changes requested):
It looks like the Unknowns are still displaying above the "White" racial categories in the bar charts -- can you move the Unknowns to the very bottom?

kkatzen:
> It looks like the Unknowns are still displaying above the "White" racial categories in the bar charts -- can you move the Unknowns to the very bottom?

Done!

efregoso (approved):

Issue #308, opened 2021-01-23
Split Covid Tracking Project output into separate tables for cases, deaths, hospitalizations, and tests


jenniebrown:
Because metadata columns for CTP may differ for each variable type within a state, we want to split the standardized CTP data into a separate table for each variable. This avoids having one table with multiples of each metadata column, which would be inconvenient for joining and require string parsing the column names for lookups. Each table will have the following:

date | state | race | metadata columns ... | variable (e.g. cases)

When joined with ACS, a population column will be added. This will require no change on the frontend except to request multiple tables. Since the frontend doesn't use more than one variable at a time, this will be no different from other lookups.

More discussion can be found in https://github.com/SatcherInstitute/health-equity-tracker/pull/288#discussion_r562043485

Part of #116

PR #450, opened 2021-03-20
Split ctp data into separate tables

jenniebrown:
This change splits CTP data into a separate table for each of cases, deaths, hospitalizations, and tests. This is to help make it easier to use the metadata without having to encode data type in each metadata column name. 

A few weeks ago the CTP data was not parsing (from the March 2 update), but the newest (and final) version parses fine.

#308


jenniebrown:
Adding Aaron just for visibility

vanshkumar (approved):

PR #452, opened 2021-03-22
Install all python fix

Sethvg:


joshzarrabi (approved):

joshzarrabi (changes requested):
Hi @Sethvg, I;m still seeing the same `/usr/bin/clang` failure on my own machine

PR #477, opened 2021-03-25
Add footnote for includes_hispanic metadata, hard coded on data source

### LINKED ISSUE #475: Add includes_hispanic to static DataSourceMetadata
The CardWrapper will read DataSourceMetadata to get the includes_hispanic and update a footer/info bar to indicate which the datasets are

kkatzen:
Closes #475

### DIFF
--- a/frontend/src/cards/ui/Sources.tsx
+++ b/frontend/src/cards/ui/Sources.tsx
@@ -62,6 +62,7 @@ export function Sources(props: {
     return !response.isFieldMissing("race_and_ethnicity");
   });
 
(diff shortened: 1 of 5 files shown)


vanshkumar:
The latest on this after the discussion with Ebony is to *not* alter the race/ethnicity labels that we show (and so not have the footnote explaining that all categories include/exclude hispanic), and instead just add a static footnote at the bottom explaining what hispanic/latino means per dataset. I think you can keep this PR for the new footnote, but what it's going to say & how it's going to vary exactly per dataset is not totally clear right now

Issue #349, opened 2021-02-04
Data Dashboard: Change "total" to "all"


efregoso:
On the "COVID Cases per 100,000 people" card, the dropdown currently has "total" listed as an option.  Change to "all".

efregoso:
Completion time estimate: < 3 days for quick functional fix

efregoso:
Completion time estimate: < 2 days

PR #479, opened 2021-03-25
MapCard: Rename Total to All

kkatzen:
Addresses #349

Closes final issue in #389

### DIFF
--- a/frontend/src/cards/MapCard.tsx
+++ b/frontend/src/cards/MapCard.tsx
@@ -183,7 +183,7 @@ function MapCardWithKey(props: MapCardProps) {
                         options={filterOptions}
                         onOptionUpdate={(
                           newBreakdownDisplayName,
-                          selection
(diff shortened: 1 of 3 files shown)


vanshkumar (commented):
Looks like the filter still doesn't default to "All", at least in your deploy preview, wasn't that the goal of this change?

efregoso:
> Looks like the filter still doesn't default to "All", at least in your deploy preview, wasn't that the goal of this change?

No, the goal per the Fit and Finish Meeting Chi and Larry and I had was just to change the wording to "All".  We can make a separate issue for defaulting to the "All" option and approve this in the meantime.

efregoso:
Whoops. sorry.  I thought that button would resolve the convo, not close the issue!  Appr

efregoso (approved):

vanshkumar (approved):

Issue #209, opened 2020-11-04
Update Footer content


kkatzen:
Replace placeholders in footer or remove

efregoso:
Completion time estimate: < 1 day, pending content approval

PR #483, opened 2021-03-26
Update footer styles per mock

kkatzen:
Addresses #209

Also moving some images into subfolders


efregoso:
Completed in #468 -- thank you for filling in on this while it was still WIP :)

PR #484, opened 2021-03-26
CDC Restricted FE

vanshkumar:
Almost all of the CDC restricted FE code, tested e2e with storybook. What remains, and will come in future PRs:
- Fix unknown handling (proper alert isn't shown right now for "no data" right now)
- Fix an odd issue with 1 hospitalization per 100k in Arizona
- Compute reporting_population and reporting_population_pct properly using reported data

### DIFF
--- a/frontend/src/data/variables/CdcCovidProvider.test.ts
+++ b/frontend/src/data/variables/CdcCovidProvider.test.ts
@@ -1,0 +1,352 @@
+import CdcCovidProvider from "./CdcCovidProvider";
+import AcsPopulationProvider from "./AcsPopulationProvider";
+import { Breakdowns, BreakdownVar } from "../query/Breakdowns";
+import { MetricQuery, MetricQueryResponse } from "../query/MetricQuery";
+import { Fips } from "../utils/Fips";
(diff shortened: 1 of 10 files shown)


frontend/src/data/variables/CdcCovidProvider.test.ts
```diff
@@ -0,0 +1,352 @@
...
+const dataFetcher = getDataFetcher() as FakeDataFetcher;
+
+describe("cdcCovidProvider", () => {
+  beforeEach(() => {
+    resetCacheDebug();
+    dataFetcher.resetState();
+    dataFetcher.setFakeMetadataLoaded(FakeDatasetMetadataMap);
+  });
+
+  test("County and Race Breakdown", async () => {
```
kkatzen, line 126: You might consider also adding age & sex breakdown tests

kkatzen (approved):

frontend/src/data/variables/CdcCovidProvider.test.ts
vanshkumar: added more tests

vanshkumar:
Note that I made a small change to AcsPopulationProvider to calculate Total only when it is not present, rather than always for age/sex

PR #486, opened 2021-03-27
Move from appending to overwriting when doing the GCS to BQ step

vanshkumar:


vanshkumar:
Added all BE folks for visibility

PR #496, opened 2021-03-29
Calculate covid totals on the BE

vanshkumar:


vanshkumar:
Verified for a couple counties/states that cases per 100k match up between the dev site and the output of the modified query

joshzarrabi (approved):
Looks good!

Issue #497, opened 2021-03-30
Add title to Small Multiples overlay


adams314:
When a user views the small multiples breakdown, they lose the context of what metrics are being compared across groups.

The title structure should be:

[Metric] across all [Breakdown] groups, e.g.: 

*COVID-19 deaths per 100,000 people in United States* across all *Race and Ethnicity* groups

kkatzen:
![Screen Shot 2021-04-01 at 10 56 29 AM](https://user-images.githubusercontent.com/6402808/113335132-6087c080-92d9-11eb-983e-29f1ef5b5c21.png)

PR #513, opened 2021-04-01
Add title to Small Multiples Map dialog

kkatzen:
Addresses #497
There were no mocks, so I tried to match the card title style


kkatzen:
Update: there were mocks, I found them, here is the comparison: 
Screen Shot 2021-04-02 at 9 28 46 AM

Screen Shot 2021-04-02 at 9 28 35 AM

vanshkumar (approved):

PR #525, opened 2021-04-02
Update Race standardization on the backend

aaronsn:
- Introduce race_category_id which is a unique identifier for a race category. Similar to race_and_ethnicity but not tied to user-facing display names.
- Update Race enum to encode all race-related data for each category.
- Update datasets to include race_category_id, race, and race_includes_hispanic columns
- Add an ACS population value for Multiracial/other combined, and update CDC restricted data to use the correct ACS value for multi/other


aaronsn:
Still need to test this in GCP, working through some issues with my GCP instance setup, but wanted to get a review in the meantime

joshzarrabi (approved):
This looks good to me! I came to the same conclusion re where to use OTHER_STANDARD/ OTHER_NONSTANDARD

aaronsn:
This is ready for review, thanks!

config/run.tf
```diff
@@ -55,7 +55,8 @@ resource "google_cloud_run_service" "gcs_to_bq_service" {
 
         resources {
           limits = {
-            memory = "2G"
+            memory = "8G"
+            cpu = "2000m"
```
aaronsn, line 59: @vanshkumar So initially I set this to 4GB and it was fine for population, but still ran out of memory for I think household income. The issue is that since this is adding breakdown metadata columns to every dataset that has a race breakdown for convenience on the frontend, it adds a lot of extra raw data. Apparently you can't set memory greater than 4GB without increasing CPU, which defaults to 1000m so this doubles it. Not sure if this is an issue.

There are a few alternatives:
* Move the addition of the metadata columns to SQL in the aggregation step, and just use the race_category_id col in pandas. This would need to be done for every dataset.
* Hard-code the race_category_id => (race, race_and_ethnicity, race_includes_hispanic) mapping on the frontend and process the datasets on the frontend before using them.
* Chunk the writing to BQ so it doesn't write all rows at once
* A larger refactor which I emailed you about - I think this would be really good to do post-launch, but is definitely too much work to do before launch.

Given that the pipeline isn't run very often, my take is to keep the memory/cpu increase and then look into the larger refactor post-launch. Happy to hear yours/others' thoughts or discuss further though!

vanshkumar: Thanks for listing the alternatives, I'm not sure what the cost will be of increasing mem to 8GB and doubling cpu, but as the cloud run service is up at all times, I imagine there will be some cost - I'll try to figure out how much that might be.

IMO at least until launch, it's OK to up this to 4GB and leave the cpu unchanged; household income is not a variable that we are planning to include by launch anyways. We can punt for now and a do a proper data pipeline refactor post launch

vanshkumar (approved):

config/data_sources/AGG_cdc_restricted.sql
```diff
@@ -14,38 +14,47 @@ WITH
...
         SUM(cases) as cases,
         SUM(hosp_y) as hosp_y,
         SUM(hosp_n) as hosp_n,
         SUM(hosp_unknown) as hosp_unknown,
         SUM(death_y) as death_y,
         SUM(death_n) as death_n,
         SUM(death_unknown) as death_unknown,
+        'Total' as race,
+        CAST(NULL as BOOL) as race_includes_hispanic,
+        'Total' as race_and_ethnicity,
```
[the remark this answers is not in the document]
aaronsn: I thought we had decided to use the race_and_ethnicity after all based on feedback from MSM? Also I remember discussing that it would be a decently large refactor to make the frontend always use "race" instead of "race_and_ethnicity", and we didn't think it was necessarily worth doing all at once? I thought we would just supply both and the frontend can use whichever it needs. The only downside to doing both is data transfer size/memory usage.

config/run.tf
aaronsn: I'll need to retest all the datasets and make sure HHI is the only one that didn't work with 4GB

config/data_sources/AGG_cdc_restricted.sql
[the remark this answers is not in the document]
vanshkumar: Right yeah, I was referring to the larger refactor on the FE to use race instead of race_and_ethnicity, which we can do over time. I think we're on the same page

config/run.tf
jenniebrown: > Move the addition of the metadata columns to SQL in the aggregation step, and just use the race_category_id col in pandas. This would need to be done for every dataset.

Why would this need to be for every dataset? If the metadata columns are added in the aggregation step then jobs that need to use that data could read from the final table instead of the intermediate one output by gcs_to_bq.

aaronsn: I'm not sure I follow what you're saying Jennie - could you elaborate?

Vansh - I updated to 4GB. All the pipelines run except for household income. I'll file a github issue for that.

python/ingestion/standardized_columns.py
```diff
@@ -31,27 +50,134 @@
...
-    OTHER_NH = "Some other race (Non-Hispanic)"
-    TOTAL = "Total"
-    MULTI = "Two or more races"
-    MULTI_NH = "Two or more races (Non-Hispanic)"
-    WHITE = "White"
-    WHITE_NH = "White (Non-Hispanic)"
-    UNKNOWN = "Unknown race"
-    UNKNOWN_NH = "Unknown race (Non-Hispanic)"
-    API = "Asian and Pacific Islander"
+    AIAN = ("AIAN", "American Indian and Alaska Native", True)
```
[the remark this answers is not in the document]
aaronsn: Restructured these to hopefully clarify things more, lmk if it's better now

vanshkumar (approved):

PR #526, opened 2021-04-05
Newsletter form works and sharing links work

efregoso:


vanshkumar (approved):

vanshkumar:
Actually, I see the same issue for your deploy-preview as Krista's, and our dev site seems to be down.. not sure what's going on, looking into it

PR #556, opened 2021-04-08
Updating "Total" to "All" in VariableProviders

### LINKED ISSUE #534: Rename "Total" to "All" fully in the tracker across all cards

kkatzen:
Closes #534

### DIFF
--- a/frontend/src/data/utils/Constants.ts
+++ b/frontend/src/data/utils/Constants.ts
@@ -1,6 +1,8 @@
 // Constant strings to be used in tests and elsewhere
 
 // TODO add values for the other categories. Maybe convert to an enum.
+export const ALL = "All";
+export const ALL_OF_KNOWN = "All of known";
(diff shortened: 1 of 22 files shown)


frontend/src/data/utils/Constants.ts
```diff
@@ -1,6 +1,8 @@
 // Constant strings to be used in tests and elsewhere
 
 // TODO add values for the other categories. Maybe convert to an enum.
+export const ALL = "All";
+export const ALL_OF_KNOWN = "All of known";
```
vanshkumar, line 5: Doesn't seem like this is used anywhere?

vanshkumar (approved):
Thanks for doing this!

frontend/src/data/utils/Constants.ts
kkatzen: good catch!

Issue #552, opened 2021-04-08
Update population profile card


vanshkumar:
We may want to sort them differently as well, but for now let's make sure that both rows of race categories are at least visible without requiring a user click

kkatzen:
Reopening- if your browser is really short, the new length of the population card prevents you from seeing changes when you update variables in the madlib which could be confusing... I'll noodle on this I think Chi had some other mock ideas

kkatzen:
<img width="1437" alt="Screen Shot 2021-04-12 at 11 59 00 AM" src="https://user-images.githubusercontent.com/6402808/114447016-c3edda00-9b86-11eb-89ce-7e1e768e2baa.png">

On a 13" macbook pro

kkatzen:
In our last sync, eng decided we'd like to collapse this card per Chi's mocks- product said they would verify with MSM that this was okay

adams314:
![image](https://user-images.githubusercontent.com/80131692/115266620-5c93d500-a106-11eb-8549-5d16bba6fca4.png)

For reference, this is what the condensed profile card looks like.

adams314:
Confirming that it's OK to condense this by default. 

Adding to this per #615 we are going to put the disclaimer about how the racial categories are defined into the expanded profile card.

Add an info alert to the expanded cart above the list of groups:

These racial categories are defined by the ACS and US Census Bureau. While it is the standard for CDC reporting, the definition of these categories often results in not counting or miscounting people in underrepresented groups. [Learn more]().  -- Link goes to the FAQ>Known Limitations section

adams314:
Race/Ethnicity groups are still listed in order of size descending, please change to alpha sort to match the rest of the card.

PR #557, opened 2021-04-08
Population Card style changes

kkatzen:
Addressing #552


kkatzen:
Compare the old and new mobile views. We can maintain the old mobile view if we want.

New:
Screen Shot 2021-04-08 at 11 08 50 AM

Old:
Screen Shot 2021-04-08 at 11 22 54 AM

vanshkumar (approved):

Issue #544, opened 2021-04-07
About Us: UI quirks on mid-size browser


kkatzen:
![Screen Shot 2021-04-07 at 1 03 52 PM](https://user-images.githubusercontent.com/6402808/113927143-f328d380-97a1-11eb-9c44-9f1dbc016f8b.png)
![Screen Shot 2021-04-07 at 1 04 18 PM](https://user-images.githubusercontent.com/6402808/113927148-f3c16a00-97a1-11eb-94c6-19f9dd5352da.png)

Issue #541, opened 2021-04-07
Dark material button hover/click color is still blue


kkatzen:
Clicking on this button you can see it's blue:

![Uploading Screen Shot 2021-04-07 at 12.55.57 PM.png…]()
(get to this screen by filtering the data sources)

PR #561, opened 2021-04-08
Update static content button styles

kkatzen:
Addresses part of #544
note that hover color is tracked in #541

### DIFF
--- a/frontend/src/pages/AboutUs/AboutUsPage.module.scss
+++ b/frontend/src/pages/AboutUs/AboutUsPage.module.scss
@@ -30,8 +30,7 @@
   font-size: 18px !important;
   font-weight: 500;
   font-family: "DM Sans", sans-serif !important;
-  text-decoration: none !important;
-  padding: 10px 30px 10px 30px;
+  padding: 10px 30px;
 }
 
 .NoUnderline {
--- a/frontend/src/pages/AboutUs/AboutUsPage.module.scss
+++ b/frontend/src/pages/AboutUs/AboutUsPage.module.scss
@@ -24,16 +24,18 @@
 
 // For primary-themed buttons
 .PrimaryButton {
-  background-color: #0b5240;
-  color: white !important;
-  border-radius: 40px;
+  border-radius: 40px !important;
   height: 52px;
   width: 220px;
-  font-size: 18px;
+  font-size: 18px !important;
   font-weight: 500;
   font-family: "DM Sans", sans-serif !important;
-  text-decoration: none;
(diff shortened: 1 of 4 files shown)


frontend/src/pages/AboutUs/AboutUsPage.module.scss
```diff
@@ -24,16 +24,17 @@
 
 // For primary-themed buttons
 .PrimaryButton {
-  background-color: #0b5240;
```
efregoso, line 27: Verifying jic so I know I'm understanding the changes -- these are already applied via Material Button styles right?

frontend/src/pages/Landing/LandingPage.module.scss
```diff
@@ -11,14 +11,15 @@
 .PrimaryButton {
   background-color: #0b5240;
   color: white !important;
-  border-radius: 40px;
-  height: 52px;
```
efregoso, line 15: Verifying these changes rq too -- these are preserved in the Material Button stylings?

frontend/src/pages/AboutUs/TheProjectTab.tsx
```diff
@@ -112,12 +140,18 @@ function TheProjectTab() {
                     xs={12}
                   >
                     <Grid item xs={12} sm={12} md={5}>
-                      <a
-                        href={EXPLORE_DATA_PAGE_LINK}
-                        className={styles.PrimaryButton}
+                      <LinkWithStickyParams
```
efregoso, line 143: Thank u for integrating these into the new static pages!

efregoso (approved):

frontend/src/pages/Landing/LandingPage.module.scss
kkatzen: The only styles that I removed from CSS were the height and width since material will calculate those automatically to be the right size. So if someone changed the text later they wouldn't have to update the height and width.

frontend/src/pages/AboutUs/AboutUsPage.module.scss
kkatzen: Yes, the material components automatically apply certain styles based on the Theme. For instance background and font colors. You can reconfigure defaults in the Theme: https://github.com/SatcherInstitute/health-equity-tracker/blob/master/frontend/src/styles/MaterialTheme.tsx

If you're overriding Material component styles in the stylesheet in a class, sometimes you have to put !important at the end to make sure they're prioritized over the defaults.

Issue #531, opened 2021-04-05
A11y bugs


efregoso:
Previous versions of the site have gone through multiple a11y reviews.  Some of the bugs no longer apply to the new version, but others do.

Links to bug collection sheets:
- https://docs.google.com/spreadsheets/d/1lLAScKYz1NvrDhnc2H-Z108qCOnk97GoKK0yKtcNtmA/edit?pli=1&resourcekey=0-eZit_ziVRfJFjjwssv5E2Q#gid=148246793
- https://docs.google.com/spreadsheets/d/1G8tfAx8_4GjTgopR5Okwi2G2WZ4mEyqTMCVq1qpardg/edit?pli=1#gid=0

efregoso:
Generally, the non-obsolete bugs are as follows:

1. Allow users to skip header with screenreader controls
2. Labeling buttons correctly (in navigation)
3. Having clear heading hierarchy on page
4. Having clear labels for region (header, body, footer)
5. Alt text for images

PR #563, opened 2021-04-08
Address some outstanding a11y issues across static pages excluding Explore Data & minor grid breakpoint correction

efregoso:
As these have been done without consult, they will need further review by an a11y point of contact.

Tentatively addresses #531 , specifically:
- Labeling buttons correctly (in navigation)
- Having clear heading hierarchy on page
- Having clear labels for region (header, body, footer)
- Allow users to skip header with screenreader controls
- ... And some small corrections to the grid breakpoints of the "Join the Effort" section on the "What is Health Equity" page.


efregoso:
Currently doing an independent a11y review based on internal documentation -- will update PR accordingly

frontend/src/App.tsx
```diff
@@ -60,50 +60,54 @@ function MobileAppToolbar() {
...
-             className={styles.AppbarLogoImg}
-             alt="Health Equity Tracker Logo -- decorative 'H'"/>
-      </LinkWithStickyParams>
-      <Typography variant="h6" className={styles.HomeLogo}>
+      <Toolbar className={styles.AppToolbar}>
         <LinkWithStickyParams to="/">
-          Health Equity Tracker
+          <img src="img/AppbarLogo.png"
+               className={styles.AppbarLogoImg}
+               alt="Health Equity Tracker Logo -- decorative 'H'"/>
```
justinpetersen, line 92: I would consider the logo decorative (no announcement needed) and use alt="". Since the title is announced just prior, the screen reader would announce, "Health Equity Tracker. Health Equity Tracker logo decorative H", which is redundant.

frontend/src/App.tsx
```diff
@@ -141,34 +145,40 @@ function App() {
...
+                    component={TermsOfServicePage}
+                />
+                <Route exact path="/" component={LandingPage} />
+                <Route component={NotFoundPage} />
+              </Switch>
+            </main>
           </Router>
         </div>
-        <Footer />
+        <footer aria-label="Footer content">
```
justinpetersen, line 181: It is my understanding that you would not want an aria-label here. I think just using the <footer> element type is enough to provide the page structure to assistive technologies.

frontend/src/pages/ExploreData/ExploreDataPage.tsx
```diff
@@ -76,6 +76,7 @@ function ExploreDataPage() {
     <div id="ExploreData" className={styles.ExploreData}>
       <div className={styles.CarouselContainer}>
         <Carousel
+          aria-label="Mad lib selection carousel"
```
justinpetersen, line 79: I do not think you need an aria-label for the carousel container. I think just the back/forward buttons should have aria-labels, since they are interactive.

frontend/src/pages/ExploreData/ExploreDataPage.tsx
```diff
@@ -96,7 +97,8 @@ function ExploreDataPage() {
         </Carousel>
       </div>
       <div className={styles.ReportContainer}>
-        <ReportProvider madLib={madLib} setMadLib={setMadLib} />
+        <ReportProvider aria-label="Custom disparity report"
```
justinpetersen, line 100: I do not think you need an aria-label for this. If the section needs a heading, then I would use an <h2> tag and have this visible for all users. Otherwise I think the screen reader can just announce the text that is visible on the page.

frontend/src/App.tsx
```diff
@@ -60,50 +60,54 @@ function MobileAppToolbar() {
...
+        </IconButton>
+        <Drawer variant="persistent" anchor="left" open={open}>
+          <Button aria-label="Hide site navigation button"
+                  onClick={() => setOpen(false)}>
+            <ChevronLeftIcon />
+          </Button>
+          <List>
+            {Object.keys(PAGE_URL_TO_NAMES).map((pageUrl, index) => (
+                <ListItemLink href={pageUrl} key={index}
+                              aria-label={PAGE_URL_TO_NAMES[pageUrl]}>
```
justinpetersen: I do not think you need an aria-label here, as the screen reader should announce the text value instead.

frontend/src/App.tsx
```diff
@@ -141,34 +145,40 @@ function App() {
...
-                component={WhatIsHealthEquityPage}
-              />
-              <Route
-                  path={TERMS_OF_SERVICE_PAGE_LINK}
-                  component={TermsOfServicePage}
-              />
-              <Route exact path="/" component={LandingPage} />
-              <Route component={NotFoundPage} />
-            </Switch>
+            <header aria-label="Header content including site navigation">
```
justinpetersen, line 150: I do not think you need an aria-label here. Having <header> provides the page structure, and as long as the navigation links area announced, I think that is all that needs to be announced.

frontend/src/App.tsx
```diff
@@ -63,25 +60,25 @@ function MobileAppToolbar() {
...
-                              aria-label={PAGE_URL_TO_NAMES[pageUrl]}>
-                  <ListItemText primary={PAGE_URL_TO_NAMES[pageUrl]} />
-                </ListItemLink>
-            ))}
-          </List>
-        </Drawer>
-      </Toolbar>
+    <Toolbar>
+      <IconButton onClick={() => setOpen(true)}
+                  aria-label="Expand site navigation button">
```
justinpetersen, line 65: Remove "button" here. "Expand site navigation" is better (you want the screen reader to announce what the click will trigger in the form of an instruction).

frontend/src/App.tsx
```diff
@@ -63,25 +60,25 @@ function MobileAppToolbar() {
...
+        <MenuIcon className={styles.MenuIconForMobile} />
+      </IconButton>
+      <Drawer variant="persistent" anchor="left" open={open}>
+        <Button aria-label="Hide site navigation button"
```
justinpetersen, line 69: I would use "Collapse site navigation" here. "Hide" is also a fine verb, but I think expand/collapse is a more common convention. Leave out "button" so this reads like a command (the element being a <button> is enough to communicate its role).

kkatzen (approved):
awesome! thanks for doing these!

PR #565, opened 2021-04-08
Hide legend for MapCard when one result

### LINKED ISSUE #545: Maps with little/no data behaving oddly
Screen Shot 2021-04-07 at 1 08 17 PM

What is going on with the legend?

kkatzen:
Closes #545

Legends just act weird when there is only one line in the dataset (for instance looking at values for one county). This just hides the legend in those circumstances since it's pointless. See comparison: 

Screen Shot 2021-04-08 at 2 48 41 PM

Notice hover is still working.


vanshkumar (approved):
Thanks for fixing! This was one of the issue I noted in the bug bash :)

Issue #542, opened 2021-04-07
Audit data sources content


kkatzen:
Eng made up the text on the cards on this page: https://dev.healthequitytracker.org/datacatalog
Notice how the link below the title is the exact same text...

Product should create final copy

kkatzen:
Notice that the demographic granularity filter is based on the literal text... I'm not sure how useful this / it's maybe misleading since the granularity isn't all of those together but rather all of those separately... we may want to just remove

kkatzen:
From Larry: they are working on updating that content, we can go ahead and remove the expand (show more/less) toggle

kkatzen:
We'll remove all filters for now

vanshkumar:
I looked through the page, made a couple edits to the text, and downloaded each of the datasets and they all seemed usable to me

vanshkumar:
Chatted with Larry, we're good to go here except for the CDC restricted disclaimer (which is being tracked in a different issue)

PR #566, opened 2021-04-08
Updates to Data Download page

kkatzen:
Addresses part of #542

Removed filters and more/less expansion.

I also updated the title to make the style a little more cohesive? I can always change it back haha

Screen Shot 2021-04-08 at 3 15 43 PM

I did consider saving some of the existing filter code in case they wanted to add it back later, but the filters DatasetFilter class is still there so someone could always go to the original commit in github to revive it from the dead so I just deleted it.

### DIFF
--- a/frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.module.scss
+++ b/frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.module.scss
@@ -24,6 +24,7 @@
 
 .DatasetTitle {
   font-size: 20px;
+  font-family: "Taviraj", serif !important;
 }
 
 .Footer {
@@ -38,10 +39,6 @@
 
 .CardFooterRight {
   float: right;
-}
-
-.CardFooterLeft {
-  float: left;
 }
 
 .DataSubtitle {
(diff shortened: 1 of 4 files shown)


frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.module.scss
```diff
@@ -24,6 +24,7 @@
 
 .DatasetTitle {
   font-size: 20px;
+  font-family: "Taviraj", serif !important;
```
vanshkumar, line 27: Question not specific to this PR: is it possible to define this font globally somewhere and reference that? iirc we use this font all over the place

vanshkumar (approved):

frontend/src/pages/DataCatalog/dataset_explorer/DataSourceListing.module.scss
kkatzen: Yeah I'm actually kinda confused about the font scheme throughout the static pages... I'll file a FE backlog issue

kkatzen: https://github.com/SatcherInstitute/health-equity-tracker/issues/567

PR #577, opened 2021-04-09
Removed the Hispanic and White Non Hispanic from Totals as these are subsets.

Sethvg:


frontend/src/data/variables/AcsHealthInsuranceProvider.ts
```diff
@@ -90,7 +90,14 @@ class AcsHealthInsuranceProvider extends VariableProvider {
 
     // Calculate totals where dataset doesn't provide it
     // TODO- this should be removed when Totals come from the Data Server
-    const total = df.pivot(["fips", "fips_name"], totalPivot).resetIndex();
+    const total = df
```
vanshkumar, line 93: It would be good to add a comment describing why we remove these categories in this calculation

frontend/src/data/variables/AcsPovertyProvider.ts
```diff
@@ -50,6 +57,11 @@ class AcsPovertyProvider extends VariableProvider {
     // Calculate totals where dataset doesn't provide it
     // TODO- this should be removed when Totals come from the Data Server
     const calculatedValueForAll = df
+      .where(
```
vanshkumar, line 60: Ditto for adding a comment here about why these are removed

Also, did you mean to exclude HISPANIC rather than TWO_OR_MORE?

Sethvg: Done

frontend/src/data/variables/AcsHealthInsuranceProvider.ts
Sethvg: Done

vanshkumar:
Looks like build is failing?

aaronsn (approved):

vanshkumar (approved):

PR #580, opened 2021-04-11
Health insurance disparities

Sethvg:


frontend/src/charts/storybook/DisparityBarChart.stories.tsx
```diff
@@ -1,41 +0,0 @@
-import React from "react";
-import { Story, Meta } from "@storybook/react/types-6-0";
```
vanshkumar, line 2: Why are you deleting the DisparityBarChart story?

Sethvg: Added back.  I think it was an empty file when I deleted it.

vanshkumar (approved):
FYI that you have the same issue as the BRFSS data in that racial categories for which we have no data are still showing up in the table card ([example from your deploy preview](https://screenshot.googleplex.com/5Ljrx8hQrFFj5kU)).. this issue has details & a best guess for why this is happening: https://github.com/SatcherInstitute/health-equity-tracker/issues/593.

Josh is looking into it & we probably will have to apply the same fix he does for BRFSS to ACS health insurance / poverty data

joshzarrabi:
@Sethvg I noticed that the text on the table card is wrong. It still says `Share of Health Insurance` But I think it should be `Share of uninsured Americans` or something like that

PR #581, opened 2021-04-11
Poverty Disparity Breakdown

Sethvg:


joshzarrabi:
@Sethvg Looks good! Do you think you could apply the same fix as we did here: https://github.com/SatcherInstitute/health-equity-tracker/pull/613, so that the table only shows races we have values for?

vanshkumar (approved):
I'm also OK with making the population_pct change being discussed (to show only demos we have data for) in a separate PR, I'm happy to do it as well

PR #582, opened 2021-04-11
Only show known demographics column in table card

vanshkumar:


kkatzen (approved):
Looks good! I probably could have been more clear in MetricConfig what knownBreakdownComparisonMetric was- you could maybe add a comment explaining it?

vanshkumar:
> Looks good! I probably could have been more clear in MetricConfig what knownBreakdownComparisonMetric was- you could maybe add a comment explaining it?

added a comment

PR #584, opened 2021-04-12
Added age breakdown for ACS health insurance

Sethvg:


frontend/src/data/variables/AcsHealthInsuranceProvider.ts
```diff
@@ -16,7 +16,7 @@ class AcsHealthInsuranceProvider extends VariableProvider {
   }
 
   getDatasetId(breakdowns: Breakdowns): string {
-    if (breakdowns.hasOnlySex()) {
+    if (breakdowns.hasOnlySex() || breakdowns.hasOnlyAge()) {
       return breakdowns.geography === "county"
         ? "acs_health_insurance-health_insurance_by_sex_county"
```
vanshkumar, line 21: It's a bit confusing that the name for this dataset does not include age but we are using it to get age - can you either update the names of these "by_sex" datasets to include age or at minimum add a comment about this?

vanshkumar (approved):

frontend/src/data/variables/AcsHealthInsuranceProvider.ts
Sethvg: Changed the dataset name

PR #587, opened 2021-04-13
Calculate the national percent share for COPD and Diabetes

### LINKED ISSUE #515: Investigate why BRFSS datasets aren't displaying properly
COPD and diabetes, the table isn't populating "share of diabetes cases"

let's investigate if 
1) we're actually pulling this data from the backend
2) why share of diabetes cases isn't working

joshzarrabi:
Resolves #515

### DIFF
--- a/frontend/src/data/utils/datasetutils.ts
+++ b/frontend/src/data/utils/datasetutils.ts
@@ -144,13 +144,15 @@ export function percent(numerator: number, denominator: number): number | null {
 
 /** Finds expected value of an ailment based on a population sample. */
 export function estimateTotal(
-  numerator: number,
-  denominator: number,
-  total: number
+  sample_count: number,
+  sample_population: number,
+  total_population: number
 ): number {
-  return numerator == null || denominator == null || denominator === 0
+  return sample_count == null ||
+    sample_population == null ||
+    sample_population === 0
     ? 0
-    : Math.round((numerator / denominator) * total);
+    : Math.round((sample_count / sample_population) * total_population);
 }
 
 export function asDate(dateStr: string) {
--- a/frontend/src/data/utils/datasetutils.ts
+++ b/frontend/src/data/utils/datasetutils.ts
@@ -142,7 +142,7 @@ export function percent(numerator: number, denominator: number): number | null {
     : Math.round((1000 * numerator) / denominator) / 10;
 }
 
-/** Divides two numbers. */
+/** Finds expected value of an ailment based on a population sample. */
 export function estimateTotal(
   numerator: number,
   denominator: number,
--- a/frontend/src/data/utils/datasetutils.ts
+++ b/frontend/src/data/utils/datasetutils.ts
@@ -142,6 +142,17 @@ export function percent(numerator: number, denominator: number): number | null {
     : Math.round((1000 * numerator) / denominator) / 10;
 }
 
+/** Divides two numbers. */
+export function estimateTotal(
+  numerator: number,
+  denominator: number,
+  total: number
+): number {
+  return numerator == null || denominator == null || denominator === 0
+    ? 0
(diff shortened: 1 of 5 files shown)


frontend/src/data/utils/datasetutils.ts
```diff
@@ -142,6 +142,17 @@ export function percent(numerator: number, denominator: number): number | null {
     : Math.round((1000 * numerator) / denominator) / 10;
 }
 
+/** Finds expected value of an ailment based on a population sample. */
+export function estimateTotal(
+  numerator: number,
```
vanshkumar, line 147: Might be clearer to rename the arguments to this function to sample_count, sample_population, and total_population

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -29,7 +39,48 @@ class BrfssProvider extends VariableProvider {
...
+
+    let consumedDatasetIds = ["brfss"];
+
     if (breakdowns.geography === "national") {
+      // Because we add together the estimated state totals
+      // we need to get the acs state breakdown
+      acsBreakdowns.geography = "state";
+
+      const acsQueryResponse = await this.acsProvider.getData(
+        new MetricQuery(["population", "population_pct"], acsBreakdowns)
```
vanshkumar, line 54: Do you actually use population_pct from the ACS provider anywhere? If not, you can just not request this metric

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -29,7 +39,48 @@ class BrfssProvider extends VariableProvider {
...
     df = this.renameGeoColumns(df, breakdowns);
 
+    // TODO How to handle territories?
+    let acsBreakdowns = breakdowns.copy();
+    acsBreakdowns.time = false;
+
+    let consumedDatasetIds = ["brfss"];
+
     if (breakdowns.geography === "national") {
+      // Because we add together the estimated state totals
```
vanshkumar, line 49: IMO a more descriptive comment on how the logic for this case works would be useful (e.g. BRFSS is a survey, sampling rates are diff across states, so we have to compute estimated state totals first)

vanshkumar (approved):
looks good!

frontend/src/data/utils/datasetutils.ts
joshzarrabi: For sure

frontend/src/data/variables/BrfssProvider.ts
joshzarrabi: Yeah I think I can get rid of it, thanks

frontend/src/data/variables/BrfssProvider.ts
joshzarrabi: Hundo p

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -29,7 +39,48 @@ class BrfssProvider extends VariableProvider {
...
+        estimated_total_copd: (row) =>
+          estimateTotal(
+            row.copd_count,
+            row.copd_count + row.copd_no,
+            row.population
+          ),
+      });
+
       df = df
         .pivot(breakdowns.demographicBreakdowns.race_and_ethnicity.columnName, {
```
vanshkumar, line 85: It's a bit odd how breakdowns.getSoleDemographicBreakdown().columnName is used in some places whereas breakdowns.demographicBreakdowns.race_and_ethnicity.columnName is used in others. Can you standardize this usage by defining a breakdownColumnName variable near the start of the function which is set to breakdowns.getSoleDemographicBreakdown().columnName, and using that throughout? That's what we do in other providers

vanshkumar (approved):

Issue #594, opened 2021-04-14
Don't show "White (non-hispanic)" as a race category for ACS poverty & health insurance data


vanshkumar:
Right now we show both "White" and "White (non-hispanic)" as race categories for ACS poverty and health insurance. After chatting with the SHLI team, the suggestion is to not show "White (non-hispanic)" and only show "White," as it is somewhat confusing and may look like we are singling out one specific race to break down further on ethnicity.

PR #595, opened 2021-04-14
Removed white hispanic from races

Sethvg:


frontend/src/data/variables/AcsHealthInsuranceProvider.ts
```diff
@@ -77,6 +77,14 @@ class AcsHealthInsuranceProvider extends VariableProvider {
       );
     }
 
+    //Remove white hispanic to bring inline with others
```
vanshkumar, line 80: Consider linking to the issue (https://github.com/SatcherInstitute/health-equity-tracker/issues/594)  in the comment here as a way for readers to see more detail

frontend/src/data/variables/AcsPovertyProvider.ts
```diff
@@ -53,13 +53,21 @@ class AcsPovertyProvider extends VariableProvider {
         .resetIndex();
     }
 
+    
+    //Remove white hispanic to bring inline with others
```
vanshkumar, line 57: ditto here

vanshkumar (approved):

PR #597, opened 2021-04-14
Update screenshots on Landing page

### LINKED ISSUE #589: Update screenshots in the How do I use the Tracker on home page
Update the placeholder images.

efregoso:
Closes #589


vanshkumar (approved):
Looks good, but I wonder if we should wait to update these till later (or update them again later) so that we can definitively have screenshots of the latest version of the site when launching

PR #598, opened 2021-04-14
Make 10 rows/page default and expand height of TableChart

### LINKED ISSUE #578: Expand Cases by Race/Ethnicity card
Task force members indicated they were concerned about groups being hidden on page 2 of the table, or beneath the fold within a scrolling window in window.

Expand the card to show 10 rows by default without requiring scroll so that fewer groups require the user to go to page 2 of the table

efregoso:
Closes #578


vanshkumar:
I think we should check with Chi / perhaps run this by a wider audience, as it looks pretty odd to have just this one card be significantly longs than the others.. could you start the conversation on this in slack?

adams314:
This looks ok to me and is better than the alternative of hiding several rows. We can run it by the group during the meeting today

vanshkumar (approved):

PR #607, opened 2021-04-15
Extract Unknown Dialog into top-level card

### LINKED ISSUE #602: Extract unknown map into its own card
Take the warning and the unknowns map card that's currently in the disparity bar chart and pull it into its own card

kkatzen:
Addresses #602 

Screen Shot 2021-04-15 at 1 35 27 PM
Screen Shot 2021-04-15 at 1 36 03 PM
Screen Shot 2021-04-15 at 1 36 19 PM
Screen Shot 2021-04-15 at 1 36 28 PM
Screen Shot 2021-04-15 at 1 39 34 PM


kkatzen:
This PR doesn't close the attached issue, we'll still need to revisit the alerts but my 20% time is over for the week so I'd like to submit this and then follow up with the data warnings next week

vanshkumar (approved):

Issue #593, opened 2021-04-14
Diabetes/COPD Cases by Race and Ethnicity cards have both race and race (Non-Hispanic) categories


vanshkumar:
We should only show the races for which we actually have data. Based on looking through the code, the best guess for why this is happening is that the populationComparisonMetric in the MetricConfig is coming from ACS population provider, whereas the raw metric (e.g. diabetes_count) comes from the BRFSS provider. Simplest fix is to join with ACS pop for all cases in the BRFSS provider (right now we only do it for national geo) & use the resulting population metric as the populationComparisonMetric, so we are sure to only include races for which we have data.

vanshkumar:
This issue is popping up for the ACS health insurance data, and presumably will for poverty as well (see [this comment](https://github.com/SatcherInstitute/health-equity-tracker/pull/580#pullrequestreview-636121309)). We likely will want to apply whatever fix we find for the BRFSS data to those as well

joshzarrabi:
Done in https://github.com/SatcherInstitute/health-equity-tracker/pull/613

PR #613, opened 2021-04-15
only show demographics we have stats for in brfss table card

joshzarrabi:
Resolves: https://github.com/SatcherInstitute/health-equity-tracker/issues/593

### DIFF
--- a/frontend/src/data/config/MetricConfig.ts
+++ b/frontend/src/data/config/MetricConfig.ts
@@ -2,6 +2,7 @@
 export type MetricId =
   | "population"
   | "population_pct"
+  | "my_population_pct"
   | "diabetes_count"
   | "diabetes_per_100k"
   | "diabetes_pct_share"
@@ -261,20 +262,17 @@ export const METRIC_CONFIG: Record<string, VariableConfig[]> = {
       variableDisplayName: "Cases",
       variableFullDisplayName: "Diabetes Cases",
       metrics: {
-        count: {
-          metricId: "diabetes_count",
-          fullCardTitleName: "Count of diabetes cases",
-          shortVegaLabel: "Diabetes cases",
-          type: "count",
-          populationComparisonMetric: POPULATION_VARIABLE_CONFIG.metrics.count,
-        },
         pct_share: {
           metricId: "diabetes_pct_share",
           fullCardTitleName: "Share of Diabetes cases",
           shortVegaLabel: "% of cases",
           type: "pct_share",
-          populationComparisonMetric:
-            POPULATION_VARIABLE_CONFIG.metrics.pct_share,
+          populationComparisonMetric: {
+            metricId: "my_population_pct",
+            fullCardTitleName: "Population Share",
+            shortVegaLabel: "% of total population",
+            type: "pct_share",
+          },
         },
         per100k: {
           metricId: "diabetes_per_100k",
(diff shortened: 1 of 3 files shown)


frontend/src/data/config/MetricConfig.ts
```diff
@@ -2,6 +2,7 @@
 export type MetricId =
   | "population"
   | "population_pct"
+  | "my_population_pct"
```
vanshkumar, line 5: I wonder if a better name for this might be "brfss_population_pct" ?

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -48,28 +49,25 @@ class BrfssProvider extends VariableProvider {
     let consumedDatasetIds = ["brfss"];
 
     if (breakdowns.geography === "national") {
```
vanshkumar, line 51: It would be good to add a short comment explaining why you set the acs breakdown to "state" if it is "national"

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -98,8 +96,26 @@ class BrfssProvider extends VariableProvider {
           estimated_total_diabetes: (series) => series.sum(),
         })
         .resetIndex();
+
+      // We need to get the national acs dataset here in order to
+      // get the national pct share of population for each state.
+      acsBreakdowns.geography = "national";
```
vanshkumar, line 102: Would be good to add a TODO here to see if we can convert these two calls to ACS provider into a single call

vanshkumar: Or actually, we want to do this on the BE eventually, so adding that as the TODO would be better

joshzarrabi: Yeah I'll add that, thanks

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -159,6 +176,7 @@ class BrfssProvider extends VariableProvider {
     df = df
       .dropSeries([
         "population",
+        "population_pct",
```
kkatzen, line 179: if you're renaming it, it shouldn't still be there to drop, right?

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -98,8 +96,26 @@ class BrfssProvider extends VariableProvider {
           estimated_total_diabetes: (series) => series.sum(),
         })
         .resetIndex();
+
+      // We need to get the national acs dataset here in order to
+      // get the national pct share of population for each state.
+      acsBreakdowns.geography = "national";
+      const acsQueryResponse = await this.acsProvider.getData(
```
kkatzen, line 103: Can we name acsQueryResponse here to distinguish it from that on line 55? acsNationalQueryResponse and acsStateQueryResponse?

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -98,8 +96,26 @@ class BrfssProvider extends VariableProvider {
...
+        new MetricQuery(["population_pct"], acsBreakdowns)
+      );
+      consumedDatasetIds = consumedDatasetIds.concat(
+        acsQueryResponse.consumedDatasetIds
+      );
+
+      const acsPopulation = new DataFrame(acsQueryResponse.data);
```
kkatzen, line 110: nit: since acsPopulation is only used once, i might just inline it in the joinOnCols(). If you didn't want to inline it I would at least remove the line break between them

I might also rename the two acsPopulations to be specific to the breakdowns that queried for them as well

frontend/src/data/config/MetricConfig.ts
kkatzen: +1

frontend/src/data/variables/BrfssProvider.ts
joshzarrabi: Right, good call

joshzarrabi:
@vanshkumar @kkatzen I think I addressed all of your comments, thank you!

vanshkumar (approved):

PR #629, opened 2021-04-20
For ACS health insurance, only show demographic groups we have data for

vanshkumar:
Same fix as Josh recently made for BRFSS data in #613. We'll have to do the same thing for ACS poverty once the pct_share calculations are added there (Seth has a PR for this in-progress).

### DIFF
--- a/frontend/src/data/config/MetricConfig.ts
+++ b/frontend/src/data/config/MetricConfig.ts
@@ -2,13 +2,13 @@
 export type MetricId =
   | "population"
   | "population_pct"
-  | "brfss_population_pct"
   | "diabetes_count"
   | "diabetes_per_100k"
   | "diabetes_pct_share"
   | "copd_count"
   | "copd_per_100k"
   | "copd_pct_share"
+  | "brfss_population_pct"
   | "covid_cases"
   | "covid_deaths"
   | "covid_hosp"
@@ -30,6 +30,7 @@ export type MetricId =
   | "health_insurance_count"
   | "health_insurance_per_100k"
   | "health_insurance_pct_share"
+  | "health_insurance_population_pct"
   | "poverty_count"
   | "poverty_per_100k";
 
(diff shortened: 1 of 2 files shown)


vanshkumar:
Also, I was testing the numbers out from this vs what we have in the tracker currently, and they all look pretty similar except for "Some other race" which has a higher value after this change than in the tracker today. IMO we should just move to joining ACS health insurance (and poverty) with ACS population, as we do for other data sources, rather than taking the total values in those datasets as equal to the population. It seems like that is true for most groups, but not all.

aaronsn (approved):

aaronsn:
> Also, I was testing the numbers out from this vs what we have in the tracker currently, and they all look pretty similar except for "Some other race" which has a higher value after this change than in the tracker today. IMO we should just move to joining ACS health insurance (and poverty) with ACS population, as we do for other data sources, rather than taking the total values in those datasets as equal to the population. It seems like that is true for most groups, but not all.

Yeah I noticed that total numbers are slightly lower for ACS variables like health insurance and poverty than for population - maybe they exclude some results there where the answers were unknown? Agreed that using ACS population is more accurate

vanshkumar:
> > Also, I was testing the numbers out from this vs what we have in the tracker currently, and they all look pretty similar except for "Some other race" which has a higher value after this change than in the tracker today. IMO we should just move to joining ACS health insurance (and poverty) with ACS population, as we do for other data sources, rather than taking the total values in those datasets as equal to the population. It seems like that is true for most groups, but not all.
> 
> Yeah I noticed that total numbers are slightly lower for ACS variables like health insurance and poverty than for population - maybe they exclude some results there where the answers were unknown? Agreed that using ACS population is more accurate

Actually I realized that the differences were smaller than what I had originally thought (mixed up two rows).. but still we should move to ACS population for these so that all of the numbers across the site are consistent

PR #631, opened 2021-04-21
only show unknown map if variable has a pct_share metricConfig

joshzarrabi:


joshzarrabi:
Merging without a review as a quick fix to the dev site

vanshkumar (commented):
nice job finding this so quickly! fix lgtm

PR #632, opened 2021-04-21
Age Accumulation Buckets

Sethvg:


vanshkumar:
Looks like build is breaking, could you fix that?

vanshkumar:
Before we start combining buckets to make the custom buckets requested by MSM, it would be worthwhile to spend a bit of time figuring out why the age "16" doesn't show up in the data.. is there a bug in the parsing code somewhere? Does the data itself just not include it?

python/datasources/acs_poverty.py
```diff
@@ -46,6 +48,44 @@
...
+def remove_suffix(input_string, suffix):
+    if suffix and input_string.endswith(suffix):
+        return input_string[: -len(suffix)]
+    return input_string
+
+
+def parseAgeString(in_str):
+    if in_str.endswith("+"):
+        return {"min": int(remove_suffix(in_str, "+")), "max": sys.maxsize}
+    else:
```
joshzarrabi, line 67: Don't need this `else` I think

python/datasources/acs_poverty.py
```diff
@@ -153,6 +215,33 @@ def getData(self, gcs_bucket=None):
...
+    def custom_accumulations(self):
+        new_data = {}
+
+        for data, population in self.data.items():
+            state_fip, county_fip, age, sex, race = data
+
+            above = int(population[PovertyPopulation.ABOVE])
+            below = int(population[PovertyPopulation.BELOW])
+
+            assert above != -1
```
joshzarrabi, line 227: I don't really like the idea of having `assert` statements in our code. I would prefer checking the value and `raising` an error if it is wrong.

joshzarrabi (changes requested):

joshzarrabi (approved):

PR #644, opened 2021-04-22
Update Unknown card title + grey

### LINKED ISSUE #642: Remove Geo from card titles that have geo navigation
The map cards have a row that provides the context of the locale the user is viewing. This means the text of the locale is duplicated. Let's remove the extra text from the title

image

kkatzen:
Screen Shot 2021-04-22 at 11 00 07 AM
Closes #642

### DIFF
--- a/frontend/src/cards/UnknownsMapCard.tsx
+++ b/frontend/src/cards/UnknownsMapCard.tsx
@@ -66,11 +66,9 @@ function UnknownsMapCardWithKey(props: UnknownsMapCardProps) {
     <CardWrapper
       queries={[query]}
       title={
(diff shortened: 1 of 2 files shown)


vanshkumar:
Quick question about the copy for the title, are we sure we want to change from what we have now ("Share of total COVID-19 cases: Unknown Race and Ethnicity") to what you have here?

kkatzen:
> Quick question about the copy for the title, are we sure we want to change from what we have now ("Share of total COVID-19 cases: Unknown Race and Ethnicity") to what you have here?

Confirmed with larry

vanshkumar (approved):

PR #652, opened 2021-04-22
Adding report level disclaimer

### LINKED ISSUE #638: Add page level warning and reference to missing data explanations at bottom of Explore Data page
This is a screenshot of a page hacked up to show where the alert should go.

image

Use the following text and formatting for the alert:

Structural racism and oppression yield health inequities. Missing data plagues our understanding of the incidence and prevalence of cases, which in turn prevents the creation of swift and effective protocols to protect against disease and promote health in the United States. 

The maps and tables below reflect the best data we have but there are major known gaps in the data. [Read more about missing and misidentified people.](#linktobottomofpage)

#615

kkatzen:
Closes #638

### DIFF
--- a/frontend/src/reports/ui/DisclaimerAlert.tsx
+++ b/frontend/src/reports/ui/DisclaimerAlert.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import Alert from "@material-ui/lab/Alert";
+import AlertTitle from "@material-ui/lab/AlertTitle";
 import Button from "@material-ui/core/Button";
 import styles from "./DisclaimerAlert.module.scss";
 import FlagIcon from "@material-ui/icons/Flag";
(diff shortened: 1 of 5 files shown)


frontend/src/reports/ui/DisclaimerAlert.tsx
```diff
@@ -0,0 +1,35 @@
...
+      severity="warning"
+      className={styles.ReportAlert}
+      icon={<FlagIcon />}
+    >
+      <AlertTitle>Major gaps in the data</AlertTitle>
+      <p>
+        <span style={{ lineHeight: "21px" }}>
+          Structural racism and oppression create health inequities, which leads
+          to missing data. The maps and tables below reflect the best data we
+          have, but there are major known gaps in the data. We're working to
```
vanshkumar, line 20: Do we care that we are using the word "major" twice (in the title and in the text here)?

vanshkumar (approved):

frontend/src/reports/ui/DisclaimerAlert.tsx
kkatzen: Chi and Larry both okay'ed this copy so I'll go ahead and submit. Easy to change later! I think it wouldn't hurt to have a "copy bash" before launch for all these tweaks

PR #653, opened 2021-04-22
Compare unknown order

### LINKED ISSUE #649: Change order of compare report
first card should be incidence, second be unknowns map

kkatzen:
Closes #649
I also noticed that the compare reports are getting mis-aligned if certain metrics are missing so i adjusted that. See screenshots:
Screen Shot 2021-04-22 at 2 13 04 PM
Screen Shot 2021-04-22 at 2 13 11 PM
Screen Shot 2021-04-22 at 2 13 18 PM
Screen Shot 2021-04-22 at 2 13 32 PM

### DIFF
--- a/frontend/src/reports/TwoVariableReport.tsx
+++ b/frontend/src/reports/TwoVariableReport.tsx
@@ -11,11 +11,13 @@ import { DropdownVarId } from "../utils/MadLibs";
 import { Fips } from "../data/utils/Fips";
 import {
   METRIC_CONFIG,
+  MetricConfig,
   VariableConfig,
   getPer100kAndPctShareMetrics,
 } from "../data/config/MetricConfig";
 import ReportToggleControls from "./ui/ReportToggleControls";
 import NoDataAlert from "./ui/NoDataAlert";
+import Alert from "@material-ui/lab/Alert";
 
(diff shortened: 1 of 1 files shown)


frontend/src/reports/TwoVariableReport.tsx
```diff
@@ -168,47 +171,93 @@ function TwoVariableReport(props: {
...
+              )}
+            />
           </>
         )
       )}
     </Grid>
   );
 }
 
+function RowOfOptionalMetrics(props: {
```
vanshkumar, line 215: Would a better name for this be RowOfTwoOptionalMetrics to make it clear it's only for two variables?

vanshkumar (approved):

vanshkumar:
Also FYI that pct_share will be added in for poverty very soon (in PR) so the missing metrics alignment issue shouldn't be a problem after that

frontend/src/reports/TwoVariableReport.tsx
kkatzen: ya good idea

kkatzen:
Ya I did realize that poverty pct_share would be added soon but I figure this probably won't be the last time some weird edge resurfaces this issue I might as well fix it

PR #659, opened 2021-04-22
Capitalize uninsured to match other data & consolidate terminology to "uninsured individuals"

vanshkumar:


vanshkumar:
This is what it looks like before this PR:

image

kkatzen:
Should we also consider consolidating the three variants: "Uninsured individuals", "Uninsured Americans", "Uninsured people" into the same copy?

kkatzen (approved):

vanshkumar:
> Should we also consider consolidating the three variants: "Uninsured individuals", "Uninsured Americans", "Uninsured people" into the same copy?

I like it, will bring this up in slack as well

PR #672, opened 2021-04-23
Update a bit of copy

vanshkumar:


kkatzen (approved):

kkatzen:
I think you've duplicated this in https://github.com/SatcherInstitute/health-equity-tracker/pull/671

vanshkumar:
> I think you've duplicated this in #671

Oops yea good catch will revert from this one

PR #680, opened 2021-04-24
UnknownsAlert

### LINKED ISSUE #602: Extract unknown map into its own card

kkatzen:
closes #602

### DIFF
--- a/frontend/src/cards/DisparityBarChartCard.tsx
+++ b/frontend/src/cards/DisparityBarChartCard.tsx
@@ -84,6 +84,8 @@ function DisparityBarChartCardWithKey(props: DisparityBarChartCardProps) {
               metricConfig={props.metricConfig}
               queryResponse={queryResponse}
               breakdownVar={props.breakdownVar}
+              displayType="chart"
+              known={true}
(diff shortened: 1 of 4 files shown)


vanshkumar:
Looks like the unknown map card still says "The chart below only displays data for cases where race and ethnicity was known" - I think it should be "unknown". Also, do we want to change "chart" to "map" for this card? IMO that makes a bit more sense and also helps distinguish the warnings slightly more

kkatzen:
Good call on updating text... I thought about moving the Alert out of a component into the places themselves but thought it might be nice to have the logic for computing the value in one place. Plus it will make it easier to add it to new cards later...

Screen Shot 2021-04-26 at 5 48 09 PM

vanshkumar:
Thanks for update, lgtm. One note - it would be good to run this by users (perhaps HETF) to see if they can parse these two alerts and figure out what these cards are actually showing, I could see it being a bit confusing

vanshkumar (approved):

PR #682, opened 2021-04-26
Use new poverty_pop_pct metric

vanshkumar:


vanshkumar:
For some reason in the deploy preview the "Population Share" column shows "No Data".. is there an issue with the calculation of the poverty_population_pct metric? Perhaps a bug in #581 ?

Sethvg:
This shouldn't be necessary anymore.

Sethvg:
Also, just as an fyi, the reason this doesnt work is because 1) you need to add the new metricId to the types, and 2) you also need to add the calculation to the provider :)

vanshkumar:
> Also, just as an fyi, the reason this doesnt work is because 1) you need to add the new metricId to the types, and 2) you also need to add the calculation to the provider :)

Hmm what do you mean? I thought your latest PR (#581) already did those things? This PR has those changes merged already

I don't think your latest PR fixed the issue as on the dev site, you can still see all race/ethnicities, even those for which we don't have data (screenshot)

Screen Shot 2021-04-26 at 12 34 08 PM

vanshkumar:
Figured it out, it was because there was no "total" column being calculated in the poverty provider before it was being used, fixed

Sethvg (approved):

PR #684, opened 2021-04-26
Use _RACE category from brfss survey 

### LINKED ISSUE #591: Use proper `_RACE` race categorization with BRFSS data instead of `Imputed Race`
### Background

The BRFSS survey has a few different ways of categorizing race, all defined [here](https://www.cdc.gov/brfss/annual_data/2019/pdf/codebook19_llcp-v2-508.HTML).

Currently we use the `Imputed Race` category, which is defined as: `Imputed race/ethnicity value  (This value is the reported race/ethnicity or an imputed race/ethnicity, if the respondent refused to give a race/ethnicity. The value of the imputed race/ethnicity will be the most common race/ethnicity response for that region of the state)`.

It allows for the following race categorizations:

* White, Non-Hispanic
* Black, Non-Hispanic
* Asian, Non-Hispanic
* American Indian/Alaskan Native, Non-Hispanic
* Hispanic
* Other race, Non-Hispanic

### Problems

There are a few issues with using the `Imputed Race` category

* It is unclear how multi-race people are counted.
* There is no category for "Pacific Islander and Native Hawaiian", which the other data sets we use do have.
* It lacks a more detailed description in the [computed values](https://www.cdc.gov/brfss/annual_data/2019/pdf/2019-calculated-variables-version4-508.pdf) document, which describes in detail what other race categories mean.
* The fact that values are imputed with the most common race/ethnicity for that region of the state is a questionable way to guess and could lead to over or underrepresentation of demographics.  

### Proposed Solution

We should switch to using the `_RACE` column in the BRFSS data because it ....

* Properly categorizes multi-racial people.
* Has a category for "Native Hawaiian and Pacific Islander".
* Does not attempt go guess or "impute" anyone's racial category.
* Is properly defined, as shown below

Screen Shot 2021-04-12 at 4 18 12 PM

joshzarrabi:
* Gives us Asian American and Pacific Islander and Unknown categories 

[Resolves #591]


frontend/src/data/utils/datasetutils.ts
```diff
@@ -144,10 +144,39 @@ export function percent(numerator: number, denominator: number): number | null {
 
 /** Finds expected value of an ailment based on a population sample. */
 export function estimateTotal(
+  sRow: Row,
```
Sethvg, line 147: Nit: abbr is confusing.  Not quite sure what sRow is.

frontend/src/data/utils/datasetutils.ts
```diff
@@ -144,10 +144,39 @@ export function percent(numerator: number, denominator: number): number | null {
...
+      total_population = state_population.population;
+    }
+
+    const sample_population = stateTotals
+      .where((row) => row.fips === sRow.fips)
+      .resetIndex()
+      .at(0).total_sample_size;
+
+    return sample_count == null ||
+      sample_population == null ||
```
Sethvg, line 174: I *think* this can be simplified by using bool coalescence.

 !!sample_population ? Math.round(((sample_count ?? 0)/ sample_population) * total_population) : 0

frontend/src/data/variables/BrfssProvider.test.ts
```diff
@@ -164,9 +170,39 @@ describe("BrfssProvider", () => {
     const rawAcsData = [AL_ACS_ASIAN_ROW, NC_ACS_ASIAN_ROW, NC_ACS_WHITE_ROW];
 
     // Create final rows with diabetes_count & diabetes_per_100k
-    const NC_ASIAN_FINAL = finalRow(NC, RACE, ASIAN_NH, 400, 40000, 16.7, 40);
-    const NC_WHITE_FINAL = finalRow(NC, RACE, WHITE_NH, 600, 60000, 83.3, 60);
-    const NC_ALL_FINAL = finalRow(NC, RACE, ALL, 1000, 50000, 100, 100);
+    const NC_ASIAN_FINAL = finalRow(
```
Sethvg, line 173: Nit: I know it was like this before, and I also should do this for the other providers, but can you add comments on the params for the numbers for easier readability?

Sethvg (approved):

frontend/src/data/variables/BrfssProvider.ts
```diff
@@ -173,6 +230,26 @@ class BrfssProvider extends VariableProvider {
...
+          [
+            "copd_count_share_of_known",
+            "diabetes_count_share_of_known",
+          ].includes(metricId)
+        );
+
+        shareOfUnknownMetrics.forEach((shareOfUnknownColumnName) => {
+          const rawCountColunn = shareOfUnknownColumnName.slice(
+            0,
+            -"_count_share_of_known".length
```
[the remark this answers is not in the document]
joshzarrabi: It's done like this in other places, it might be good to keep it consistent

PR #705, opened 2021-04-29
Cards take VariableConfig instead of MetricConfig

kkatzen:
Currently cards take a MetricConfig instead of a VariableConfig. This change has the cards take a VariableConfig and decide themselves which Metric they want.

Reasoning
1) To date, each card is only used with the same metric types each time it is called.
2) Reducing amount of specifications required when calling a card- we can anticipate less misconfigurations if someone tries to use a card with an unsupported metric type. When working on the card, don't have to worry about catching misconfiguration behaviors.
3) Some Cards need multiple metrics (table does now and soon map card will)- it seems neater to pass the entire VariableConfig and have the Cards select which metrics they want off of it rather than passing each seperately.


vanshkumar (approved):
This is a good cleanup, thanks!

PR #707, opened 2021-04-29
Change unknowns map card color

### LINKED ISSUE #683: Look into doing a crosshatch overlay for TX, LA, WY, NH in covid unknowns map
Unknowns map looks weird for these states for which we are suppressing all data, as they show up as light gray. We should see if it's possible to do a crosshatch overlay / generally how hard it is to show different styles on the map for these specific states in Vega

image

kkatzen:
Closes #683

Screen Shot 2021-04-29 at 11 38 16 AM


adams314 (approved):
LGTM

PR #722, opened 2021-04-30
Split ACS Poverty into unique Datasets

Sethvg:


python/datasources/acs_poverty.py
```diff
@@ -311,80 +321,165 @@ def accumulate_acs_data(self, data):
...
-                        state_fip,
-                        self.state_fips[state_fip],
-                        race,
-                        age,
-                        sex,
-                        above,
-                        below,
-                    ]
-                )
+                if race is not None:
```
joshzarrabi, line 345: I think for these you can just do `if race:`, `if sex:` `if age:`

Sethvg: Done

python/datasources/acs_poverty.py
```diff
@@ -311,80 +321,164 @@ def accumulate_acs_data(self, data):
...
-        # (state_fip, County_fip, Age, Sex, Race): {PopulationObj}
+        # (state_fip, county_fip, Age, Sex, Race): {PopulationObj}
         for data, population in self.data.items():
             state_fip, county_fip, age, sex, race = data
 
             population = self.data[data]
-            above = population[PovertyPopulation.ABOVE]
             below = population[PovertyPopulation.BELOW]
 
+            if (age is not None) + (sex is not None) + (race is not None) != 1:
```
vanshkumar, line 340: This is a bit of an odd construction, can you add a comment explaining that this triggers if not exactly one of {race, sex, age} are not None?

python/datasources/acs_poverty.py
```diff
@@ -311,80 +321,164 @@ def accumulate_acs_data(self, data):
...
-                        state_fip,
-                        self.state_fips[state_fip],
-                        race,
-                        age,
-                        sex,
-                        above,
-                        below,
-                    ]
-                )
+                if race:
```
vanshkumar, line 344: In general, it's preferred to compare explicitly against None, e.g. if race is not None

Ditto for all of the comparisons like this below

Details: https://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none

python/datasources/acs_poverty.py
```diff
@@ -311,80 +321,164 @@ def accumulate_acs_data(self, data):
...
+                    race_county_data.append(
+                        [
+                            state_fip,
```
vanshkumar, line 375: Most of the array you append is basically the same for all of the clauses.. can you pull it out into a variable above and just add to that, e.g. race_county_data.append(county_base + [race, below]) or the like? Same for the state data in the section above

Sethvg: Done

python/datasources/acs_poverty.py
Sethvg: Okay I had that origionally but Josh asked me to remove it so I added it back

python/datasources/acs_poverty.py
Sethvg: Done

vanshkumar (approved):

PR #723, opened 2021-04-30
new datasets

Sethvg:


frontend/src/data/variables/AcsPovertyProvider.ts
```diff
@@ -24,9 +24,21 @@ class AcsPovertyProvider extends VariableProvider {
   }
   // ALERT! Make sure you update DataSourceMetadata if you update dataset IDs
   getDatasetId(breakdowns: Breakdowns): string {
+    let datasetPrefix = "acs_poverty_dataset-poverty_by_";
+
+    let dataset_selector;
```
vanshkumar, line 29: You can simplify this logic a bit by starting by setting dataset_selector to "sex" and removing the else branch below

Also, a better name for this variable may be "breakdown"

frontend/src/data/variables/AcsPovertyProvider.ts
```diff
@@ -24,9 +24,21 @@ class AcsPovertyProvider extends VariableProvider {
   }
   // ALERT! Make sure you update DataSourceMetadata if you update dataset IDs
   getDatasetId(breakdowns: Breakdowns): string {
+    let datasetPrefix = "acs_poverty_dataset-poverty_by_";
```
vanshkumar, line 27: It's odd to have different variable naming styles (camel case vs underscores) for datasetPrefix and dataset_select.. can you stick with one (preferably the one that is used everywhere else in this file)?

vanshkumar (approved):

PR #724, opened 2021-04-30
Fix conditional that broke combined race categories

jenniebrown:
#116

- Fixes combined race categories by fixing conditional statement
- Refactors CTP and CTP metadata ingestion to be easier to test
- Formatting

### DIFF
--- a/python/datasources/covid_tracking_project.py
+++ b/python/datasources/covid_tracking_project.py
@@ -10,83 +10,101 @@ from ingestion.standardized_columns import Race
 
 # Covid Tracking Project race data by state from covidtracking.com/race
 class CovidTrackingProject(DataSource):
-
     @staticmethod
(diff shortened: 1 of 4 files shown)


python/datasources/covid_tracking_project.py
```diff
@@ -134,7 +154,6 @@ def _replace_value(row: pd.Series, indicator_column: str,
             the race value.
         old_name: The race category to change
         new_name: The race category to rename to"""
-        if (row[indicator_column] is True and
-                row[col_std.RACE_COL] == old_name.race):
+        if row[indicator_column] == 1 and row[col_std.RACE_COL] == old_name.race:
```
jenniebrown, line 157: This was the line that was causing the issue

vanshkumar (approved):

Issue #655, opened 2021-04-22
Update Alert on the Small Multiples Overlay


adams314:
Remove the text below "Compare with Caution" and replace with:

xx.x% of cases reported unknown {demographic}. The charts below display data for cases where {demographic} was known. [View map of where unknown values are being reported]()

Clicking on the link should update the contents of the overlay to include only the unknowns map for the selected demographic.

kkatzen:
We're talking about this dialog right?

![Screen Shot 2021-04-23 at 7 30 39 PM](https://user-images.githubusercontent.com/6402808/115944407-6e5cdb80-a46a-11eb-9a9a-f858b45ac337.png)

by "only unknowns map for the selected demographic" do you mean the same chart as the Unknowns top-level card? so the multiple maps are replaced by just 1 map that's already on the main screen? Is it a link or a toggle? I'm not sure I understand...

adams314:
Yes, I mean a link that pulls up the unknowns map by itself in this dialog/overlay. A toggle sounds better though so you can switch back and forth without closing the window. 

If we could switch back to the main dash and focus on the unknowns graph that might work in a single column display but in our double column display I'm worried the user gets lost and doesn't realize we're directing them to the unknowns chart.

adams314:
Per discussion in the disclaimers doc, let's just remove the caution in the overlay for now and leave the unknown warnings to the main page

PR #729, opened 2021-05-02
Remove quantile scale alert

vanshkumar:


joshzarrabi (approved):

kkatzen:
I actually already did this in https://github.com/SatcherInstitute/health-equity-tracker/pull/706:
Could you let me do it in my PR so I don't have to deal with merge conflicts? Thanks!

vanshkumar:
> I actually already did this in #706:
> Could you let me do it in my PR so I don't have to deal with merge conflicts? Thanks!
> 
> Also issue #655 includes: "xx.x% of cases reported unknown {demographic}. The charts below display data for cases where {demographic} was known. View map of where unknown values are being reported" which I addressed in my PR so this PR doesn't close that issue anyways.
> 
> I think it would be best if we don't try to address issues others have already claimed without discussing it first- I think that could lead to a lot of confusion.

Ah, I didn't see your PR. Re: the actual text though, the latest decision here was to *not* add the message (per discussion in the disclaimers doc / the latest comment in the issue) and instead just remove this quantile scale alert. Happy to close this one if you want to handle the removal of the message in your PR

kkatzen:
Ah gotcha na it's okay you just do this one i'll close mine

PR #754, opened 2021-05-06
Replace newsletter form on home page with button linking to "How do I join the effort?" section

### LINKED ISSUE #740: Change email signup on bottom of homepage to a navigational button
The email signup at the bottom superimposed over the image of three people should be a navigational button with the text:

"Join our movement"

which links to: https://dev.healthequitytracker.org/whatishealthequity - How do I join the effort? section

efregoso:
Closes #740


frontend/src/pages/WhatIsHealthEquity/EquityTab.tsx
```diff
@@ -434,7 +434,7 @@ function EquityTab() {
               <FaqSection />
             </Grid>
           </Grid>
-          <Grid item className={styles.JoinTheEffortHeaderRow}>
+          <Grid item className={styles.JoinTheEffortHeaderRow} id="join">
```
vanshkumar, line 437: Is it possible to use the WIHE_JOIN_THE_EFFORT_SECTION_ID variable instead of the string "join" here?

vanshkumar (commented):
The green on green for the button makes it difficult to see, can we use a different color?

image

vanshkumar (approved):

Issue #738, opened 2021-05-04
Add new "sample size too small" map for BRFSS in place of unknowns map


vanshkumar:
Some details in this email:

![image](https://user-images.githubusercontent.com/3394642/117040755-11a9be00-acbf-11eb-8a8c-637ab6c39a40.png)

vanshkumar:
Latest is not add a new map

PR #762, opened 2021-05-06
DNS

kkatzen:
#738


kkatzen:
Want to consider different colors for the hover but we can easily change those later.

vanshkumar:
I think this can be closed as we no longer are planning to add this new map

PR #766, opened 2021-05-06
Updating map hover behavior

### LINKED ISSUE #764: Update map highlight
states that have data: Dark Blue #255792

States that DO NOT have data: Dark Red #A93038
Add in the tool tip: Sample size too small (instead of no data)

For survey data (brfss) - "no data" to "sample size too small"

kkatzen:
Closes #764


frontend/src/cards/MapCard.tsx
```diff
@@ -145,6 +145,9 @@ function MapCardWithKey(props: MapCardProps) {
             <MultiMapDialog
               fips={props.fips}
               metricConfig={metricConfig}
+              surveyCollectedData={
+                props.variableConfig.surveyCollectedData || false
```
vanshkumar, line 149: Why do you need to || with false here?

vanshkumar (approved):

PR #768, opened 2021-05-06
Remove special case renaming race to race_and_ethnicity now that pipeline has run

aaronsn:


aaronsn:
Note: this should not be pushed to prod until the prod pipeline has run with health insurance changes

vanshkumar:
Seems to not build?

kkatzen (approved):

PR #770, opened 2021-05-06
Add alert for survey data

kkatzen:


kkatzen:
Screen Shot 2021-05-06 at 12 56 50 PM

vanshkumar:
Let's wait on this PR until we have more time to think about the proper design/copy & also so we can incorporate feedback from Drs Taylor/Baily and the task force

vanshkumar:
I think we can go ahead and close this now as per the latest from MSM

PR #778, opened 2021-05-06
Update Missing Data text

adams314:
Updating the missing data text to reflect changes to suppression rules

### DIFF
--- a/frontend/src/reports/ReportProvider.tsx
+++ b/frontend/src/reports/ReportProvider.tsx
@@ -139,29 +139,26 @@ function ReportProvider(props: { madLib: MadLib; setMadLib: Function }) {
         </p>
         <h3>Missing Cases</h3>
         <p>
-          For COVID related reports, this tracker uses disaggregated, individual
-          case level data reported by states to the CDC. Unfortunately, some
-          states have not provided this disaggregated data to the CDC, so we
-          cannot report accurate metrics of health equity for the states of{" "}
-          <b>Louisiana</b>, <b>New Hampshire</b>, <b>Texas</b>, and{" "}
-          <b>Wyoming</b>. A number of other states report disaggregated data,
-          but the number of individual cases reported to the CDC is far fewer
-          than the aggregate numbers they report. These states' data are
-          included, but their data should be interpreted with caution since the
-          cases reported may not be representative of the population at large:{" "}
-          <b>Connecticut</b>, <b>Florida</b>, <b>Kentucky</b>, <b>Maryland</b>,{" "}
-          <b>Michigan</b>, <b>Mississippi</b>, <b>Missouri</b>, <b>Nebraska</b>,{" "}
-          <b>New Mexico</b>, <b>North Dakota</b>, <b>Ohio</b>,{" "}
-          <b>Rhode Island</b>, <b>West Virginia</b>.
+	  For COVID related reports, this tracker uses disaggregated, individual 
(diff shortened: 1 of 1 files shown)


frontend/src/reports/ReportProvider.tsx
```diff
@@ -139,29 +139,28 @@ function ReportProvider(props: { madLib: MadLib; setMadLib: Function }) {
...
-          <b>Connecticut</b>, <b>Florida</b>, <b>Kentucky</b>, <b>Maryland</b>,{" "}
-          <b>Michigan</b>, <b>Mississippi</b>, <b>Missouri</b>, <b>Nebraska</b>,{" "}
-          <b>New Mexico</b>, <b>North Dakota</b>, <b>Ohio</b>,{" "}
-          <b>Rhode Island</b>, <b>West Virginia</b>.
+	  For COVID related reports, this tracker uses disaggregated, individual{" "} 
+	  <a href="https://www.cdc.gov/coronavirus/2019-ncov/cases-updates/about-us-cases-deaths.html">
+	  case level data reported by states, territories, and other jurisdictions 
+	  to the CDC</a>. We cannot report accurate metrics for states that do not 
+	  provide disaggregated data to the CDC, so these states appear as  	 
+	  grey on the maps reporting Covid cases, hospitalizations and deaths:
```
vanshkumar, line 147: Few comments:
- Should the language be even more neutral/positive than this? E.g. "The following states are still working on providing disaggregated data to the CDC, and appear as grey..."
- Should "Covid" be "COVID"? That's how it's styled above
- Looks like we are no longer calling out states which have incomplete but not unusable data; I'm ok with that but curious if Ebony & others are?

frontend/src/reports/ReportProvider.tsx
```diff
@@ -139,29 +139,28 @@ function ReportProvider(props: { madLib: MadLib; setMadLib: Function }) {
...
-          Furthermore, many of the individual COVID case records are incomplete
-          with an unknown hospitalization and/or death status, meaning that even
-          in states that are reporting Covid case data, we have an incomplete
-          picture of its overall impact. Due to the nature of surveillance data,
-          we expect this picture to become more complete over time and will use
-          the Health Equity Tracker to record the progress.
+	  Furthermore, many COVID case records are incomplete, with an unknown 
+	  hospitalization and/or death status. This means that some states that 
+	  report disaggregated Covid case data still do not provide a complete 
+	  picture of its overall impact. Due to the nature of surveillance data,
```
vanshkumar, line 156: Couple comments:
- "Covid" -> "COVID" to be consistent everywhere
- I think you have an extra "on" in the sentence "following states appear as grey on when viewing"

frontend/src/reports/ReportProvider.tsx
adams314: re: tone: It might still be a little strong but let's get some opinions from the task force

I switched everything to "Covid" which seems to be the convention news orgs use. We probably need to go through the chart titles and resolve consistency.

I put the line back in about the states that are partially complete, removing those that are now included in the suppression list.

frontend/src/reports/ReportProvider.tsx
adams314: Thanks. removed the "on"

Making everything "Covid" in the paragraph text and will go back and make other mentions consistent tomorrow.

vanshkumar (approved):

PR #781, opened 2021-05-06
Change links into buttons

### LINKED ISSUE #709: Nav bar links should be buttons
Screen Shot 2021-04-29 at 1 00 23 PM

kkatzen:
Closes #709


aaronsn:
Nice! they look like real buttons now :) 

what do you think about also converting this to a button?

https://github.com/SatcherInstitute/health-equity-tracker/blob/b84ac6d5ace72d7f6b6538af11e60759c6ad276f/frontend/src/App.tsx#L94-L101

aaronsn (approved):
Approving, but looks like you have merge conflicts that need to be resolved

kkatzen:
> Nice! they look like real buttons now :)
> 
> what do you think about also converting this to a button?
> 
> https://github.com/SatcherInstitute/health-equity-tracker/blob/b84ac6d5ace72d7f6b6538af11e60759c6ad276f/frontend/src/App.tsx#L94-L101

I updated the logo but I thought the text looked kinda odd as a button so I left it alone

aaronsn (changes requested):
I just realized playing around with the deploy preview - using buttons causes the whole page to reload, whereas the links were instant changes. This seems like a pretty bad side effect - site navigation is much slower.

aaronsn:
> I just realized playing around with the deploy preview - using buttons causes the whole page to reload, whereas the links were instant changes. This seems like a pretty bad side effect - site navigation is much slower.

The LinkButton suggested here https://stackoverflow.com/a/49439893 seems like a pretty good solution. (a very similar version here: https://usehooks.com/useRouter/)

frontend/src/Footer.tsx
```diff
@@ -96,9 +97,7 @@ function Logos() {
       <Grid item className={styles.LogosLeft}>
         <Button
           className={styles.ImageButton}
-          onClick={() =>
-            (window.location.href = "https://healthequitytracker.org/")
-          }
+          href="https://healthequitytracker.org/"
```
aaronsn: What's going on here? This hard-coding seems wrong for a few reasons:
1) it won't work for the dev site or deploy previews
2) I believe linking to full url causes a full reload.

Can this just be a ReactRouterLinkButton that links to "/"?

frontend/src/App.tsx
```diff
@@ -91,9 +93,9 @@ function MobileAppToolbar() {
 function AppToolbar() {
   return (
     <Toolbar className={styles.AppToolbar}>
-      <LinkWithStickyParams to="/">
+      <Button component={Link} to="/">
```
aaronsn, line 96: Why can't you use the ReactRouterLinkButton here too?

aaronsn (approved):

frontend/src/Footer.tsx
kkatzen: already updated this

frontend/src/App.tsx
kkatzen: updated to accept react nodes too

PR #785, opened 2021-05-06
Age Sort by Number with All and Unbounded

Sethvg:


frontend/src/charts/TableChart.tsx
```diff
@@ -52,6 +52,28 @@ export function TableChart(props: TableChartProps) {
     },
   ].concat(columns);
 
+  if (breakdownVar === "age") {
```
vanshkumar, line 55: Can you make this logic simpler if you assume that anything that is unbounded (i.e. "x+") will be at the end of the list?

vanshkumar (commented):
Looks like the 6-11 age bucket for poverty is still not sorted properly (from your deploy-preview)?

image

frontend/src/charts/utils.ts
```diff
@@ -30,3 +30,23 @@ export function addLineBreakDelimitersToField(rawData: Row[], field: string) {
     return { ...data, [field]: lines.join(DELIMITER) };
   });
 }
+
+export const sortAgeParsedNumerically = (l: any, r: any) => {
```
kkatzen, line 34: We always want this sorted this way, right? If so, I don't think we should do this in charts.  I think it should be done in VariableProviders like we are doing for race here:
https://github.com/SatcherInstitute/health-equity-tracker/blob/6aa1713f25a580639e7904bde8ce6f6fbc886b53/frontend/src/data/utils/datasetutils.ts#L209

We should be calling maybeApplyRowReorder() in every VariableProvider already.

kkatzen (changes requested):

frontend/src/charts/utils.ts
kkatzen: To clarify, "for race" isn't quite right- this is reorders rows for any breakdowns which have "all" and "unknown" values (which really only effected race data)

vanshkumar: +1 to do this across the board via the function called in variable provider

Sethvg: I do think this belongs in the charting, and here's my reasons:

In OO programming we have the idea of [Seperation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns#:~:text=In%20computer%20science%2C%20separation%20of,code%20of%20a%20computer%20program.)

At a simplified level we have two concerns with our program.  Gathering the correct data, and displaying the data.  The variable provider should be responsible for choosing the correct dateset, requesting, and making sure the correct collumns and geos are present.  It should return order agnostic data, because quite frankly, it shouldn't know about the tables at all or how they work internally.  This means if we swapped out charting libraries it would still work without requiring a refactor.

Seperating the logic has a few advantages:

1) Keeping the provider from being a monolith of code where people just throw any and all logic into them.  Even now that's sorta what were seeing happen right now.

2) Not all charting may want the data in a certain order.  For instance, in the future, we may want to support race + age.  One table may want race sorted first and one table may want age sorted first.  This isnt future proofing, its just building a more Object oriented design.

3) Keeping tests simple.  If we test the data providers, we dont need to care about data order.  This allows easier testing of the code and is a direct result of not having monolithic classes.  

I have spoken my peace: ill defer to yours and vanshes final decisions if you still want it moved out just lmk :)

aaronsn: +1 to everything Krista said

To add a bit of context on the refactor Krista referenced, I think we can more or less merge all the VariableProviders into one class that can be customized per data source either via configuration or a lightweight interface (see more about that [here](https://docs.google.com/document/d/1TqsNgKIU-bLh0DhMqEIDs44xbgAXdaf2BMrxVOL92qo/edit?resourcekey=0-wiHCXAb64Ci1jHXaI7Aplg#heading=h.t5vfif6wk09l)). In my opinion, most of the complexity and "monolithic" aspects of the variable providers stem from the fact that we're writing very similar and somewhat complex code many times (one per provider), not from the variable providers taking on too many responsibilities. I agree with what Krista said - data sorting is one aspect of data processing and belongs in the providers, and if we want to support different sort orders we can add that to the MetricQuery interface.

We've taken steps to reduce the complexity of the providers by moving helper methods into the parent class, but we didn't want to try to merge all them too soon and over-engineer based on early datasets. Now that the site is coming together, we have a much better idea of what our requirements are (we've fleshed out things like sorting, handling unknowns, handling population pct quirks, etc) and it's a much better time to try to improve this code. Doing this refactor will require some slightly stricter requirements for datasets on the backend so I don't think we'll be able to get rid of it in one go, but I was hoping to take a stab at making a generalized provider and migrating data sources to it iteratively.

Sethvg: Okay, well 

1) I never meant to imply you and Aaron didn't build a good system.  Im sorry if I came across that way or you read it that way.  

2) You make some good points and I think your definition of where we disagree is spot on, but ill defer to the group so I moved it back.

frontend/src/data/utils/datasetutils.ts
```diff
@@ -221,5 +221,30 @@ export function maybeApplyRowReorder(rows: Row[], breakdowns: Breakdowns) {
...
   }
+
+  if (breakdowns.hasOnlyAge()) {
+    finalRows.sort(sortAgeParsedNumerically);
+  }
+
   return finalRows;
 }
+
+export const sortAgeParsedNumerically = (l: any, r: any) => {
```
vanshkumar, line 232: Can you add comments to this function explaining what inputs it accepts for the ages?

frontend/src/data/utils/datasetutils.ts
```diff
@@ -221,5 +221,30 @@ export function maybeApplyRowReorder(rows: Row[], breakdowns: Breakdowns) {
...
+
+  let leftUnbounded = lAge.indexOf("+") !== -1;
+  let rightUnbounded = rAge.indexOf("+") !== -1;
+
+  if (leftUnbounded && rightUnbounded) return 0;
+  else if (leftUnbounded) return 1;
+  else if (rightUnbounded) return -1;
+
+  let lMin = lAge.split("-")[0];
+  let rMin = rAge.split("-")[1];
```
vanshkumar, line 248: Did you mean to do rAge.split("-")[0] here instead of [1]? I would imagine you want to compare ages A and X when sorting A-B and X-Y.

vanshkumar: Thinking a bit more about this, I suppose what I am saying vs what you are doing would differ in the case of e.g. 6-11 and 5-7.. mine would order them as 5-7, 6-11, whereas yours would order them as 6-11, 5-7. In the case of non-overlapping buckets this doesn't matter, so I'm fine with leaving as-is, though changing the variable name to rMax instead of rMin would be helpful

vanshkumar (changes requested):

frontend/src/data/utils/datasetutils.ts
Sethvg: fixed

PR #791, opened 2021-05-07
UrlParams

Sethvg:


frontend/src/reports/ReportProvider.tsx
```diff
@@ -76,15 +76,26 @@ function ReportProvider(props: { madLib: MadLib; setMadLib: Function }) {
           props.setMadLib(
             getMadLibWithUpdatedValue(props.madLib, 5, fips.code)
           );
+        let fips1: Fips;
+        let fips2: Fips;
+        try {
+          fips1 = new Fips(fipsCode);
+          fips2 = new Fips(fipsCode);
+        } catch (err) {
```
joshzarrabi, line 84: Do you think you could specifically check to make sure we are getting `00` as the FIPS code instead of using a catch all error case?

Sethvg: I removed these as these were a workaround for the push state being fired prematurely and when backing would make invalid selections

joshzarrabi (approved):
Amazing, thank you!

PR #797, opened 2021-05-10
Code Cleanup: Calculation grouping out of util and base class with testing calculation methods directly

Sethvg:


joshzarrabi (approved):
Looks great!

frontend/src/data/utils/DatasetCalculator.ts
```diff
@@ -0,0 +1,128 @@
...
+      const total = totalRow.first()[rawCountCol];
+      return group
+        .generateSeries({
+          [pctShareCol]: (row) => this.percent(row[rawCountCol], total),
+        })
+        .resetIndex();
+    });
+  }
+
+  calculatePctShareOfKnown(
```
vanshkumar, line 65: It would be good to add a test for this function as you are adding tests for all the others

vanshkumar (dismissed):
thanks for adding testing!

frontend/src/data/utils/DatasetCalculator.ts
Sethvg: okie

vanshkumar (approved):

PR #803, opened 2021-05-10
only show races we have data for in brfss table card

### LINKED ISSUE #789: Stop showing racial groups for which we don't have diabetes/COPD data for
Similar to an issue that we had before:

image
image

Not sure why this is happening as we do have a brfss_population_pct metric which should exclude these rows..

joshzarrabi:
closes #789

### DIFF
--- a/frontend/src/cards/TableCard.tsx
+++ b/frontend/src/cards/TableCard.tsx
@@ -65,6 +65,9 @@ export function TableCard(props: TableCardProps) {
     .map((config) => config.metricId)
     .some((metricId) => metricId.includes("covid"));
 
+  //TODO figure out a better way to figure this out
+  const populationComparisonMetric = "brfss_population_pct";
+
   return (
     <CardWrapper
       queries={[query]}
@@ -77,7 +80,7 @@ export function TableCard(props: TableCardProps) {
       {([queryResponse]) => {
         const dataWithoutUnknowns = queryResponse
           .getValidRowsForFields(
-            metricIds.filter((m) => m !== "brfss_population_pct")
+            metricIds.filter((m) => m !== populationComparisonMetric)
           )
           .filter(
             (row: Row) =>
--- a/frontend/src/cards/TableCard.tsx
+++ b/frontend/src/cards/TableCard.tsx
@@ -43,6 +43,7 @@ export function TableCard(props: TableCardProps) {
       : undefined
   );
   let metricConfigs: Record<string, MetricConfig> = {};
+  var populationComparisonMetric: string;
   metrics.forEach((metricConfig) => {
     // We prefer to show the known breakdown metric over the vanilla metric, if
     // it is available.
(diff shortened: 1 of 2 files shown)


frontend/src/data/query/MetricQuery.ts
```diff
@@ -86,6 +86,16 @@ export class MetricQueryResponse {
     );
   }
 
+  getValidRowsForFields(fieldNames: string[]) {
+    let data = this.data;
+    fieldNames.forEach((name) => {
+      data = data.filter(
+        (row: Row) => row[name] !== undefined && row[name] !== null
```
vanshkumar, line 93: Can you just call getValidRowsForField instead of duplicating the logic?

frontend/src/cards/TableCard.tsx
```diff
@@ -75,11 +75,15 @@ export function TableCard(props: TableCardProps) {
...
     >
       {([queryResponse]) => {
-        const dataWithoutUnknowns = queryResponse.data.filter(
-          (row: Row) =>
-            row[props.breakdownVar] !== UNKNOWN &&
-            row[props.breakdownVar] !== UNKNOWN_RACE
-        );
+        const dataWithoutUnknowns = queryResponse
+          .getValidRowsForFields(
+            metricIds.filter((m) => m !== "brfss_population_pct")
```
vanshkumar, line 80: This doesn't seem like a particularly robust solution to this issue, but I can't think of a relatively simple one right now.. can we open an issue to think about how we can solve this better when we refactor etc?

vanshkumar (approved):

frontend/src/cards/TableCard.tsx
joshzarrabi: Its the same thing we do for the other cards 🤷‍♂️

vanshkumar: I don't think we filter for specific metric ids in other cards?

joshzarrabi: Right, I wasn't sure why that wasn't an issue for the other cards.

vanshkumar: As in why races for which we don't have data show up for other cards? For the disparities per 100k card (I think SimpleBarChartCard.tsx), it doesn't require the population_pct metric so doesn't have this issue. For the disparity bar chart card, it does use the population_pct metric, but it explicitly checks that the primary metric (i.e. % of cases for diabetes) is not null:

image

One way to make this more robust actually would be to filter out the populationComparisonMetric rather than just the specific "brfss_population_pct" metric, ie change to

getValidRowsForFields(
  metricIds.filter((m) => m !== populationComparisonMetric)

Where populationComparisonMetric is a variable you pull out and save above, in the same function. This would also make this fix work for all data sources rather than just brfss which is generally useful

joshzarrabi: Pushed!

PR #824, opened 2021-05-12
Show a % sign after numbers if metric is pct_share (applies to population card)

vanshkumar:


frontend/src/charts/SimpleHorizontalBarChart.tsx
```diff
@@ -204,6 +205,7 @@ export function SimpleHorizontalBarChart(props: SimpleHorizontalBarChartProps) {
           BREAKDOWN_VAR_DISPLAY_NAMES[props.breakdownVar],
           props.metric.metricId,
           props.metric.shortVegaLabel,
+          props.metric.type === "pct_share" ? "%" : "",
```
kkatzen, line 208: Would this helper work here? https://github.com/SatcherInstitute/health-equity-tracker/blob/2c02770c7f86613830ef7569e607b82413f9c734/frontend/src/data/config/MetricConfig.ts#L93

kkatzen (approved):
Not sure if helper is useful or not but I'll go ahead and approve- I think either way is fine your call

frontend/src/charts/SimpleHorizontalBarChart.tsx
vanshkumar: It would be ideal to use this helper yeah..I gave it a brief try but I wasn't able to figure out how to make it callable within the signal in SimpleHorizontalBarChart; if you know of an easy way to do this, lmk, otherwise will just submit as-is

joshzarrabi (approved):

Issue #826, opened 2021-05-12
Add dropdown to show five most and least affected locales


joshzarrabi:
### Problem

As a user of the site, it is difficult to find the most and least affected locales for a given demographic. For example, I would like to know which county has the most COVID cases for hispanic people in California. 

### Proposed solution

* A banner underneath the map will be clickable to show the locales with the 5 highest cases per 100k and 5 lowest cases per 100k for the shown demographic. 

* Only those locales will be highlighted on the map, with the 5 highest cases being dark green and 5 lowest cases being light green. (This is optional if its too hard)

See Image 3 for reference

![image](https://user-images.githubusercontent.com/3689026/118017299-70f17900-b324-11eb-9876-72dfc3a39e0d.png)

achipham:
Can we make the whole gray bar clickable for expand / collapse? Instead of just the down carrot.

kkatzen:
We need to revisit the map updating - either removing it or working on reducing blinking and also fixing tooltips - "shouldn't say no data /misisng data" when we're filtering in this use case

vanshkumar:
The main work is done here, leaving it open in backlog to discuss how we want the highlighting & tooltips to work post-launch

kkatzen:
we could just add a boolean to not show any tooltips on non-highlighted geographies- would be pretty easy

vanshkumar:
Yeah I think that's a good solution - do you want to send a fix, or should I?

PR #827, opened 2021-05-12
Adding highest/lowest lists to MapCard

kkatzen:
Addresses #826

There is some tacky blinking when the vega map has to reload...  Not really sure how much can be done about it but we can look into it
Screen Shot 2021-05-12 at 1 09 18 PM


joshzarrabi:
This is great! Maybe we should get rid of the map reload until we figure out how to make it fast? @vanshkumar what do you think?

frontend/src/cards/MapCard.tsx
```diff
@@ -125,6 +130,17 @@ function MapCardWithKey(props: MapCardProps) {
             (row: Row) => row[activeBreakdownVar] === activeBreakdownFilter
           );
 
+        const highestFive = dataForActiveBreakdownFilter
+          .sort((rowA: Row, rowB: Row) =>
+            rowA[metricConfig.metricId] <= rowB[metricConfig.metricId] ? 1 : -1
+          )
+          .slice(0, 5);
```
vanshkumar, line 137: Can we make the 5 into a constant at the top, and then also change the variable names here to mostAffectedGeos or something like that?

vanshkumar (commented):
> This is great! Maybe we should get rid of the map reload until we figure out how to make it fast? @vanshkumar what do you think?

Yeah, that's where I lean as well, though let's get feedback from SHLI folks in the sync w Daniel tomorrow morning

Thanks Krista, this is awesome! Just a note, I imagine we will want to do this for the unknowns map card later as well & at that time it makes sense to refactor this into something reusable, but this looks great for now!

frontend/src/cards/MapCard.tsx
kkatzen: Why the variable name change? What about highestRates and lowestRates ? I think mostAffectedGeos is less clear if you think about other metrics like "uninsured individuals" i think highest rates of uninsured is more straightforward than most affected geos for uninsured ?

vanshkumar: Suggesting the variable name change as the 5 will become a constant which can then be updated independently

Good point, highestRates/lowestRates sgtm

kkatzen: ah gotcha word

vanshkumar: Oh, I see now that the text "Top 5 Highest Rates" has a hard-coded 5 as well, we should use the constant there as well if possible - if it's not possible, I'm fine with leaving everything as-is (ie leave the 5 hard coded everywhere)

vanshkumar (approved):

kkatzen:
I moved the list into its own component so it'll be faster to add it to unknowns if we want to.

I left the map as it is - also worth noting the tooltips say "no data" for all places when it is expanded and we would probably want to change that, right? I was thinking I'd just leave the map updating for now so you can show Daniel the PR tomorrow and then we can submit without the map changing and follow up with a neater version?

joshzarrabi (approved):

PR #844, opened 2021-05-14
Add a max percentage resolution to fix the 'Other' gender issue

vanshkumar:
Makes it so that the %age of covid cases for Other becomes "no data" rather than 0%


vanshkumar:
Verified that this fixes the "Other" issue. It does have effects on certain other groups that are super tiny as well, though, e.g.:

Before:
image

After:
image

joshzarrabi (approved):

vanshkumar:
Going to extend this max percent resolution to the per 100k metrics as well just for consistency across all metrics

Issue #147, opened 2020-12-16
Covid Tracking Project Aggregation and Export


jenniebrown:
This is a tracking issue for identifying aggregations and writing aggregation routines for CTP race data.

jenniebrown:
We definitely want to join with FIPS codes. We also (probably) want to join with ACS population data. There are some states and territories which don't report any race data (reports_race or reports_ethnicity == 0). In those cases, we should not compare to population.

The only aggregation we want to do is to the national level, which should be done for both CRDT and population (don't use total US population from ACS, use aggregate from state/territory pops).

jenniebrown:
We will export 4 CRDT tables:

covid_tracking_project_cases_std
covid_tracking_project_deaths_std
covid_tracking_project_hosp_std
covid_tracking_project_tests_std

vanshkumar:
Joining CTP data with ACS is the only thing left here, moving to backlog

PR #850, opened 2021-05-14
Ctp export

jenniebrown:
#147 

Adds Exporter DAG step for CTP and updates the frontend data downloads page to pull from the exported tables instead of static files.

Does not include aggregations or joins with FIPS, ACS.


vanshkumar (approved):

aaronsn (commented):
Do you mind deleting the CTP files from the `frontend/public/tmp` directory?

Also, it's probably better to just submit the server-side changes, then only update the frontend code once the dev and prod pipelines have run, otherwise the frontend can break in the time between deploying the new code and running the new pipeline (I'm working on writing up guidelines for this). Since we're not launched yet this is optional for now

jenniebrown:
> Do you mind deleting the CTP files from the `frontend/public/tmp` directory?

Sure, I will do that in a followup PR.

> Also, it's probably better to just submit the server-side changes, then only update the frontend code once the dev and prod pipelines have run, otherwise the frontend can break in the time between deploying the new code and running the new pipeline (I'm working on writing up guidelines for this). Since we're not launched yet this is optional for now

That's a good point, it's important to run the pipelines in a timely fashion after submitting this. In this case I'm going to submit them together. They all need to be checked in before the code freeze next week, and I'd argue that the error that shows up when the dataset can't be retrieved is actually better than showing fake data.

PR #851, opened 2021-05-14
Load geographies as a dataset, and re-enable top/bottom 5 dropdown

aaronsn:
This allow the Choropleth map to update the data shown without re-loading the geography data from the server, which fixes the map flashing issue. It also allows caching the geographies so they're just loaded once


kkatzen (approved):
neat!! :D

joshzarrabi:
Really great, makes the site feel much faster and more responsive

frontend/src/data/react/WithLoadingOrErrorUI.tsx
```diff
@@ -120,7 +143,21 @@ export function WithMetadataAndMetricsWithKey(
...
+            if (!props.loadGeographies) {
+              return props.children(metadata, queryResponses);
+            }
+            return (
+              <WithDatasets
+                datasetIds={[GEOGRAPHIES_DATASET_ID]}
+                loadingComponent={props.loadingComponent}
+              >
+                {(datasets) =>
+                  props.children(metadata, queryResponses, datasets[0].rows)
```
vanshkumar, line 156: Using datasets[0] feels a bit fragile - any other way to pull out the geographies dataset here?

vanshkumar (approved):
dope

frontend/src/data/react/WithLoadingOrErrorUI.tsx
aaronsn: I think this is an existing problem with these utility components - they all take an array of things to load and return an array of values expected in the same order (modeled after Promise.all). I changed to array destructuring to make it more explicit, and added a TODO to consider changing to a map since I'd want to update all of these at the same time for consistency.

Issue #835, opened 2021-05-13
Add information about not having a map of territories


joshzarrabi:
### Use Case

As a user of the site
I want to know why the HET doesn't show Puerto Rico and other US territories on the map

### Potential Solution

Add information to the methodologies page about how vega (our mapping library) doesnt support US territories

related:
https://github.com/SatcherInstitute/health-equity-tracker/issues/760
https://github.com/SatcherInstitute/health-equity-tracker/issues/816

kkatzen:
Not sure if https://github.com/SatcherInstitute/health-equity-tracker/pull/853 depcreates this or we just want to update about the national level map?

vanshkumar:
I think it's probably worthwhile to still call out that the national level map doesn't show territories

Issue #816, opened 2021-05-11
Suppress map card in the absence of US territory representation


efregoso:
Currently, Vega does not support displaying information for US territories.

Until progress is made on #815, we want to hide the map card.

vanshkumar:
Hey Elizabeth, is this a dupe of #760 ?

kkatzen:
We sure make sure that when we hide the map card, the paired charts still line up in compare views

kkatzen:
Does https://github.com/SatcherInstitute/health-equity-tracker/pull/853 deprecate this issue?

vanshkumar:
Yes, I believe so

vanshkumar:
Closing as #853 fixed this

Issue #815, opened 2021-05-11
Replace Vega library with one that supports US territories


efregoso:
Currently, Vega does not support displaying information from US territories. 

We want to either amend Vega to support this (nontrivial; not recommended) or find another library that does support displaying US territories (preferred).

aaronsn:
I don't think this is inherent to Vega, I think it's just the geography dataset we're using

efregoso:
I think our most recent datasets do include data from US territories.  IIRC
from one of our syncs, they are being displayed in tables and other cards
where they show up but not in the mapcard.  @Vansh?

aaronsn:
By "geography dataset" I meant the dataset that defines the outlines of the states/counties. The one we're using is here: https://github.com/SatcherInstitute/health-equity-tracker/blob/master/frontend/public/counties-10m.json.

The Vega library doesn't have the geographies built-in, we provide our own geographies. Not sure where @kkatzen got the one we have from but I'm sure we can find another one that has territories.

Krista can confirm or correct if I've misstated anything here ^

kkatzen:
I think this is a dupe of #210
See more information about the issue here: https://gis.stackexchange.com/questions/366189/how-to-show-puerto-rico-in-albersusa-projection-in-vega-lite-or-vega

kkatzen:
This could be a solution: https://www.npmjs.com/package/geo-albers-usa-territories

joshzarrabi:
Another option we discussed was switching the national map to a hex based map. Another benefit to this choice would be to show each state as the same size, which is a bit more equitable.

kkatzen:
I will try to get a demo where we have additional submaps to show the territories on the national level and we'll see how we like it

kkatzen:
Territory circles were added to national level here: https://github.com/SatcherInstitute/health-equity-tracker/pull/988
With this change, I think we should close this bug and just stick with vega personally

benhammondmusic:
@joshzarrabi can we close this is is replacing the territory circled with their actual shapes on the national map still in consideration? I agree with @kkatzen we can close this

joshzarrabi:
Lets close it! Thanks :)

PR #853, opened 2021-05-14
Change projection to support territories

kkatzen:
See how Albers projection works for territory levels but not national level:
Screen Shot 2021-05-14 at 1 40 09 PM
Screen Shot 2021-05-14 at 1 40 03 PM

So we use AlbersUSA for every case except territories

Relevant issues:
#816
#815
#835


frontend/src/charts/ChoroplethMap.tsx
```diff
@@ -176,7 +176,9 @@ export function ChoroplethMap(props: ChoroplethMapProps) {
       projections: [
         {
           name: US_PROJECTION,
-          type: "albersUsa",
+          type: TERRITORY_CODES.includes(props.fips.code)
```
vanshkumar, line 179: Can you use the isTerritory function here? ie something like Fips(props.fips.code).isTerritory()

vanshkumar (approved):

frontend/src/charts/ChoroplethMap.tsx
kkatzen: o yaaa! nice

PR #854, opened 2021-05-14
Add "territory" as display name in Fips

kkatzen:
Screen Shot 2021-05-14 at 2 04 08 PM
Screen Shot 2021-05-14 at 2 03 50 PM


frontend/src/data/utils/Fips.ts
```diff
@@ -2,6 +2,14 @@ export const USA_DISPLAY_NAME = "United States";
 // Fake FIPS code used to represent totals in USA for convenience
 export const USA_FIPS = "00";
 
+export const TERRITORY_CODES = [
+  /*American Samoa-*/ "60",
+  /*Guam-*/ "66",
+  /*Northern Mariana Islands-*/ "60",
```
vanshkumar, line 8: Is this right? You seem to have "60" twice

joshzarrabi: I think Northern Marian Islands should be 69

kkatzen: oo good catch thanks!

vanshkumar (approved):

PR #863, opened 2021-05-14
Hide tooltips for greyed data when top/bottom list is open on MapCard

kkatzen:
Addressing #826


frontend/src/charts/ChoroplethMap.tsx
```diff
@@ -208,11 +209,15 @@ export function ChoroplethMap(props: ChoroplethMapProps) {
           type: "shape",
           from: { data: MISSING_DATASET },
           encode: {
-            enter: {
-              tooltip: {
-                signal: missingDataTooltipValue,
-              },
-            },
+            enter:
+              props.hideMissingDataTooltip === undefined ||
```
vanshkumar, line 213: Would it be simpler to check !== true here instead?

vanshkumar (approved):

frontend/src/charts/ChoroplethMap.tsx
kkatzen: genius! xD

PR #869, opened 2021-05-15
Legend label improvements

### LINKED ISSUE #868: Deaths legend labels should not be float
image

The labels should be integers here

vanshkumar:
Closes #868 

- Makes the legend labels on the unknowns map into %ages as they should be
- Adds a strategy to reduce label overlap & define overlapping as < 10 px of space between labels
- Change format of the labels for the map card & the small multiples to integer, as they are all per 100k metrics

### DIFF
--- a/frontend/src/charts/ChoroplethMap.tsx
+++ b/frontend/src/charts/ChoroplethMap.tsx
@@ -98,10 +98,7 @@ export function ChoroplethMap(props: ChoroplethMapProps) {
       ? "Sample size too small"
       : "No data";
     const geographyName = props.showCounties ? "County" : "State";
-    const tooltipDatum =
-      props.numberFormat === "percentage"
-        ? `format(datum.${props.metric.metricId}, '0.1%')`
-        : `format(datum.${props.metric.metricId}, ',')`;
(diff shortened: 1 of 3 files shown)


frontend/src/charts/ChoroplethMap.tsx
```diff
@@ -116,10 +113,21 @@ export function ChoroplethMap(props: ChoroplethMapProps) {
       titleLimit: 0,
       font: "monospace",
       labelFont: "monospace",
+      labelOverlap: "greedy",
+      labelSeparation: 10,
       offset: 10,
+      format: "d",
     };
     if (props.numberFormat === "percentage") {
```
kkatzen, line 121: could we switch numberFormat to MetricConfig.type? They should always be the same and it reduces chance of people forgetting to set the numberFormat properly (I think numberFormat may have been created before we added the MetricConfig)

kkatzen (approved):

frontend/src/charts/ChoroplethMap.tsx
vanshkumar: Sg, done

PR #870, opened 2021-05-16
Remove max percent resolution

vanshkumar:
Upon further thought, I don't think it's the right thing to do to have a max percent resolution. There are cases such as 0 deaths per100k for age 0-9 in Alaska, which I think we should indeed show as 0 per100k rather than saying "no data." Basically, I don't think it makes sense to suppress a data point if it's < 1 per100k and instead we should just show the 0 - "no data" has a specific meaning in the tracker (no underlying reported data) and being < 1 per100k doesn't fit that meaning.

Note that the upshot is that we will see 0% share of cases for gender "Other" in the US; IMO it's better to launch with this minor weirdness and think through the proper way to fix this post-launch.

### DIFF
--- a/frontend/src/data/utils/DatasetCalculator.test.ts
+++ b/frontend/src/data/utils/DatasetCalculator.test.ts
@@ -1,8 +1,5 @@
 import { DataFrame, IDataFrame } from "data-forge";
-import {
-  DatasetCalculator,
-  MAXIMUM_PERCENTAGE_RESOLUTION,
-} from "./DatasetCalculator";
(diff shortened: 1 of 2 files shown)


aaronsn (approved):
Is there an issue filed for finding a better solution for the 0/no data issue? If not mind filing one?

vanshkumar:
> Is there an issue filed for finding a better solution for the 0/no data issue? If not mind filing one?

Yes, will file an issue and throw it on the backlog

PR #871, opened 2021-05-16
Quick fix to remove DC county-level (fips 11001) data from the tracker

### LINKED ISSUE #872: Fix funky DC data
District of Columbia shows up in the CDC covid data as both a state (fips 11) and a county (fips 11001). Below are the case counts for it as a state & as a county:

image
image

Clearly, the case counts for 11001 are basically zero and way lower than the case counts for it as a state (fips 11). This is an oddity that we should follow up with the CDC on, as in theory all cases that are assigned to DC the state should also be assigned to DC the county, as they are one and the same. Until we can follow up with the CDC to get this fixed properly, a simple quick fix is to just remove the county-level data for fips 11001 from the tracker. This is what things look like in the tracker today for DC:

Screen Shot 2021-05-15 at 7 22 01 PM
Screen Shot 2021-05-15 at 7 22 15 PM

vanshkumar:
Closes #872 

See the issue for details

### DIFF
--- a/frontend/src/data/variables/CdcCovidProvider.ts
+++ b/frontend/src/data/variables/CdcCovidProvider.ts
@@ -119,18 +119,22 @@ class CdcCovidProvider extends VariableProvider {
     // Drop unused columns for simplicity.
     df = df.dropSeries(["death_n", "death_unknown", "hosp_n", "hosp_unknown"]);
 
-    // Hotfix for District of Columbia - remove all county-level DC data.
-    // df = df.where((row) => row.fips !== DC_COUNTY_FIPS);
-    df = df.select((row) => {
-      return {
-        fips: row.fips,
-        fips_name: row.fips_name,
-        [breakdownColumnName]: row.breakdownColumnName,
-        covid_cases: row.fips === DC_COUNTY_FIPS ? null : row.covid_cases,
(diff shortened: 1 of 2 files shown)


kkatzen (approved):

frontend/src/data/variables/CdcCovidProvider.ts
```diff
@@ -116,6 +116,27 @@ class CdcCovidProvider extends VariableProvider {
       })
       .resetIndex();
 
+    // Drop unused columns for simplicity.
+    df = df.dropSeries(["death_n", "death_unknown", "hosp_n", "hosp_unknown"]);
+
+    // Clear all county-level DC data. See issue for more details:
+    // https://github.com/SatcherInstitute/health-equity-tracker/issues/872.
+    // TODO - fix this the right way.
+    df = df.withSeries({
```
kkatzen, line 125: is it worth extracting into a util so it's easy to reuse if needed in the future?

kkatzen (approved):

frontend/src/data/variables/CdcCovidProvider.ts
vanshkumar: Hmm, my feeling is that's not needed for now, and we can extract into a util in the future as needed

joshzarrabi (approved):

PR #874, opened 2021-05-16
Add disclaimer to highest/lowest rates

### LINKED ISSUE #840: Add disclaimer to bottom side of top/bottom X dropdown list
### Problem

When a user click to see the top and bottom states for COVID outcomes, they will get a misleading picture of the states with the lowest rates of COVID. The states in the "lowest rates" list are all there because they do a poor job of reporting, not due to great COVID suppression. See the image below.

Screen Shot 2021-05-13 at 4 55 15 PM

### Proposed Solution

After the text `See the states with the highest and lowest rates of COVID-19 Cases` We should add the text `Consider the impact of Gaps in Data when interpreting highest and lowest rates.` that links to the disclaimers at the bottom of the page.

vanshkumar:
Closes #840

### DIFF
--- a/frontend/src/cards/ui/HighestLowestList.tsx
+++ b/frontend/src/cards/ui/HighestLowestList.tsx
@@ -29,7 +29,7 @@ export interface HighestLowestListProps {
   // List of rows with lowest rates
   lowestRatesList: Row[];
   // Callback for jumping to the missing data disclaimer.
-  jumpToDisclaimer: () => void;
+  jumpToDisclaimer?: () => void;
 }
 
 /*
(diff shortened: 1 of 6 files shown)


frontend/src/cards/ui/HighestLowestList.tsx
```diff
@@ -98,6 +101,18 @@ export function HighestLowestList(props: HighestLowestListProps) {
         </Grid>
       </div>
       <p>All rates are reported as: {props.metricConfig.fullCardTitleName}</p>
+      <p>
+        Consider the possible impact of
+        <Button
+          onClick={() =>
```
kkatzen, line 107: I wonder if it's better to just do a plain document.getElementById here rather than just pass the reference down two layers?

joshzarrabi (approved):

frontend/src/cards/ui/HighestLowestList.tsx
vanshkumar: much simpler, done

PR #875, opened 2021-05-16
Show races for which we don't have data in the table card for Covid

vanshkumar:


vanshkumar:
Still need to add tests, but it works.. comparison of deploy-preview vs current dev site for NM:

Screen Shot 2021-05-16 at 1 21 54 PM
Screen Shot 2021-05-16 at 1 18 03 PM

joshzarrabi (approved):
Approving but tests would be great!

vanshkumar:
> Approving but tests would be great!

Added tests

vanshkumar:
Btw, one consequence of this change is that "Other" shows up with "no data" for all states now.. IMO this is fine but just wanted to call it out

vanshkumar:
Upon further review, I think this slows the calculations for Covid down too much to be worth it.. going to just move the underlying issue to the backlog and we can fix it properly (i.e. add in the missing rows on the BE) later

PR #876, opened 2021-05-16
Revamp of methodology page

vanshkumar:
After this PR, would like to make the "what data are missing" callout in the text into a link

### DIFF
--- a/frontend/src/pages/DataCatalog/MethodologyTab.tsx
+++ b/frontend/src/pages/DataCatalog/MethodologyTab.tsx
@@ -192,9 +192,10 @@ function MethodologyTab() {
               </h2>
               <div className={styles.MethodologyAnswer}>
                 <p>
-                  None of the metrics shown on the tracker are age-adjusted. We
-                  are aware that showing non-adjusted data can mask disparities
-                  and are working on moving to age-adjusted data.
+                  None of the metrics/data shown on the tracker are
+                  age-adjusted. We are aware that showing non-adjusted data can
+                  mask disparities and are working on moving to age-adjusted
+                  data.
                 </p>
                 <p>
                   In the definitions below, we use COVID-19 Cases as the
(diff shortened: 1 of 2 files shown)


vanshkumar:
Decided to add in the metric definitions in this PR as well, working on that now

frontend/src/pages/DataCatalog/MethodologyTab.tsx
```diff
@@ -51,40 +51,111 @@ function MethodologyTab() {
...
+                <h4 className={styles.MethodologySubsubheaderText}>COVID-19</h4>
                 <p>
-                  Unfortunately, with these publically available data sets,
-                  there are crucial pieces missing, including but not limited
-                  to: comprehensive city-, census tract-, and county-level data;
-                  comprehensive race and ethnicity breakdowns; comprehensive
-                  gender and age breakdowns by county, etc.
+                  {/* TODO - this should really be a button... */}
+                  For a description of some of the gaps in COVID-19 data, please
+                  see the "What Data Are Missing" section on the "Explore the
```
adams314, line 63: Add link?

adams314 (approved):
LGTM but please check the adjustments I made

vanshkumar:
> LGTM but please check the adjustments I made

All of your adjustments make sense, except for the very last one re: racial categories, the change was

Before:
Generally, all race/ethnicities on the tracker include Hispanic/Latino unless otherwise specified.

After:
Generally, all race/ethnicities on the tracker include Hispanic/Latino unless otherwise specified. A notable exception is Covid data, where cases, hospitalizations and deaths that are recorded as  Hispanic/Latino ignore any other race that is reported.

This applies to both Covid and BRFSS data, and also I don't think we need to call it out here as this is one of the examples presented in the list below

adams314:
SGTM thanks!

frontend/src/pages/DataCatalog/MethodologyTab.tsx
vanshkumar: Planning to submit this first and then add it after (somewhat dependent on another PR)

PR #888, opened 2021-05-18
Text share fix

Sethvg:


frontend/src/reports/ui/ShareDialog.tsx
```diff
@@ -1,23 +1,22 @@
...
-import Button from "@material-ui/core/Button";
+import { getMadLibPhraseText, MadLib } from "../../utils/MadLibs";
 
 function ShareDialog(props: {
   madLib: MadLib;
   shareModalOpen: boolean;
   setShareModalOpen: (shareModalOpen: boolean) => void;
 }) {
   const [textCopied, setTextCopied] = useState(false);
+  const text = window.location.href;
```
aaronsn, line 19: I'm not entirely sure how react re-rendering works with window.location.href.

I think it might be better to pass this in as a prop so that the component re-renders whenever the url changes. Probably in practice this doesn't matter since the share dialog isn't expected to change, but I think just good practice?

Sethvg: Yeah this should never need to be rerendered but its just following previous behavior :).  TBH this dialogue shouldnt even exist but 🤷‍♂️

PR #897, opened 2021-05-19
Fix console warnings (mostly missing keys), add clearer error message…

aaronsn:
… for non-200 status codes, fix a no-op setState call, and log errors to console at correct severity

### DIFF
--- a/frontend/src/cards/MapCard.tsx
+++ b/frontend/src/cards/MapCard.tsx
@@ -72,7 +72,7 @@ function MapCardWithKey(props: MapCardProps) {
   const [listExpanded, setListExpanded] = useState(false);
 
   const [activeBreakdownFilter, setActiveBreakdownFilter] = useState<string>(
-    ""
+    ALL
   );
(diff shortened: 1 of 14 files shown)


frontend/src/cards/MapCard.tsx
```diff
@@ -128,13 +128,6 @@ function MapCardWithKey(props: MapCardProps) {
 
         breakdownValues.sort.apply(breakdownValues, sortArgs);
 
-        if (
-          activeBreakdownFilter === "" ||
-          activeBreakdownFilter === undefined
-        ) {
-          setActiveBreakdownFilter(ALL || breakdownValues[0]);
-        }
-
```
Sethvg, line 137: Is this intention?

Sethvg (approved):

frontend/src/cards/MapCard.tsx
aaronsn: Yes, I don't know why this was written this way but it was generating a console warning and it isn't doing anything because it just always sets it to ALL, so I moved ALL to the default value in useState()

PR #909, opened 2021-05-19
Format percentages to at least one decimal place, and add formatted d…

aaronsn:
…ata outside of vega so we can reuse the formatting logic


aaronsn:
Before:
image
image

After:
image
image

vanshkumar (approved):

PR #916, opened 2021-05-20
Change margin on profiles so that they are 5 per row in the About Us > Our Team Leadership section

### LINKED ISSUE #752: About Us > Our Team: Fix alignment of Leaders profiles
We have received feedback from Mahia that the profiles don't look evenly distributed between the two rows in the Leaders section.

Given that the number of profiles in this section is now static (and even), I think we can come up with a solution where the profiles are evenly distributed between two rows at different breakpoints.

efregoso:
Closes #752


efregoso:
Closing & re-pushing after merge bc Github is weird about merging image files.

PR #924, opened 2021-05-20
Disable three dots Vega menu everywhere

vanshkumar:


joshzarrabi (approved):

benhammondmusic:
FYI the "right click fix" PR I made will allow users to right click any map and choose save image. This is another potential reason to eliminate the three dots

jgonzalezmsm:
> FYI the "right click fix" PR I made will allow users to right click any map and choose save image. This is another potential reason to eliminate the three dots

I think removing the 3 dots, at least until we can develop a better use for them would make sense right now.

In the future I would like to be able to add "share" options or similar than can send their current selection

benhammondmusic:
Currently the three dots does not pass color contrast a11y. If we do decide to keep the three dot feature, we will need to adjust the color and make it more visible

benhammondmusic:
fixes #1043

benhammondmusic:
Make sure we update the "how to share" on FAQ if this PR is merged:

"How do I share the graphs?
Next to each graph, there is a circle-shaped button with three dots in it. Click on that button to see a menu where you can save each graph as a PNG or SVG."

jgonzalezmsm:
@benhammondmusic is it possible to add new functionality for the 3 dots? Share to social media platforms, email, and keep the save as png (as another option to right clicking)?

if it's too complex or not possible then I'd say let's add removing the icon to the next meeting agenda to move forward with removing

benhammondmusic:
I didn't think it's possible to alter the Vega 3-dot widget that much from what I've seen; certainly possible to build our own little widget that would go somewhere on the card though. Would look similar to the "share" button that copies the site link to your clipboard but would be per card/map/chart and would offer various options.

benhammondmusic:
NOTE: discussed with team, Christian and others said the "export as PNG" option is being used, so we should hold on on merging this until we have a better option. Filing separate issue to address existing problems with this Vega 3-dot "actions" menu

PR #928, opened 2021-05-21
Fix NaNs in highest/lowest

### LINKED ISSUE #927: deaths and hosps for Covid show NaN on top 5 dropdown
on national map

vanshkumar:
Closes #927

### DIFF
--- a/frontend/src/data/utils/datasetutils.ts
+++ b/frontend/src/data/utils/datasetutils.ts
@@ -146,6 +146,7 @@ export const getLowestN = (
   listSize: number
 ): Row[] => {
   return data
+    .filter((row: Row) => !isNaN(row[fieldName]) && row[fieldName] !== null)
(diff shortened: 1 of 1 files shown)


joshzarrabi (approved):

frontend/src/data/utils/datasetutils.ts
```diff
@@ -146,6 +146,7 @@ export const getLowestN = (
   listSize: number
 ): Row[] => {
   return data
+    .filter((row: Row) => !isNaN(row[fieldName]) && row[fieldName] !== null)
```
aaronsn, line 149: Is it possible the data has undefined in it? You can use `!=` instead of `!==` to capture both null and undefined

vanshkumar: It is possible, did not know that you could do that, done!

PR #930, opened 2021-05-24
update date for the cdc data on the methodologies page

joshzarrabi:


vanshkumar (commented):
frontend/src/data/config/FakeDatasetMetadata.ts also needs to be updated to say "May" for the CDC data

vanshkumar (approved):

PR #934, opened 2021-05-26
Replace all &nbsp characters with regular spaces

jenniebrown:
Non-breaking spaces in the What Data Are Missing? section caused text not to wrap in the state lists on mobile. Replaces the &nbsp characters with regular spaces to allow proper text wrapping.

Before:

Before

After:

After


benhammondmusic:
this appears to have been addressed elsewhere, as this problem is no longer occurring; closing this PR draft

PR #954, opened 2021-06-03
Landing page CSS tidying

kkatzen:
Using @media instead of <hidden> to modify styles at different breakpoints


joshzarrabi (approved):

joshzarrabi:
Thanks!

PR #988, opened 2021-06-11
Tiny Territory maps demo

kkatzen:


joshzarrabi (approved):
looks great, we will merge it in once this territory data is up to date

kkatzen:
> looks great, we will merge it in once this territory data is up to date

I added some always false conditions to prevent this feature from being shown. I filed a [github issue](https://github.com/SatcherInstitute/health-equity-tracker/issues/1011 ) and added TODO(1011) in each place where you'll need to remove the false to enable the feature.

That way you can go ahead and merge it in now and take your time enabling it without worrying about merge conflicts later :)

PR #989, opened 2021-06-14
return mime type to text

Sethvg:
With Text Mime Type.

First Request: 509ms
Second+ Request: 30ms.

Without Text MimeType.
First: 649.
Second: 584

This is local only, and there is a small chance it breaks the backend, if so the backend will need to use a JSON.parse on the front end to get it from text format back to JS object styles. Its very difficult to test as its proxied via the frontend_server, and wiring them up locally is kind of a pain.


vanshkumar:
What is the impact of this change? Could you provide some latency before/after numbers, if that's what this is intended to speed up?

vanshkumar (approved):
LGTM, though please wait for Josh to review as this may break the dev site as you noted

joshzarrabi:
Hey, thanks, yeah honestly testing this will be a pretty big pain, I was never able to successfully wire everything up. I think its possible and I know how to do it but it would take kind of a long time, I think we should merge and test it on the dev server.

joshzarrabi (approved):

joshzarrabi:
I'm not convinced we should do this, it is a json. https://stackoverflow.com/questions/4350753/advantages-of-using-application-json-over-text-plain

joshzarrabi:
Closing this for now

PR #990, opened 2021-06-15
Small refactor to CDC local script ahead of adding tests

vanshkumar:


joshzarrabi (changes requested):
Would you consider adding flags to take in the input? Its just a little easier to use IMO

vanshkumar:
> Would you consider adding flags to take in the input? Its just a little easier to use IMO

Will do in a follow-up PR

joshzarrabi (approved):

PR #992, opened 2021-06-16
Make unknowns map for county-level territories work

vanshkumar:
Unknown map card for Guam, which works now:

Screen Shot 2021-06-17 at 9 01 52 PM

All other cards are unchanged


joshzarrabi:
Looks great, good to show the data that we have :)

joshzarrabi (approved):

joshzarrabi (changes requested):
We decided to only show the unknowns map when we don't have population data, because showing the percentage of cases without the corresponding population data can be misleading and does not present the data in the proper context

vanshkumar:
> We decided to only show the unknowns map when we don't have population data, because showing the percentage of cases without the corresponding population data can be misleading and does not present the data in the proper context

Changed

joshzarrabi (approved):

Issue #963, opened 2021-06-03
Move renaming of TOTAL to ALL to the BE


vanshkumar:
Currently we do this in the FE via the renameTotalToAll method in VariableProvider.ts; we should do this on the BE instead

PR #994, opened 2021-06-17
backend changes to productionalize uhc data

joshzarrabi:
- pulls the America's Health Ranking data from the website
- fills the `uhc_data` table with the info


airflow/dags/uhc.py
```diff
@@ -0,0 +1,35 @@
...
+_UHC_DATASET_NAME = 'uhc_data'
+
+default_args = {
+    'start_date': days_ago(0),
+}
+
+data_ingestion_dag = DAG(
+    'uhc_ingestion_dag',
+    default_args=default_args,
+    schedule_interval='@yearly',
```
vanshkumar, line 18: Do you necessarily want a yearly schedule for this, as opposed to just having it be manually run?

config/data_sources/uhc_data.tf
```diff
@@ -0,0 +1,7 @@
+# Resources and routines for CDC restricted data ingestion.
```
vanshkumar, line 1: Please update names in comments to UHC from CDC restricted

python/ingestion/constants.py
```diff
@@ -52,6 +52,66 @@
     "Wyoming"
 ]
 
+STATE_NAMES_TO_FIPS = {
```
vanshkumar, line 55: Rather than hard-coding a mapping here, IMO it would be better to do the state name -> FIPS conversion in an SQL aggregation routine for now, similar to what we do for CDC restricted here (join with a public bq table which has state names & fips):

https://github.com/SatcherInstitute/health-equity-tracker/blob/master/config/data_sources/AGG_cdc_restricted.sql#L22

python/datasources/uhc.py
```diff
@@ -0,0 +1,122 @@
...
+UHC_RACE_GROUPS_TO_STANDARD = {
+    'American Indian/Alaska Native': Race.AIAN_NH.value,
+    'Asian': Race.ASIAN_NH.value,
+    'Black': Race.BLACK_NH.value,
+    'Hispanic': Race.HISP.value,
+    'Hawaiian/Pacific Islander': Race.NHPI_NH.value,
+    'Other Race': Race.OTHER_STANDARD_NH.value,
+    'White': Race.WHITE_NH.value,
+    'Multiracial': Race.MULTI_NH.value,
+    'All': Race.ALL.value,
```
vanshkumar, line 34: Rather than adding a new Race.ALL, why not just use Race.Total?

The convention so far has been to use "Total" everywhere in the BE and then we convert "Total" -> "All" on the FE (not a great system, but that's how it's been done so far). If the join with ACS population is planned to be done on the BE for this dataset at some point, you will have to use "Total" rather than "All" for it to work as things currently stand (for race, age, and sex)

python/datasources/uhc.py
```diff
@@ -0,0 +1,122 @@
...
+    def upload_to_gcs(self, _, **attrs):
+        raise NotImplementedError(
+            'upload_to_gcs should not be called for UHCData')
+
+    def write_to_bq(self, dataset, gcs_bucket, **attrs):
+        df = gcs_to_bq_util.load_csv_as_dataframe_from_web(
+                BASE_UHC_URL, dtype={std_col.STATE_FIPS_COL: str}
+            )
+
+        for b in [std_col.RACE_OR_HISPANIC_COL, std_col.AGE_COL, std_col.SEX_COL]:
```
vanshkumar, line 64: nit: please use an expressive name like "breakdown" for the variable in this loop

python/datasources/uhc.py
```diff
@@ -0,0 +1,122 @@
...
+
+            if std_col.RACE_INCLUDES_HISPANIC_COL in breakdown_df.columns:
+                column_types[std_col.RACE_INCLUDES_HISPANIC_COL] = 'BOOL'
+
+            breakdown_df.to_csv("test_%s.csv" % b)
+            gcs_to_bq_util.add_dataframe_to_bq(
+                breakdown_df, dataset, b, column_types=column_types)
+
+    def generate_breakdown(self, breakdown, df):
+        breakdown_map = {
```
vanshkumar, line 79: This seems like something that should be a top-level constant rather than defined in a method

python/datasources/uhc.py
```diff
@@ -0,0 +1,122 @@
...
+                            (df['State Name'] == state) &
+                            (df['Measure Name'].str.contains(determinent)) &
+                            (df['Measure Name'].str.contains(breakdown_value))]
+
+                        if len(row) > 0:
+                            pct = row['Value'].values[0]
+                            if pct:
+                                output_row[UHC_DETERMINANTS_OF_HEALTH[determinent]] = pct
+
+                output = output.append(output_row, ignore_index=True)
```
vanshkumar, line 117: In general, it's discouraged to create an empty df and keep appending rows to it, as it is quite inefficient (eg https://stackoverflow.com/questions/13784192/creating-an-empty-pandas-dataframe-then-filling-it/56746204#56746204). An easy alternative is to just build a list of lists, where each list is a row in the df, and then construct it at once at the end while also specifying the columns, e.g.:

df = pd.DataFrame(list_of_lists, columns=columns)

python/datasources/uhc.py
```diff
@@ -0,0 +1,122 @@
...
+                output_row['state_name'] = state
+
+                if breakdown == std_col.RACE_OR_HISPANIC_COL:
+                    output_row[std_col.RACE_CATEGORY_ID_COL] = UHC_RACE_GROUPS_TO_STANDARD[breakdown_value]
+                else:
+                    output_row[breakdown] = breakdown_value
+
+                output_row['state_fips'] = constants.STATE_NAMES_TO_FIPS[state]
+
+                for determinent in UHC_DETERMINANTS_OF_HEALTH:
```
vanshkumar, line 100: typo: should be "determinant" rather than "determinent"

airflow/dags/uhc.py
```diff
@@ -0,0 +1,35 @@
+# Ignore the Airflow module, it is installed in both dev and prod
+from airflow import DAG  # type: ignore
+from airflow.utils.dates import days_ago  # type: ignore
+
+import util
+
+_UHC_BASE_URL = "https://www.americashealthrankings.org/api/v1/downloads/210"
```
vanshkumar, line 7: This seems to be unused?

python/datasources/uhc.py
```diff
@@ -0,0 +1,122 @@
...
+                    if breakdown_value == 'All':
+                        output_row[UHC_DETERMINANTS_OF_HEALTH[determinent]] = df.loc[
+                            (df['State Name'] == state) &
+                            (df['Measure Name'] == determinent)]['Value'].values[0]
+
+                    else:
+                        row = df.loc[
+                            (df['State Name'] == state) &
+                            (df['Measure Name'].str.contains(determinent)) &
```
vanshkumar, line 109: I'm not a huge fan of doing str.contains to find the right entry in the df, but the raw data is in kind of a weird format. What do you think about building the correct phrase, e.g. "Diabetes - White", and then looking it up directly?

vanshkumar (commented):
Looks good, will take another pass once you get a chance to address comments

python/datasources/uhc.py
joshzarrabi: I was trying to sty ahead of the times https://github.com/SatcherInstitute/health-equity-tracker/issues/963

python/datasources/uhc.py
joshzarrabi: Makes sense, its definitely really slow

python/datasources/uhc.py
vanshkumar: Ah OK, in that case sgtm to leave what you have as-is :)

python/datasources/uhc.py
joshzarrabi: We would have to add another if condition to special case for age in that case. TBH I would rather avoid that and leave it like this.

vanshkumar: SG

joshzarrabi:
Thanks for the great feedback @vanshkumar !

python/datasources/uhc.py
vanshkumar: think this typo still needs to be fixed

python/datasources/uhc.py
```diff
@@ -0,0 +1,128 @@
...
+            columns.append(std_col.RACE_CATEGORY_ID_COL)
+        else:
+            columns.append(breakdown)
+
+        for state in states:
+            for breakdown_value in BREAKDOWN_MAP[breakdown]:
+                output_row = {}
+                output_row[std_col.STATE_NAME_COL] = state
+
+                if state == "United States":
```
vanshkumar, line 96: Is this still needed as you handle this case in the SQL?

vanshkumar (approved):

python/datasources/uhc.py
joshzarrabi: right

PR #997, opened 2021-06-18
Add missing demographics for all geos for CDC data

### LINKED ISSUE #841: Show races for which we don't have data in the table card

vanshkumar:
Closes #841

Note that the CDC data generated by this change is already in dev. Once #996 gets in, everything should work - we can push #996 to prod and put the new CDC data generated by this change into prod after.


joshzarrabi:
Looks great!

joshzarrabi (approved):

Issue #993, opened 2021-06-17
Puerto Rico Total Cases by Race and Ethnicity 


jgonzalezmsm:
Puerto Rico Data is confusing in Total COVID-19 Cases By Race And Ethnicity due to calculation and high amount of unknowns + single race/ethnicity recorded. Results in both Hispanic and Latino and All categories showing 100%.

Proposal to remove "All" as a category from table and move to alert/info box similar to unknown map. 

Add an info bubble to the map card saying what the per100k number is for the overall geography
Remove “All” from the table card
Add “with known demographics” to the share of label text

Krista to make change, JC to write infobox text.

vanshkumar:
Don't think the last change (add "with known demographics") is needed if we make the first two changes

kkatzen:
Here's what I came up with for the map card:
![Screen Shot 2021-07-02 at 1 58 08 PM](https://user-images.githubusercontent.com/6402808/124328067-c3d5f880-db3d-11eb-9271-ae00976f70c5.png)

PR #1004, opened 2021-07-01
Moving overall from table to map card

kkatzen:
Closes the eng parts of #993

Additionally, the map card was still supporting querying for multiple breakdowns at once (age, sex, race) based on an older version that we moved away from a long time ago. I don't think we'll bring that back so I went ahead and removed that logic.


joshzarrabi:
This all looks good to me, thanks @kkatzen !

joshzarrabi:
@kkatzen do you think this would make sense to push to a branch on the main repo so I can change it later if I need to ?

kkatzen:
> @kkatzen do you think this would make sense to push to a branch on the main repo so I can change it later if I need to ?

I think since this PR actually changes quite a bit of logic in the Map Card I'd be worried about annoying merge conflicts later if it sits in another branch for awhile- if you need to get rid of this in a hurry in the main repo I would just comment it out

kkatzen:
> > @kkatzen do you think this would make sense to push to a branch on the main repo so I can change it later if I need to ?
> 
> I think since this PR actually changes quite a bit of logic in the Map Card I'd be worried about annoying merge conflicts later if it sits in another branch for awhile- if you need to get rid of this in a hurry in the main repo I would just comment it out

I went ahead and just disabled the text and added a todo for #993 to reenable when the copy is reviewed

joshzarrabi:
@kkatzen Awesome, I guess the only problem is that we lose the all measure all together right now. How hard would it be to leave that in the table for now?

kkatzen:
> @kkatzen Awesome, I guess the only problem is that we lose the all measure all together right now. How hard would it be to leave that in the table for now?

oh good catch- reverted that change and added a TODO with the new code in the comment also linked with 993 :)

joshzarrabi:
Thank you!

joshzarrabi (approved):
Thank you so much!

PR #1020, opened 2021-07-30
properly cite UHC sources

### LINKED ISSUE #1019: Properly cite UHC sources
Right now we are not properly citing UHC sources when we use their data in a map or graph. We should fix this.

joshzarrabi:
resolves #1019


benhammondmusic (approved):
looks good

PR #1023, opened 2021-07-30
Ben dropdown - place existing variables into categories

benhammondmusic:


joshzarrabi (changes requested):
Do we want to check in the `package-lock` changes?

frontend_server/.env
```diff
@@ -0,0 +1,14 @@
+# This is an example file to use for local development.
```
joshzarrabi, line 1: I don't think we want this file either

joshzarrabi (changes requested):

joshzarrabi:
Thanks Ben, Looks great outside of the extra files you committed :)

PR #1030, opened 2021-08-03
Dropdown - place existing variables into categories

benhammondmusic:
- remade PR without extra committed files
- adds categories array and maps over it to place data variables into those dropdown categories
- allows for future variables to be added easily, and allows a variable to exist in multiple categories if desired


joshzarrabi (approved):
Looks great, thanks!

benhammondmusic:
Comments from @efregoso and Chi that should be implemented / mocked up:
- no need for left margin on list items, as the category heading is already bold
- should implement dropdown sub-menus for when the number of variables becomes overwhelming to the user. This is a tradeoff for "hiding" the options behind the category label, so perhaps a good compromise would be to use the sub menus for closely related items (e.g. variants of the Corona virus, or all mental health conditions, etc. Anything the target user would expect to find under that category heading)

benhammondmusic:
Turned out I couldn't just unindent the list items, because the hover/selected state creates a grey box with significant padding around it. The other option to align the category heading with the category item was to indent the heading to match, as shown in the pic. 
options selector categories in line

benhammondmusic:
Here's with a "dense" list, tighter vertical margins on the list items; I think I prefer it.
dense options selector

benhammondmusic:
Dense vertical layout for list items:
dense options selector

benhammondmusic:
Dense vertical layout; even spacing from the category. Need to confirm on mobile / single column display 
dense proper category spacing

PR #1031, opened 2021-08-03
gitignores frontend/packagelock

benhammondmusic:


joshzarrabi (approved):
Thanks!

benhammondmusic:
Decided to close; makes sense for us to commit our package lock file so we are using the same versions of dependencies

PR #1032, opened 2021-08-03
Friendly tab routes, with back button support

### LINKED ISSUE #717: Back button could navigated back to same tab
If I'm on the non default tab in About Us, click to a different page then click back, I end up back on the default tab instead of the one I was on. It would be nice if it took me back to the tab i was just on

benhammondmusic:
Fixes #717

- Loading any tab populates a query param in the address bar, yielding a unique link for every sub tab (for sharing and external linking)
- query param displays user friendly word rather than numerical tab index
- manually intercepts back button and force refreshes the page; this hack to compensate for react router not recognizing changes in params as actual page changes. 
- adds keys to mapped arrays to fix react warnings

### DIFF
--- a/frontend/src/pages/AboutUs/AboutUsPage.tsx
+++ b/frontend/src/pages/AboutUs/AboutUsPage.tsx
@@ -13,9 +13,9 @@ import {
 import { useHistory } from "react-router-dom";
 import { friendlyTabNamesAboutUs } from "../../utils/urlutils";
 
-export const ABOUT_US_PROJECT_TAB_INDEX = 0;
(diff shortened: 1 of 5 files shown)


joshzarrabi:
Is there any way you could get the forward button working?

benhammondmusic:
I didn't have any luck enabling forward. I also realized there are tabs on the "data methodology page" so I'll need to add those.

benhammondmusic:
closing for now; refresh isn't great and forward functionality not working. Either need to replicate what is happening in the "explore data" page, or potentially choose full url routes for the tabs (eg healthequitytracker.org/faq)

PR #1034, opened 2021-08-05
maps listen for full click rather than just mousedown. 

### LINKED ISSUE #1033: Right clicking map causes a left-click "drill down" behavior
on Brave/MacOS. Seems like right-clicking should provide the OS context menu, or eventually some export/share options.

https://user-images.githubusercontent.com/41567007/128257225-1fe2e572-258f-4275-a409-1d99f30f111d.mov

benhammondmusic:
allows right clicking on maps to provide OS context menu with img save, inspect element, etc.
full click behavior is more expected rather than just mousedown 

fixes #1033


joshzarrabi (approved):
Thanks!

Issue #1037, opened 2021-08-06
FAQ-definitions headings display


benhammondmusic:
On mobile (Brave iOS and others), the section headings under Definitions on the [full FAQ page](https://healthequitytracker.org/whatishealthequity?tab=1) are vertically squashed together and unreadable. 

![image](https://user-images.githubusercontent.com/41567007/128560381-16801c59-d894-4186-a933-ffc189d6207f.png)

PR #1056, opened 2021-08-11
A11y updates from spreadsheet

benhammondmusic:
Fixes and addresses many of the a11y concerns identified on the google doc spreadsheet including:
- blank alt text for decorative images
- descriptive alt and/or aria labels for screen reader users
- proper color contrast
- responsive spacing
- proper alert msgs
- nest links causing multiple issues
- render static images instead of animations when a user pref is used-reduce-motion


benhammondmusic:
Fixes #1037

frontend/src/pages/Landing/LandingPage.tsx
```diff
@@ -92,34 +96,30 @@ function LandingPage() {
...
-                </LinkWithStickyParams>
+                  Explore the Health Equity Tracker
+                </Button>
               </Box>
             </Grid>
             <Grid item xs={12} sm={12} md={6} className={styles.HeaderImgItem}>
               <img
                 src="img/shutterstock_1414416191 2 (1).png"
                 className={styles.HeaderImg}
-                alt="A man and woman laying with their two children"
```
joshzarrabi, line 122: Did you get rid of this on purpose?

frontend/src/pages/Landing/LandingPage.tsx
```diff
@@ -208,7 +225,7 @@ function LandingPage() {
                 <img
                   src="img/118172286-e3fffb80-b3c7-11eb-9553-127462881af3.png"
                   className={styles.PrioritizeHealthEquityImg}
-                  alt="Three women embracing a small baby in between them"
```
joshzarrabi, line 211: Same here

joshzarrabi (approved):

joshzarrabi:
Looks good, just curious about why you removed some of the alt text

frontend/src/pages/Landing/LandingPage.tsx
benhammondmusic: It's a judgement call, but from what I've researched and the guidance given on that A11y spreadsheet, alt text for images should be a blank string `alt=""` for any decorative images, making the screen reader intentionally skip the element rather than try and make sense of it. Although the photos provide an inclusive atmosphere for sighted users, someone using a screen reader might be confused as to why the photograph is there, and if it serves a functional or informational purpose (which I decided they don't).

benhammondmusic:
Text on buttons removed from outer component need centering image

frontend/src/pages/Landing/LandingPage.tsx
joshzarrabi: Cool, makes sense to me

benhammondmusic:
closes #1037

PR #1077, opened 2021-08-17
take ALL out of the table card and show it under the map

### LINKED ISSUE #993: Puerto Rico Total Cases by Race and Ethnicity

joshzarrabi:
fixes #993


benhammondmusic:
#1118 closes this PR

PR #1079, opened 2021-08-17
replaces certain unknown race/ethnicity "AND"s with "OR"s for accuracy

benhammondmusic:


benhammondmusic:
changes:
- TITLE and ALERT on UNKNOWNS MAP
 - ALERT on RACIAL DISPARITY BAR CHART
 
 works on all 3 carousel settings

benhammondmusic:
changes and to or

benhammondmusic:
and or grammar

benhammondmusic:
fixes #1081

joshzarrabi:
I'm still seeing the wrong end of the sentence on the map card.

Screen Shot 2021-08-17 at 5 25 19 PM

benhammondmusic:
Hmm I'll look

benhammondmusic:
@joshzarrabi 
What did you think it should change to? I think this is correct; should we have it read "where EITHER race or ethnicity was unknown" 

It's singular so I think it should be "was" not "were"

joshzarrabi:
@benhammondmusic My bad,  youre totally right

joshzarrabi:
@benhammondmusic I think adding `either` would be good

joshzarrabi:
@benhammondmusic Also I noticed `sex` and `age` are incorrectly picking up the change.
Screen Shot 2021-08-17 at 5 44 37 PM

benhammondmusic:
ah bummer; i thought i'd checked that. will look into it

benhammondmusic:
@joshzarrabi OK updated it to only affect the correct toggle-tab setting

frontend/src/reports/TwoVariableReport.tsx
```diff
@@ -211,6 +211,7 @@ function TwoVariableReport(props: {
           updateFips: (fips: Fips) => void
         ) => (
           <UnknownsMapCard
+            overrideAndWithOr={true}
```
joshzarrabi, line 214: do we need to make sure this is only true in the case of race_and_ethnicity as well?

benhammondmusic: Yes, good catch. I'll fix that

joshzarrabi (approved):

PR #1101, opened 2021-08-20
Highest/Lowest Box: Shortens label to prevent clipping when viewing on mobile

benhammondmusic:
also changes the verb when menu expands/collapses to guide the user


joshzarrabi (approved):
Perfect, thank you Ben!

Issue #1036, opened 2021-08-06
Add vaccination Information


joshzarrabi:
### Background

Users would like to see the demographic breakdown of vaccination information in the same way they can see other conditions. 

### Sources

We can source the vaccination information from the following locations

* National data: https://data.cdc.gov/Vaccinations/COVID-19-Vaccination-Demographics-in-the-United-St/km4m-vcsb
* State data: https://www.kff.org/other/state-indicator/covid-19-vaccinations-by-race-ethnicity/?currentTimeframe=0&sortModel=%7B%22colId%22:%22Location%22,%22sort%22:%22asc%22%7D
* County level: (no demographic information): https://data.cdc.gov/Vaccinations/COVID-19-Vaccinations-in-the-United-States-County/8xkx-amqh

PR #1116, opened 2021-08-26
Vax backend

joshzarrabi:


config/data_sources/AGG_cdc_vaccination_county.sql
```diff
@@ -0,0 +1,30 @@
+-- County-level joins with ACS population tables. Fir we join with
```
benhammondmusic, line 1: Typo in comment: "first"

benhammondmusic (approved):
Awesome work. I didn't see anything I obvious, but excited to play around with it with the Frontend changes

python/datasources/kff_vaccination.py
```diff
@@ -0,0 +1,192 @@
+import pandas as pd
+
+from ingestion.standardized_columns import Race
+import ingestion.standardized_columns as std_col
+
+from datasources.data_source import DataSource
+from ingestion import gcs_to_bq_util, github_util
+
+BASE_KFF_URL_TOTALS_STATE = ('https://raw.githubusercontent.com/KFFData/COVID-19-Data/'
```
benhammondmusic, line 9: Do you think this is fragile? Is KFF or GitHub likely to change the way this file is stored? Obviously it's good for now but just thinking we should have it in mind that this might be a point of failure or whatever

joshzarrabi: Yes I do think its fragile but I'm not sure what to do except for wait for it to fail and change it. I could add more error cases.

python/datasources/kff_vaccination.py
```diff
@@ -0,0 +1,192 @@
...
+    output_row[std_col.VACCINATED_PCT_SHARE] = str(state_row_pct_share[generate_pct_share_key(race)].values[0])
+
+    if race in KFF_RACES_PCT_TOTAL:
+        output_row[std_col.VACCINATED_PCT] = str(state_row_pct_total[generate_total_pct_key(race)].values[0])
+
+    return output_row
+
+
+def generate_total_row(state_row_totals, state):
+    """Generates the total vaccinated percentage row for a givcen state
```
benhammondmusic, line 122: typo in comment: "given"

benhammondmusic (approved):
Sweet

python/datasources/kff_vaccination.py
benhammondmusic: How is it handling an error currently; say for instance KFF renames that file in their GitHub or something; what will a user of our site see ?

benhammondmusic:
#1036

python/datasources/kff_vaccination.py
joshzarrabi: Nothing, the data pipeline will just fail and we won't be able to update the data until we fix it. The error handling will just be for us to more quickly find the issue.

joshzarrabi: Absolute worst case scenario == The pipeline works but somehow gets the wrong data, and never throw an error to stop it. This is possible I guess but very unlikely.

joshzarrabi: These data pipelines run asynchronously whenever we want them to (when we get new data). They then update the s3 buckets that the frontend talks to. If the pipelines fail then the buckets just dont get updated.

benhammondmusic: Oh right; so error wouldn't really be user facing anyways. Doesn't seem like as big a deal

PR #1118, opened 2021-08-27
Move "Overall" metric to Map Card; coordinate with design

### LINKED ISSUE #993: Puerto Rico Total Cases by Race and Ethnicity

benhammondmusic:
This PR takes over #1077 and fixes #993

- moves the "ALL" metric from the table to the map
- changes the "compare across" alert to be a clickable link since it isn't really an alert
- Places "OVERALL" and "COMPARE ACROSS" inline to associate the two options and brings the user's attention to the group/demographic breakdown options 
- changes "compare" to "visualize" which gives the user a better idea of what to expect from clicking the link
- hides the "OVERALL" number if data isn't available. missing data Alert is already displaying in that case
- better to render the full group name, eg "Native Hawaiian and Pacific Islander (Non-Hispanic)". The clarity is nice and outweighs concerns over being listed directly above in the dropdown button, and duplicate text adding additional visual noise

131171105-91731bf5-82a3-4ea7-a0c6-884a35d3f280

overall county no data

overall national selected group specified


benhammondmusic:
Actually, now that we are providing INFO in the form of the occurrences/100k, and INFO ALERT box makes sense and I think is attractive / fits the design.

alert OVERALL

joshzarrabi:
This looks excellent, thank you for putting so much thought into this :)

PR #1123, opened 2021-08-31
Removes unneeded words "Total" from cards; Adds "100k" to table

### LINKED ISSUE #1122: "TOTAL" or "per 100k" on card headers, but not both
on Explore Data; several of the cards (containing maps/charts/tables) include the word "TOTAL" even though they are representing a per 100k metric. 

Additionally the table display (bottom left on desktop) is lacking the "per 100k" wording in the title even though that is what is being displayed

REQUIRED CHANGES:
- remove the word "TOTAL" from top left map, the bottom left table, and the middle right horizontal bar chart
- add the words "Per 100k People" to the bottom left table
TOTAL or PER 100k for card headers

benhammondmusic:
fixes #1122 

This brings the COVID language in line with all other variables (eg COPD, uninsured individuals) where the word TOTAL is only used when referring to the SHARE OF TOTAL

### DIFF
--- a/frontend/src/cards/TableCard.tsx
+++ b/frontend/src/cards/TableCard.tsx
@@ -26,6 +26,7 @@ import { Row } from "../data/utils/DatasetTypes";
 import MissingDataAlert from "./ui/MissingDataAlert";
 import Alert from "@material-ui/lab/Alert";
 import Divider from "@material-ui/core/Divider";
+import { ALL } from "../data/utils/Constants";
 
 export interface TableCardProps {
   fips: Fips;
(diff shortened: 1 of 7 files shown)


benhammondmusic:
It looks like for all other variables (eg COPD, uninsured individuals, etc) the word TOTAL is already not used; this was only appearing on COVID

benhammondmusic:
https://user-images.githubusercontent.com/41567007/131555409-b18ee241-35d8-4c9e-8001-113d1ec7ea7c.mov

frontend/src/cards/TableCard.tsx
```diff
@@ -71,7 +71,7 @@ export function TableCard(props: TableCardProps) {
     <CardWrapper
       queries={[query]}
       title={
-        <>{`${props.variableConfig.variableFullDisplayName} By ${
+        <>{`${props.variableConfig.metrics.per100k.fullCardTitleName} By ${
```
joshzarrabi, line 74: I don't think we should make this change, because the table shows more than just per 100k metrics.

frontend/src/data/config/MetricConfig.ts
```diff
@@ -170,12 +170,12 @@ export const METRIC_CONFIG: Record<string, VariableConfig[]> = {
...
       variableId: "cases",
       variableDisplayName: "Cases",
-      variableFullDisplayName: "Total COVID-19 Cases",
+      variableFullDisplayName: "COVID-19 Cases", //Total
       metrics: {
         count: {
           metricId: "covid_cases",
-          fullCardTitleName: "Total COVID-19 Cases",
-          shortVegaLabel: "total COVID-19 cases",
+          fullCardTitleName: "COVID-19 Cases", //Total
```
joshzarrabi, line 177: Not sure what these comments add

joshzarrabi (changes requested):

frontend/src/data/config/MetricConfig.ts
benhammondmusic: those were temporary (where I pulled them out of) until approval from you guys that we removed the correct "totals". will def remove them entirely once we are happy with the results

benhammondmusic: removed

joshzarrabi (approved):

PR #1126, opened 2021-08-31
Vaxx: remove demographic UI when data is missing

### LINKED ISSUE #1113: Allow disabling toggle-bar options when breakdowns are unavailable (race/age/etc)

benhammondmusic:
fixes #1113


benhammondmusic:
disable

This is showing poverty just as an example, it will be for Vaccinated once that is in place. 

Currently it DISABLES (grays out and doesn't allow clicking) the demographic toggle, and it also stops the dropdown SELECT GROUP and the COMPARE ACROSS XYZ links/buttons from rendering

joshzarrabi:
I was just going to suggest that, thanks Ben

PR #1137, opened 2021-09-03
Vax frontend

joshzarrabi:


frontend/src/cards/ui/UnknownsAlert.tsx
```diff
@@ -47,9 +59,33 @@ function UnknownsAlert(props: {
...
       } was unknown.`;
 
+  const raceEthDiffMapText = `In cases where race and ethnicity are reported
+    separately, the map shows the higher of the two metrics.`;
+
   const percentageUnknown = unknowns[0][props.metricConfig.metricId];
 
-  return (
+  const diffRaceEthnicityText = raceEthnicityDiff
+    ? `This state reports race and ethnicity seperately.
```
benhammondmusic, line 68: Typo, I think this is the one Allyson noticed in the demo maybe? "sepArately"

frontend/src/data/config/MetricConfig.ts
```diff
@@ -408,4 +412,37 @@ export const METRIC_CONFIG: Record<string, VariableConfig[]> = {
       },
     },
   ],
+  vaccinated: [
+    {
+      variableId: "vaccinated",
+      variableDisplayName: "Vaccinated",
+      variableFullDisplayName: "Vaccinated Individuals",
```
benhammondmusic, line 419: Site-wide it seems like we are inconsistent with when we use PEOPLE vs INDIVIDUALS. Is there a pattern, or is this something we should standardize?

frontend/src/reports/ReportProvider.tsx
```diff
@@ -132,6 +132,9 @@ function ReportProvider(props: { madLib: MadLib; setMadLib: Function }) {
         className={styles.MissingDataInfo}
         ref={fieldRef}
       >
+        <h3> Notes </h3>
```
benhammondmusic, line 135: I think we should redo this wording / layout but I can work on it after it's merged in. I think it's important to keep the "What Data Are Missing" larger up top. Then I think we should include the note about 1-dose but phrased as a missing data problem; as in, we would include the "fully" dosed people if they had the breakdown by demographic, correct? The only reason we are using single dosed is because that's the category that includes demographic breakdown. Do we also need to mention that data is missing from the CDC at county level and that we rely on a different data source?

benhammondmusic: maybe the specific details will be on the methodology page I guess? but I still like the idea of keeping this note as a missing data problem if we can, and under the larger heading that's already there

benhammondmusic: should be plural ? 

> "Vaccinated Individuals" refers to people who have received at least one dose of a COVID-19 vaccine.

benhammondmusic (approved):

benhammondmusic:
closes #1036

frontend/src/reports/ReportProvider.tsx
joshzarrabi: For sure, I do think we need to put a conspicuous note on the page explicitly saying that we are talking about at least one dose though.

joshzarrabi: And yeah I should add a vaccine section to the "what data is missing" , will do that,

frontend/src/data/config/MetricConfig.ts
joshzarrabi: Good point, I just copied from "uninsured individuals", but we should talk about it and decide what we want.

frontend/src/cards/ui/UnknownsAlert.tsx
joshzarrabi: thanks

frontend/src/reports/ReportProvider.tsx
joshzarrabi: @benhammondmusic where do you think we should put this note?

benhammondmusic: I'm almost thinking having a large heading (same size as "what data are missing" called "definitions" or something to that effect, and then we define our variables or otherwise clarify what the terminology means. We could also include our note about using inclusive language re "some other race" there later on maybe? 

I think it looks a little strange to have a smaller subheading first for "notes" and then the large heading "what data are missing" below it.

joshzarrabi: I agree with that, but I kind of think it should be somewhere more prominent on the page than in the middle of that large block of text.

joshzarrabi:
Keeping this in draft until we have the backend stuff on prod.

frontend/src/reports/ReportProvider.tsx
benhammondmusic: Yes, definitely. I was thinking up top, like this. We could also consider finding a home higher on the page for important definitions or clarifications if they are essential to understanding the visualizations 

definitions vaxx

frontend/src/data/variables/MergedPopulationProvider.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  private async getDataInternalWithoutPercents(
+    breakdowns: Breakdowns
+  ): Promise<IDataFrame> {
+    const mergedPopDataset = await getDataManager().loadDataset(
+      this.getDatasetId(breakdowns)
+    );
+    let mergedPopDataFrame = mergedPopDataset.toDataFrame();
+
+    // If requested, filter geography by state or coacs2010ty level
```
benhammondmusic, line 56: is that supposed to be "county" ?

joshzarrabi: This file shouldnt be here LOL

benhammondmusic:
For Vaccine, at the state level, our population comparison numbers are provided by KFF
Unfortunately they do not include numbers for AIAL nor NHPI groups.
In an effort to provide the best data possible, these groups will have the original ACS data as a population comparison metric.
To delineate the different sources, the disparity chart will use a different color/transparency and a unique legend title/color

benhammondmusic:
BUG: "at the [] level"
guam wording error

benhammondmusic:
Some race options on dropdown were crashing site; refactored to conditionally calculate and render the demographic total phrase when possible based on available data

https://user-images.githubusercontent.com/41567007/136106657-6a8c0597-5372-48ad-9a8e-01f083e5df43.mp4

benhammondmusic:
I think we need to make it more clear that this is COVID vaccinations... like right now it makes sense because everyone is talking about getting "vaccinated" and people know it's against covid. But there are lots of other vaccines people can and should be getting that tie in with health equity generally and health outcomes. As we edge away from being "just a covid tracker" I think we need to keep this in mind and clarify. @jgonzalezmsm @joshzarrabi

benhammondmusic:
Merged dropdown categories redesign and made responsive across devices:

Issue #716, opened 2021-04-29
Clean Vega Editor Opening


kkatzen:
Opening in the vega editor exposes the "*~*" delimiter hack used to get Vega labels on multiple lines

PR #1140, opened 2021-09-07
3-dot (Vega Actions Menu) improvements

benhammondmusic:
- better naming for saved png files
- screen reader labels
- removes SVG, VegaEditor and View Source options


benhammondmusic:
Some screenshots to preview these updates:
- only leave "Save as PNG" in the 3-dot menu
png only option

frontend/src/cards/ui/MultiMapDialog.tsx
```diff
@@ -104,6 +104,9 @@ export function MultiMapDialog(props: MultiMapDialogProps) {
                     hideActions={false}
                     scaleType="quantile"
                     geoData={props.geoData}
+                    filename={`${props.metricConfig.fullCardTitleName}${
+                      breakdownValue === "All" ? "" : ` for ${breakdownValue}`
```
joshzarrabi, line 108: This is generating "COVID-19 Cases Per 100K People in United Statesfalsefalse - Health Equity Tracker.png" for the total US map

joshzarrabi (changes requested):

frontend/src/cards/ui/MultiMapDialog.tsx
benhammondmusic: Issue was in MapCard.jsx; can't use short circuit evaluation inside a string template literal like: 
```
`${someBoolean && "print this if true"}` 
```
and instead need to use a ternary:
```
`${someBoolean ? "print this if true" : ""}`
```
to prevent JS from coercing a Boolean shortcircuit into a string (eg `false` into "false")

benhammondmusic:
Closes #716 as we will no longer have the Vega Editor option available so users won't see the ~ hack referenced. Unclear if seeing that hack is really an issue regardless

joshzarrabi (approved):
Looks good, I wouldnt worry about the hack thing.

benhammondmusic:
Closes #1141

PR #1149, opened 2021-09-10
adds commas w oxfords and "AND" between multiple cited data sources

### LINKED ISSUE #972: Add commas and "and" between data sources in footers
See todo in Sources.tsx:

Screen Shot 2021-06-03 at 11 51 50 AM

benhammondmusic:
closes #972 

Does not add a period at the end of the citation; is that wanted as we are using periods at the ends of our alerts? Not sure standard practice for data citations.


joshzarrabi:
Looks good to me

joshzarrabi (approved):

benhammondmusic:
I changed my mind and it DOES add a final period because it's basically a sentence. We can easily remove if this is inappropriate in a scientific context

Issue #1166, opened 2021-09-17
remove "SVG" from FAQ


benhammondmusic:
"How do I share the graphs?
Next to each graph, there is a circle-shaped button with three dots in it. Click on that button to see a menu where you can save each graph as a PNG or SVG."

PR #1167, opened 2021-09-17
fix react warning on FAQs; make PSHandler logs development only

benhammondmusic:
the `parse()` command is interpreting a string into HTML; only JSX needs `className` ; using `class` fixed that warning

also, the `PShandler` logs were going into the console in production which doesn't seem like great practice.


benhammondmusic:
closes #1166

joshzarrabi (approved):

PR #1168, opened 2021-09-17
adds alert when population pct adds up to over 100%

### LINKED ISSUE #1165: Show note on disparity bar chart explaining why total population adds up to over 100%
### Problem

Whenever the disparity bar chart is displaying population groups that include Hispanic, as well as hispanic itself (like for poverty or health insurance), the total population adds up to over 100%, because Hispanic individuals are being counted twice. 

Screen Shot 2021-09-16 at 10 42 44 AM

*We have heard feedback* that this confuses people and makes them think we don't know how to do math. 

### Proposed solution

Below any disparities bar chart that displays races that include hispanic, and thus the population %'s will add up to over 100%, we should put a note explaining why, something like:

*Population percentages add up to over 100% because the racial categories reported for poverty include Hispanic individuals, so hispanic individuals are being counted twice.* (or something better if you can think of it)

### How will this help policy influencers improve health equity

If a policy influencer looks at this and thinks that we are improperly calculating our statistics they may be weary of using our data to push for policy that improves health equity.

joshzarrabi:
fixes #1165


frontend/src/cards/DisparityBarChartCard.tsx
```diff
@@ -79,6 +80,19 @@ function DisparityBarChartCardWithKey(props: DisparityBarChartCardProps) {
               row[props.breakdownVar] !== UNKNOWN_RACE
           );
 
+        let shouldShowDoesntAddUpMessage = false;
+        if (
+          props.breakdownVar === "race_and_ethnicity" &&
```
benhammondmusic, line 85: Can we import this string from metric config just in case it ever changes in the future?

benhammondmusic (approved):
Glad we are adding this

frontend/src/cards/DisparityBarChartCard.tsx
joshzarrabi: good call

joshzarrabi:
FWIW, I decide not to add up the percentages to see if they are over 100% because some of our metrics round out to 101%. Also I think this is more semantic, if there is any duplication of hispanic people we should show that message.

PR #1172, opened 2021-09-17
use Sass color variables

### LINKED ISSUE #1169: share variables between SASS and React / Vega
right now we have many colors, pixel widths and other values hard-coded throughout the app. It would be great to utilize SASS variables for all of these values, for consistency and code-health. However, since we are generating the Vega viz via React, we need to export those SASS variables and then import them into the React side. This looks like an easy setup for that: https://css-tricks.com/getting-javascript-to-talk-to-css-and-sass/

benhammondmusic:
Using a discrete set of consistent design values (eg colors, pixels widths) saves time and prevents drift as the app expands and changes. This PR replaces all hard-coded hex values in SCSS files with with $sass-variables, and also closes #1169 exporting those variables for use in React / Vega visualizations.


benhammondmusic:
https://mattferderer.com/use-sass-variables-in-typescript-and-javascript

benhammondmusic:
closes #1169

benhammondmusic:
Should adjust to only export and type the select colors that are needed in the HTML/vega.

Should also confirm if we can reduce some of these colors... I believe some were typos and some others were "close enough" versions of the same colors...

joshzarrabi:
Sorry for letting this sit for so long!

joshzarrabi:
Great work!

joshzarrabi (approved):

PR #1177, opened 2021-09-24
Population Card - match rest of site

### LINKED ISSUE #1142: No "Export as PNG" on Population Card Charts
The population card charts currently don't offer the "3-dot" Vega Actions menus. Since we are keeping the 3-dot menus on our other maps/charts (until we have a better share option), should we be consistent and enable "Export as PNG" on the population card Vega visualizations as well?
png only option
pop card doesnt have 3 buttons

benhammondmusic:
- improve spacing and labels 
- adds EXPORT menus to match other visualizations ( closes #1142 )
- use "race and ethnicity" instead of just "race"


joshzarrabi (approved):
Looks great, thank you!

PR #1180, opened 2021-09-28
uses AAPI category for states that use that categorization when

joshzarrabi:
reporting vaccines


joshzarrabi:
Waiting for a go ahead on naming before we merge this.

joshzarrabi:
We have decided to use "Asian, Native Hawaiian, and Pacific Islander" as the descriptor for this category

PR #1182, opened 2021-09-29
Use single column layout

### LINKED ISSUE #1351: Need to reorder cards
Since we'll be using single column, we need to adjust the card order. Per team meeting, proposed is:

- 100k map
- 100k graph
- unknown map
- disparity graph
- data table

benhammondmusic:
closes #1351 

PROBLEMS
- Seeing two columns on the first carousel setting is confusing and overwhelming; by instinct the user wants to make side by side comparisons despite the graphs/charts not being equivalent in the two columns (and thereby making the 2nd and 3rd compare settings more intuitive by implementing the side-by-side view)
- Using only half of the available screen width shrinks our horizontal bar charts, making them harder to read and masking disparities (particularly in the most marginalized communities)
- Single column view is already in place for small screens like mobile; potentially confusing to users who might visit on multiple devices to have the layout so significantly different (as is)

SOLUTION
- For the single variable / single geography carousel setting, cards should be in one vertical column rather than two. For the 2nd and 3rd "compare" settings on the carousel, it will switch to a two column side by side layout
- Adding some padding around the single column prevents the maps from growing too large, and potentially gives us room for "notes" and other editorial content in the margins
Screen Shot 2021-11-11 at 10 05 17 AM


benhammondmusic:
Another bonus to this is that our charts are currently all horizontal; having more room will better reflect the data; and disparities between groups; particularly for those groups with the smallest population.

joshzarrabi:
@benhammondmusic should we add a max width? It gets kind of wide on big screens. I might have said the opposite earlier but I do think the proportions can get a little weird if its too big

benhammondmusic:
> @benhammondmusic should we add a max width? It gets kind of wide on big screens. I might have said the opposite earlier but I do think the proportions can get a little weird if its too big

sets max width for the report (everything under the madlib) to be 1000px max for single column, 2000px for double column. 

## 4k
4k 2col
4k 1col

## 2k
2k 2col
2k 1col

## 1k
1k 1col
1k 2col

benhammondmusic:
Videos

https://user-images.githubusercontent.com/41567007/144650434-8baef9c0-ded0-4e80-933b-0e5f98618b0b.mov


https://user-images.githubusercontent.com/41567007/144650456-65d0f460-9356-4568-a883-0cb1c90d7803.mov


https://user-images.githubusercontent.com/41567007/144650480-7bc95778-bff4-4300-bdea-4ee41931ec6e.mov

joshzarrabi (approved):

PR #1185, opened 2021-09-30
only get kff one_dose numbers if they are not NA

joshzarrabi:


benhammondmusic (approved):
Looks good from what I can tell. I don't know what [~] is in Python though

joshzarrabi:
> Looks good from what I can tell. I don't know what [~] is in Python though

@benhammondmusic Its a pandas operator meaning `not`. 

See https://pandas-docs.github.io/pandas-docs-travis/user_guide/indexing.html#boolean-indexing

PR #1187, opened 2021-10-01
show alt population comparison metric when needed

benhammondmusic:
For Vaccine, at the state level, our population comparison numbers are provided by KFF
Unfortunately they do not include numbers for AIAL nor NHPI groups.
 In an effort to provide the best data possible, these groups will have the original ACS data as a population comparison metric. 
To delineate the different sources, the disparity chart will use a different color/transparency and a unique legend title/color


benhammondmusic:
light blue

PR #1189, opened 2021-10-04
Improvements to TableChart

benhammondmusic:
- sets default `isSorted` col, which makes that sort arrow visible on load. Should clear up confusion a bit with the "hidden" sort function and cue user that the options exist. GOAL: clear up confusing UI
- increases rows displayed before pagination appears, effectively removing unneeded pagination for our current data. If in the future we display something with longer number of rows (right now >20) the pagination will automatically re-display. GOAL: reduce visual noise by removing unused buttons and information from screen.

table tweaks


joshzarrabi (approved):

benhammondmusic:
Adds hover tooltips to explain "toggle" and "sort by" options 

https://user-images.githubusercontent.com/41567007/135887670-a62f4ca1-5679-4a9d-8492-d278158cb279.mov

Issue #1202, opened 2021-10-08
Calculate percent share of population on the backend


joshzarrabi:
## Background

The slowest operation on the site, by far, is when the user clicks into a state’s COVID numbers for the first time. The state map spins for a while (sometimes like ten seconds), and then the browser locks up for a few seconds and then the state map appears.  On worse computers the browser will completely crash while viewing this in compare mode with other heavy programs running (like Microsoft Teams).

Clearly, this is not a good user experience, but we have dealt with it for the last few months because we wanted to add other, more health equity minded things to the site, like territory and vaccine data. Also, this is only a real problem the first time a user clicks into a state, as we cache everything for every subsequent state and county. 

## Technical Info

The specific reason that loading that page is so slow is described below.

The order of operations is

1. Download the COVID by county and race dataset
1. Download the ACS population by county and race dataset
1. Merge the datasets

Once all of the steps are complete we are able to show the user what they want to see. 

The real problems are steps 2 and 3. It takes a long time to download the large ACS dataset, which is represented on the browser by the spinner, and then merging the datasets is what causes the browser to lock up and crash. 

## Why do we do this

The reason we download the ACS dataset at all is because we want to get the population comparison metrics for each county. Interestingly enough we *already* have the population numbers that we need, as we get them on the backend, but we don’t calculate the population percentages on the backend, instead doing them on the frontend. This is silly. 

## Proposal

We do all population calculations on the back end, thus removing the need for steps 2 and 3 all together, as they would all be merged into one, smaller dataset. This should provide a much better user experience for people on their first visit to the site. 

None of this applies to the national level (for now) because we do all of our aggregation there on the frontend. 

## Next steps

Get rid of as many merges and unnecessary calculations as possible

- [x] Don't do acs pct share calculations on frontend acs provider
- [x] Don't do acs pct share calculations on fronentend acs2010 provider
- [x] #1281
- [x] UHC

joshzarrabi:
> This will close #953 right?

Yeah

jgonzalezmsm:
Josh completed proof of concept testing (small piece)
Noticed improvement in speed/performance.

(anecdotal - 10sec down to 1sec load !!)

Fixing rounding error before expanding

jgonzalezmsm:
move all % share calculations to backend - expand to all variables

joshzarrabi:
Because UHC data is not separated out based on national/state, I think we should wait until we generate the national level calculations on the backend to do it.

[...]

PR #1219, opened 2021-10-14
Pop pct backend

joshzarrabi:
Calculate the population percentage on the backend, and merge to COVID datasets in big query

* This allows us to avoid doing the costly procedures of downloading the acs dataset, calculating the pct share, and merging with the other datasets on the frontend.
* Only done for COVID right now because that was causing the biggest problem, will do the rest in subsequent pr's

#1202


config/run.tf
```diff
@@ -55,7 +56,8 @@ resource "google_cloud_run_service" "gcs_to_bq_service" {
 
         resources {
           limits = {
-            memory = "4G"
+            memory = "8Gi"
```
benhammondmusic, line 59: i assume it's fine to use `G` or `Gi` for these units?

benhammondmusic (approved):
👍

config/run.tf
joshzarrabi: Yeah its fine

PR #1233, opened 2021-10-18
Improve Vega Legend (Gray Legend = No Data, Layout Flow)

### LINKED ISSUE #1069: Add [GREY CIRCLE = No Data] as explicit entry to chart legend
Right now the "no data" regions are colored gray, but it is never explained to the user explicitly that's what grey means. It is reflected in the hover tooptips, which are not available to mobile users nor potentially some keyboard / screen reader users. 

I think that it would be a worthwhile feature to include this list item in the existing Vega legend (see non-styled mockup below)

no data legend

benhammondmusic:
Multiple improvements to Vega legends including:
- [x] adds "No Data" to legends on Small Multiples maps
- [x] Responsive layout to adjust dynamically based on screen size; allows small multiples to display in optimal way across all devices
- [x] adds "No Data" to Unknowns Maps
- [x] add "No Data" to /100k Map 
- [x] coordinates fonts and shapes between legends 

closes #1069

1xs
2sm
5xl

### DIFF
--- a/frontend/src/cards/MapCard.tsx
+++ b/frontend/src/cards/MapCard.tsx
@@ -273,28 +273,15 @@ function MapCardWithKey(props: MapCardProps) {
 
                       {/* Compare across XYZ for all variables except vaccinated at county level */}
                       {!hideDemographicUI && (
-                        <span
-                          onClick={() => setSmallMultiplesDialogOpen(true)}
-                          role="button"
-                          className={styles.CompareAcrossLink}
-                          aria-label={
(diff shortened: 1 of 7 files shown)


benhammondmusic:
Got "No data" legend on other maps; had to use absolute positioning to line it up with the color legend. There is likely a more robust solution but as long as this is thoroughly tested in different variables/device widths it should be good to go

benhammondmusic:
A little closer between legends
closer

PR #1235, opened 2021-10-19
adds k units to 100k map (DEMO)

### LINKED ISSUE #1234: Consider using "k" units when using "100k" as metric
Made a quick demo image to see

use k figures on per 100k

benhammondmusic:
closes #1234


benhammondmusic:
Make these changes on:
- [x] 100k legend and tooltip
- [ ] 100k map alert for selected population
- [ ] 100k simple barchart hover tooltip, bar label, x-axis ticks

Leave the table info as more detailed number, eg `12,453 cases per 100k`

benhammondmusic:
closing; the only place this still might make sense is in the legend, but we will revisit as needed

PR #1239, opened 2021-10-20
Improves PROJECT TAB "Number of Variables" and "Number of Sources" card

benhammondmusic:
- adjusts wording to emphasize more than just COVID
- Original "15 variables" was likely `5 conditions x 3 demographic options`. This updates the dynamic calculation to include all COVID sub-variables, along with accounting for these demographic multipliers. Currently brings total variables to 24
- Adds hyperlinks to condition names directly to a national level report on the data tracker. Makes it easy to navigate directly to where a user might have interest


benhammondmusic:
project sources variables

PR #1256, opened 2021-10-28
Adds explicit "per 100k" units to the table and hi/low

benhammondmusic:
There has been ongoing confusion with `per 100k` metrics; namely mistaking them for absolute numbers (rather than treating as a rate) and summing the numbers. Perhaps adding the units on each table cell would be helpful (in the same way we add `%` to every cell and don't rely on the user to deduce that it's a percentage from the table heading label alone).

For design consistency this styles the `/100k` and `%` in the same smaller, lighter font


benhammondmusic:
Adds `per 100k` to highest lowest box as well, in matching font

Screen Shot 2021-11-11 at 10 32 33 AM

benhammondmusic:
Use `per 100k` rather than `/100k` for consistency and readability
Screen Shot 2021-11-11 at 10 35 23 AM

joshzarrabi (approved):

PR #1261, opened 2021-10-29
Retain Tracker Settings

### LINKED ISSUE #1197: Tracker shouldn’t reset to covid/us when navigating around site

benhammondmusic:
Problem: when a user has made detailed selections including multiple variables or geo locations, it can be frustrating for them to click on another page (eg "Methodology" for an explanation) and then click back to the tracker and realize that they have to input all of their selections again. 

Solution: use sessionStorage (similar to a cookie that gets erased whenever a user closes their browser window) to memorize 
- the carousel setting (eg "Investigate", "Explore Relationships")
- the condition and geographic settings for each carousel setting (eg "COPD" in "California")

Additionally: This PR removes some code that wasn't accomplishing anything, and also triggers the population of URL params (eg `?mls=1.covid-3.00&mlp=disparity` at the end of the address bar) which was missing in some instances

Considerations: 
- There is no longer a way to "reset" the tracker to the default settings, short of closing the window and reopening. This doesn't really seem like a problem for users, and only a potential annoyance for us in development.
- We have the option to use `localStorage` instead of `sessionStorage` which will remember the settings across the closing and reopening of windows. We need to discuss if this would be helpful or annoying for users.

Fixes #1197

### DIFF
--- a/frontend/src/pages/ExploreData/ExploreDataPage.tsx
+++ b/frontend/src/pages/ExploreData/ExploreDataPage.tsx
@@ -17,7 +17,6 @@ import {
   MADLIB_PHRASE_PARAM,
   MADLIB_SELECTIONS_PARAM,
   parseMls,
-  psSubscribe,
   setParameter,
   setParameters,
   SHOW_ONBOARDING_PARAM,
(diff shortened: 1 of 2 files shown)


benhammondmusic:
Video demo of making detailed selections and navigating on and off the tracker, with the most recent settings remembered.


https://user-images.githubusercontent.com/41567007/139514163-cf1a94f2-a95f-4ad0-a9d2-0c486ca0d788.mov

benhammondmusic:
Should consider sharing remembered fips and variables somehow between carousel settings

benhammondmusic:
not fully working. it appears all of the reading/writing params code happens many times at many stages.... would be best to streamline these first and implement the cache lookup as part of the new streamlined process rather than piling more code on top

Issue #1263, opened 2021-11-01
Proposal: remove "What Data Is Missing" Link from Methodology; organize info properly


benhammondmusic:
This link causes circular clicking for the user... the EXPLORE DATA page should be the tool that you interact with and get the charts, the METHODOLOGY page should be the source of further details, subtle distinctions, and meta-info. It doesn't make sense for a user on the EXPLORE page to click on "methodology" for further details, only to get there and click "what data is missing" and be taken back to the EXPLORE DATA page. 

Linking to specific sections across pages is doable, but somewhat clunky using React Router. It requires intercepting the `#missingDataInfo` in the URL and then using javascript to scroll the user to the appropriate section. Since the EXPLORE DATA page already has a lot of processing going on, the results aren't great and we should avoid where possible. 

PROPOSAL: Important clarifying details should live on the METHODOLOGY page, and a limited, select few details can be duplicated on the bottom of the EXPLORE DATA page when absolutely needed. This will reduce code complexity, streamline the user flow and identify to the user where exactly they should go for which information. 

<img width="948" alt="remove missingData link" src="https://user-images.githubusercontent.com/41567007/139738232-cda00061-aaea-4c19-8c7c-fcc08fc4bfdd.png">

PR #1264, opened 2021-11-01
Fixes issues with linking to specific page sections

### LINKED ISSUE #1192: internal anchor links not always working
links such as `healthequitytracker.org/exploredata#missingDataInfo` aren't functioning properly. This is due React Router. The internal link is handled by a `scrollIntoView()` function; we need to implement a useEffect on pages that might receive `#jump-to-section` links and intercept that hash and execute the scroll into view.

benhammondmusic:
fixes #1192

due to React Router it's not always possible to simply link to a page section using `<a href="page#specificSection">`; this PR fixes problematic links by intercepting the `#specificSection` and executing a scroll into view command, or simply passing the function with ref in scope down via props to the sending component button. I have tested all links throughout the site and they all are working, but further ongoing testing should be done in other browsers.

We also should consider #1263


frontend/src/cards/ui/HighestLowestList.tsx
```diff
@@ -104,8 +106,13 @@ export function HighestLowestList(props: HighestLowestListProps) {
       <p>All rates are reported as: {props.metricConfig.fullCardTitleName}</p>
       <p>
         Consider the possible impact of{" "}
-        <a href="#missingDataInfo">data reporting gaps</a> when interpreting the
-        highest and lowest rates.
+        <Button
+          onClick={() => props.jumpToData!()}
```
benhammondmusic, line 110: Maybe better to account for situations where the jumpTo method hasn’t been passed; or require it as a prop?

frontend/src/pages/DataCatalog/MethodologyTab.tsx
```diff
@@ -120,9 +121,9 @@ function MethodologyTab() {
                 <p>
                   For a description of some of the gaps in COVID-19 data, please
                   see the{" "}
-                  <a href={EXPLORE_DATA_PAGE_WHAT_DATA_ARE_MISSING_LINK}>
+                  <Link to={EXPLORE_DATA_PAGE_WHAT_DATA_ARE_MISSING_LINK}>
```
joshzarrabi, line 124: This one didn't work for me, it brought me to the middle of the exploredata page

joshzarrabi (changes requested):

frontend/src/pages/DataCatalog/MethodologyTab.tsx
benhammondmusic: weird, it's broken for me now; was working before. I wondered why we are linking back to this section at all, I made another issue for this but to make i feel like the INFO should live on the methodology and the DATA should live on the tracker...

joshzarrabi (approved):
Very fancy, lets look into not linking back to the tracker in a later issue.

frontend/src/pages/DataCatalog/MethodologyTab.tsx
benhammondmusic: Implemented some min heights for the cards; that should be in place anyways because cumulative layout shift is a) really annoying to the user and b) penalized by Google ranking. By adding min heights to the objects that take a while to load you can minimuze how much the page shifts around. This seems to have fixed / helped a lot with letting the link scroll to the info section properly

benhammondmusic:
Adding min heights to the cards bumped desktop lighthouse from a 35 to a 40; reducing CLS from 0.477 to 0.277

And it seems to fix the inaccurate link-to-section scrolling

before
after

PR #1270, opened 2021-11-02
dont round down to zero unless pop pct is less than .005

joshzarrabi:
- mimics what is currently on the frontend

[#1202]


python/ingestion/dataset_utils.py
```diff
@@ -13,7 +13,7 @@ def generate_pct_share_col(df, raw_count_col, pct_share_col, breakdown_col, tota
        total_val: The value representing 'ALL' or 'TOTAL'"""
 
     def calc_pct_share(record, total_value):
-        record[pct_share_col] = round((float(record[raw_count_col]) / float(total)) * 100, 1)
+        record[pct_share_col] = percent_avoid_rouding_to_zero(record[raw_count_col], total)
```
benhammondmusic, line 16: is it worth fixing the typo in the function name? `rounding_to`

python/ingestion/dataset_utils.py
```diff
@@ -37,6 +37,22 @@ def calc_pct_share(record, total_value):
     return pd.concat(with_pct_share).reset_index(drop=True)
 
 
+def percent_avoid_rouding_to_zero(numerator, denominator, default_decimals=1, max_decimals=2):
```
benhammondmusic, line 40: typo `rouding`

python/ingestion/dataset_utils.py
```diff
@@ -37,6 +37,22 @@ def calc_pct_share(record, total_value):
...
+    """Calculates percentage to `default_decimals` number of decimal places. If
+       the percentage would round to 0, calculates with more decimal places until
+       either it doesn't round to 0, or until `max_decimals`. `defaultDecimals`
+       and `maxd_decimals` should be >= 0 and `max_decimals` should be >=
```
benhammondmusic, line 44: this is supposed to be `max_decimals` not `maxd_decimals` right?

benhammondmusic (approved):
good catch with this rounding thing. I think important to show the smallest groups wherever possible. the typos are consistent with the function name but we should probably fix it before merging right?

joshzarrabi:
Copies this function: https://github.com/SatcherInstitute/health-equity-tracker/blob/3b7662a90cf5afb9efdec7eda8b48a73f8e17a3d/frontend/src/data/utils/DatasetCalculator.ts#L58-L75

python/ingestion/dataset_utils.py
joshzarrabi: Yup thanks

python/ingestion/dataset_utils.py
joshzarrabi: thanks

python/ingestion/dataset_utils.py
joshzarrabi: thank you!

PR #1276, opened 2021-11-03
Improve Map Legends on Mobile

### LINKED ISSUE #1274: Mobile “No Data” gets too close to some states
Perhaps on these small screen layouts we can push the entire map down to accommodate the extra vertical space taken by the "no data" image

benhammondmusic:
- makes map fill entire box
- changes aspect ratio for better fit on mobile
- prevents no data legend hitting map on mobile some states

fixes #1274


benhammondmusic:
Before New Mexico was bumping into the "no data" legend on mobile. Slight change gives it space, while also allowing other more horizont
mew mexico not covered anymore
al geos to look ok. This only affects smallest mobile screens

joshzarrabi (approved):

PR #1286, opened 2021-11-04
Update wording on suppressed states

### LINKED ISSUE #1284: clean up phrasing on state list
keep the sentence structure consistent I think it should be:
```
<b>Delaware</b> is included when viewing hospitalizations, but appears as grey when viewing reports on deaths.
<b>Rhode Island</b> is included when viewing deaths, but appears as grey when viewing reports on hospitalizations.
```

Also, I feel like all of these `appears as grey` should just be `appears grey`

benhammondmusic:
closes #1284


joshzarrabi (approved):
Thanks for this

PR #1288, opened 2021-11-04
High / Low units

benhammondmusic:
- adds units to highest.lowest list items in matching tiny font
- uses BOLD for state names as elsewhere on site
- uses `:` instead of `-`
Screen Shot 2021-11-04 at 2 11 46 PM


benhammondmusic:
merged into #1256

PR #1292, opened 2021-11-05
Labels inside bars; mobile responsive formatting for all labels

### LINKED ISSUE #1290: Suggestion: move bar labels inside bars (when possible)
@EbonyRespress suggested a technique of moving the bar labels inside of or overlayed on the largest bars in a visible color and font to help with our issue of wanting more horizontal space to properly show disparities. This should be investigated independently of issues regarding using `k` units / adding explicit `/100k` units to the labels, as horizontal space is at a premium for these charts

benhammondmusic:
closes #1290 

Screen Shot 2021-11-05 at 10 30 05 AM

### DIFF
--- a/frontend/src/charts/DisparityBarChart.tsx
+++ b/frontend/src/charts/DisparityBarChart.tsx
@@ -304,7 +304,6 @@ function getSpec(
         title: pageIsTiny
           ? [`${lightMeasureDisplayName}`, `vs.`, `${darkMeasureDisplayName}`]
           : `${lightMeasureDisplayName} vs. ${darkMeasureDisplayName}`,
-        // titleAnchor: pageIsTiny ? "end" : "null",
         labelFlush: true,
         labelOverlap: true,
         tickCount: { signal: `ceil(width/${BAR_HEIGHT})` },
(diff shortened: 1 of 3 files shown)


benhammondmusic:
got it working (with hard coded cutoff)
black white labels

benhammondmusic:
Now working using JS to calculate the breakpoint (set at 1/3 of the max value from the dataset). There is a likely a way to do it via Vega using `aggregate` but I just did it with JS instead

Screen Shot 2021-11-05 at 10 13 11 PM
Screen Shot 2021-11-05 at 10 13 32 PM

benhammondmusic:
Added same logic to the disparity bar chart... might be worth considering. These labels are longer so compress it horizontally even more, particularly on mobile. This helps with that issue quite a bit.
Screen Shot 2021-11-05 at 10 36 33 PM

benhammondmusic:
A few small tweaks to allow labels to display on smallest mobile screens incl multiline titles and small alignment adjustments (on mobile only)
Screen Shot 2021-11-08 at 2 08 51 PM

benhammondmusic:
- [ ] disable “nice” on mobile to maximize horizontal space

joshzarrabi (approved):
I think this looks great, thanks for working on this.

PR #1297, opened 2021-11-08
better link from pop card

### LINKED ISSUE #1265: Improve Population Card link
Currently, the population card displays a "Learn More" link that simply drops the user onto the FAQ page. This could be improved in several ways:
- Use of generic "Learn More", "Read More" type links is discouraged as it is helpful from a UX and A11y viewpoint to provide the user context for where they will actually be taken
- Dropping the user onto the FAQ page and not even targeting a specific question seems wrong; race categories aren't really addressed on the FAQ in depth. For now I think we should scroll the user down to the "Missing Data" section, and eventually implement  #1263 to have a better flow of information overall
Screen Shot 2021-11-02 at 9 09 13 AM

benhammondmusic:
fixes #1265 

smooth scroll to bottom "Missing People" section, rather than just dropping the user at the top of the faq page (which doesn't address that issue at all)
Screen Shot 2021-11-08 at 3 42 44 PM

### DIFF
--- a/frontend/src/cards/storybook/PopulationCard.stories.tsx
+++ b/frontend/src/cards/storybook/PopulationCard.stories.tsx
@@ -10,6 +10,7 @@ interface PopulationCardProps {
   fips: Fips;
   breakdownVar: BreakdownVar;
   metrics: MetricConfig[];
+  jumpToData: Function;
 }
 export default {
   title: "Cards/PopulationCard",
@@ -18,7 +19,10 @@ export default {
 } as Meta;
 
 const Template: Story<PopulationCardProps> = (args) => (
-  <PopulationCard {...args} />
+  <PopulationCard
+    {...args}
+    jumpToData={() => console.log("Jump to Missing Data")}
+  />
 );
 
 export const Usa = Template.bind({});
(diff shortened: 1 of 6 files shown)


frontend/src/cards/storybook/PopulationCard.stories.tsx
```diff
@@ -18,7 +19,10 @@ export default {
 } as Meta;
 
 const Template: Story<PopulationCardProps> = (args) => (
-  <PopulationCard {...args} />
+  <PopulationCard
+    {...args}
+    jumpToData={() => console.log("Jump to Missing Data")}
```
joshzarrabi, line 24: Do we want this console.log in here?

joshzarrabi (approved):
Works great, thanks for this.

frontend/src/cards/storybook/PopulationCard.stories.tsx
benhammondmusic: I did that since in the storybook i don't think the components relate to one another... so there was no way to show it "jumping down". I honestly haven't used storybook at all and probably dig into  it a bit; i know a lot of people like it.

PR #1305, opened 2021-11-15
Frontend acs merge kff

joshzarrabi:
* Assume acs population pct data is merged on backend for kff vaccine data in `acs_vaccine_population_pct` column

[#1202]


frontend/src/cards/TableCard.tsx
```diff
@@ -39,6 +40,11 @@ export interface TableCardProps {
   variableConfig: VariableConfig;
 }
 
+export const NeverShowProperties = [
```
benhammondmusic, line 43: in general we've been using `SCREAMING_SNAKE_CASE` for constants; not sure if it's worth changing for consistency

benhammondmusic: maybe worth adding a comment here too so we know later why we added this box of never-show properties

joshzarrabi: For sure

frontend/src/cards/TableCard.tsx
```diff
@@ -83,13 +96,28 @@ export function TableCard(props: TableCardProps) {
...
           (row: Row) =>
             row[props.breakdownVar] !== UNKNOWN &&
             row[props.breakdownVar] !== UNKNOWN_RACE &&
             row[props.breakdownVar] !== UNKNOWN_ETHNICITY
         );
 
+        if (showAltPopCompare(props)) {
+          // This should only happen in the vaccine kff state case
+          dataWithoutUnknowns = dataWithoutUnknowns.map((item) => {
+            return {
```
benhammondmusic, line 109: i think we can do this to avoid having to redeclare all of the other properties that already exist on `item`; i'd have to test this out though not positive. 
```
const { vaccine_population_pct, acs_vaccine_population_pct, ...restOfItem } = item
return {  
              vaccine_population_pct: vaccine_population_pct || acs_vaccine_population_pct,
                ...restOfItem
            };
```

benhammondmusic (approved):
sweet. so nice to get all these calculations on the backend async rather than dragging down the client

frontend/src/cards/TableCard.tsx
joshzarrabi: Sorry just saw this, yeah that would be better let me try it out

joshzarrabi: @benhammondmusic That totally works, way more elegant, thanks :)

PR #1306, opened 2021-11-15
Geo dropdown UX tweaks

### LINKED ISSUE #1228: Why is dropdown arrow in geo input?
Right now it obscures the placeholder text and doesn't do anything when you click. 

Two ideas:
- remove it
- make clicking it populate the search results with relevant options. Possibly user's state/county (based on geolocation API) or previously searched locations (based on cookie/local storage) or just United States

image

benhammondmusic:
- fixes #1228 
- adds group names for dropdown options (eg `National`, `States`, `{State} Counties`)
- auto focus user into the search box (better ux for keyboard users, a11y)
- auto-highlight so user can just press enter and get the first result

### DIFF
--- a/config/data_sources/AGG_kff_vaccination.sql
+++ b/config/data_sources/AGG_kff_vaccination.sql
@@ -18,12 +18,12 @@ WITH
         ON a.state_name = b.state_name
   ),
   all_acs as (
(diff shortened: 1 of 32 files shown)


joshzarrabi (approved):
Looks great

Issue #1117, opened 2021-08-27
Search by city on data tracker


benhammondmusic:
![search by city](https://user-images.githubusercontent.com/41567007/131157741-bb7584fd-edc0-40ab-a935-dce3c7a9b72b.png)

Currently the tracker location search box is unable to handle city requests; this is suboptimal as many people might a) attempt a search by city as their first search and b) not even know the name of the county they live in or especially not know the names of their surrounding counties. 

Proposal to implement a lookup and if there is a matching city (or cities) to display those as options with the containing county, which will then map on click to the corresponding county

jgonzalezmsm:
This is a really cool idea and nice workaround to not having cities in the data

jgonzalezmsm:
can search not be an exact match?

benhammondmusic:
mui autocomplete allows different variables for the `value` and the `displayValue`; this could be one implementation, where city names are entries that contain the `displayValue` of the city but the `value` of the parent county (since county level is the lowest level we can show on the tracker)

benhammondmusic:
> add the word "county" after every county name on region select.

Addressed in #1310

benhammondmusic:
> can search not be an exact match?

Filed as new issue #1311

PR #1310, opened 2021-11-17
Add word COUNTY as needed

benhammondmusic:
PROBLEM: 

Particularly until we address #1117 , it's confusing that the tracker doesn't use the word `county`, and they could be confused for the city with the same name.

SOLUTION: 
 
Rather than hard code in the word `county` into the fips and geographies files, I've added some condition logic to the `getDisplayName()` methods that will append the word `county` where it is needed (specifically, non-territory counties that don't already have the word `borough` or `area`)

Screen Shot 2021-11-16 at 8 24 11 PM

### DIFF
--- a/config/data_sources/AGG_kff_vaccination.sql
+++ b/config/data_sources/AGG_kff_vaccination.sql
@@ -18,12 +18,12 @@ WITH
         ON a.state_name = b.state_name
   ),
   all_acs as (
-      SELECT * FROM `acs_population.by_race_state_std`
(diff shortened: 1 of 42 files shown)


joshzarrabi:
This is great, but it looks like you duplicated some of the work thats in here too? #1309

joshzarrabi (approved):

benhammondmusic:
> This is great, but it looks like you duplicated some of the work thats in here too? #1309

ah mustve not started fresh on this one. sorry about that

PR #1327, opened 2021-11-23
replace deprecated methods

benhammondmusic:
Quoting from the MDN docs about MediaQueryList:

MediaQueryList.addListener() Adds a listener to the MediaQueryListener that will run a custom callback function in response to the media query status changing. This is basically an alias for EventTarget.addEventListener(), for backwards compatibility purposes.

addEventListener needs an event type as the first argument, so it becomes this:

// deprecated: MediaQueryList.addListener(listener);
MediaQueryList.addEventListener('change', listener);

The same stands true for removeListener():

// deprecated: MediaQueryList.removeListener(listener);
MediaQueryList.removeEventListener('change', listener);

### DIFF
--- a/frontend/src/utils/usePrefersReducedMotion.tsx
+++ b/frontend/src/utils/usePrefersReducedMotion.tsx
@@ -15,10 +15,8 @@ export function usePrefersReducedMotion() {
     const listener = (event: { matches: any }) => {
       setPrefersReducedMotion(!event.matches);
     };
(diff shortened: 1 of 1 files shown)


benhammondmusic:
https://stackoverflow.com/questions/57371174/replacement-of-deprecated-mediaquerylist-addlistener-event-in-angular-8

joshzarrabi (approved):

PR #1332, opened 2021-11-29
adds img for social share

benhammondmusic:
this img needs to be hosted on prod so that twitter share etc will work properly in dev (and prod)


benhammondmusic:
share

Issue #1267, opened 2021-11-02
NaN% in legend for suppressed states (some?)


benhammondmusic:
Searching for covid deaths by age in some locations including 
- Delaware
- W Virginia
- S Dakota
- Nebraska
- Wyoming

(and likely others) https://healthequitytracker.org/exploredata?mls=1.covid-3.10&dt1=deaths&demo=age results in displaying an empty unknowns map with `NaN%` in the legend

This map shouldn't even display, right? It should be an alert box instead? 
![nan legend](https://user-images.githubusercontent.com/41567007/139902152-93cd8663-0191-4fec-8a52-a9a892c64694.png)

PR #1339, opened 2021-12-04
Suppress blank maps

benhammondmusic:
if there is no data, don't generate the map (rather than displaying an entirely gray map). Better user experience as maps displaying the next geo level down of data, so there may be no data on the map but there really is data in the charts... having a blank map up top may keep the user from scrolling down to see the data filled charts


joshzarrabi (approved):
Looks great, thanks for this

joshzarrabi:
I'm still seeing an empty unknowns map here: https://deploy-preview-1339--health-equity-tracker.netlify.app/exploredata?mls=1.covid-3.46&demo=race_and_ethnicity&dt1=deaths

benhammondmusic:
> I'm still seeing an empty unknowns map here: https://deploy-preview-1339--health-equity-tracker.netlify.app/exploredata?mls=1.covid-3.46&demo=race_and_ethnicity&dt1=deaths

I think this is happening because there's an error in the suppression we do somehow; related to #1267 . Like, there IS data, but not the right data, so it's showing NaN% in the legend and still rendering the map despite data. 

Should we proceed with this PR and fix that other error independently ?

joshzarrabi:
Oh interesting, yeah why don't you make an issue for that and then we can work on that :). Lets merge this though.

PR #1342, opened 2021-12-07
use real vaccine api for vaccine national cdc

joshzarrabi:
Updates to the actual supported cdc vaccine demographic api. 

Unfortunately this doesnt include population numbers like the unofficial api did, so I hardcode the population percentages for age which we cant calculate using the acs numbers.


python/datasources/cdc_vaccination_national.py
```diff
@@ -23,19 +24,37 @@
...
     'Ages_75+_yrs': '75+',
     'Age_unknown': 'Unknown',
+    'US': std_col.TOTAL_VALUE,
+}
+
+
+# The CDC uses age ranges that we can not calculate with the given acs data,
+# and they don't publish these population numbers directly anywhere, so I am
+# taking the population percentages directly off of the chart here:
+# https://covid.cdc.gov/covid-data-tracker/#vaccination-demographic
```
benhammondmusic, line 47: i haven't looked at front end yet, but is this notated for the user that we are using these numbers?

joshzarrabi: On the methodology page: 

>> For the national numbers we use the population numbers provided by the CDC, we chose to do this because they include population estimates from Palau, Micronesia, and the U.S. Marshall Islands, which are difficult to find estimations for. Furthermore the CDC has estimations for age ranges that the ACS numbers do not readily provide, as they use a per year population estimate from the ACS that we do not use anywhere else and have not added to our system.

benhammondmusic (approved):
nice we can get all the info right from the endpoint directly

python/datasources/cdc_vaccination_national.py
benhammondmusic: Cool; should we hyperlink that "population numbers provided by the CDC" to https://covid.cdc.gov/covid-data-tracker/#vaccination-demographic ?

PR #1343, opened 2021-12-07
frontend changes for national vaccination api update

joshzarrabi:
* we just get all numbers directly from the api


benhammondmusic (approved):
👍

PR #1358, opened 2021-12-16
Centralizes External URLs; Adds Status Tests

### LINKED ISSUE #1319: Implement automated testing for all external URLs
- move all external urls to lookup table created in #1333 
- use `Promise.all` or similar to send async requests to all external urls, report if any are broken
- figure out where makes the most sense for this test, as a pre-commit or possible just with a github action pushing

benhammondmusic:
closes #1319 
- [x] collect all resource links into a single exported typed object
- [x] collect other all external urls into a single exported, typed object
- [x] test that all links are using HTTPS
- [x] test that pings each URL and checks for bad HTTP response


joshzarrabi (approved):
Nice, this is smart, thanks Ben

PR #1359, opened 2021-12-18
Preserves determinant/geo when switching carousel mode

### LINKED ISSUE #1197: Tracker shouldn’t reset to covid/us when navigating around site

benhammondmusic:
Fixes #1197 

### PROBLEM: 

Currently, every change in carousel setting resets the madlibs to the default settings. So when the user selects a condition and geography on the "Investigate" setting of the carousel, when they click on a new setting (eg either of the compare reports) their previously selected settings are abandoned and it defaults to covid.diabetes and USA/Georgia. 

### PROPOSED:

Improve the UX by retaining the decisions the user has already made (eg selecting particular condition(s) and geographies). Currently, this will only preserve the CURRENT options that can fit in the NEW options, meaning if the user is on a 1 CONDITION 1 GEO report and goes to a 2 CONDITION 1 GEO report, the 2nd condition will have to be a default (covid, or covid vaccinations if covid is the first selection)

In the future, it might be nice to use localstorage to remember the last setting the user has made for all 4 potential slots `Condition1 Condition2 Geo1 Geo2` but that adds complexity and should be handled independently

### DIFF
--- a/frontend/src/pages/ExploreData/ExploreDataPage.tsx
+++ b/frontend/src/pages/ExploreData/ExploreDataPage.tsx
@@ -151,42 +151,40 @@ function ExploreDataPage() {
   const handleCarouselChange = (index: number) => {
     // Extract values from the CURRENT madlib
     const var1 = madLib.activeSelections[1];
-    let currentVar2, currentGeo1, currentGeo2;
+    let geo1;
 
     switch (madLib.id) {
       case "disparity":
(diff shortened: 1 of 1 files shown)


benhammondmusic:
https://user-images.githubusercontent.com/41567007/146625264-f54aa334-676e-4984-aa6e-580228c2bbc5.mov

joshzarrabi (approved):

PR #1363, opened 2021-12-21
Improves Loading of Page Content

benhammondmusic:
PROBLEM:
Since #1361 the loading is better, but there is still a problem where the footer flashes in the middle of the screen before the page content (everything in between the constant header and footer) fully loads. This is ugly, confusing, and affects the page's cumulative layout shift (CLS) negatively which has SEO implications

SOLUTION:
Create a fallback "page" which is simply a fully white background with the same circular spinner we are using everywhere else for loading data. On fast connections this will likely only be visible on the `Explore Data` page, but on slow connections it makes for a significantly better UX

https://user-images.githubusercontent.com/41567007/146874597-1a570f2f-7a1b-4c68-b61e-29fe3a153933.mov

### Video showing a very slow connection, with an initial spinner as the whole page loads, and then the normal secondary spinners as the maps/charts appear.


joshzarrabi:
Looks great

joshzarrabi (approved):

PR #1367, opened 2021-12-22
Standardizes Font Sizes; Removes Dead CSS Classes

benhammondmusic:
- [x] eliminates dead css class CALLS and DECLARATIONS 
- [x] eliminates stray font sizes
- [x] uses `rem` for ALL font sizing in variable declarations (best practice for a11y)
- [x] uses only sass variables for all font sizes
- [x] complete manual checking across all device widths

Also removes a lot of unused CSS


joshzarrabi:
@benhammondmusic
Screen Shot 2021-12-27 at 1 17 52 PM I noticed this change made the top bar text bigger. I think we
Screen Shot 2021-12-27 at 1 17 20 PM
 should fix that as it takes up even more screen real estate.

benhammondmusic:
> @benhammondmusic Screen Shot 2021-12-27 at 1 17 52 PM I noticed this change made the top bar text bigger. I think we Screen Shot 2021-12-27 at 1 17 20 PM should fix that as it takes up even more screen real estate.
 
Good call; I actually used a nice CSS feature `clamp()` for the madlib font size, which allows a variable font size within a minimum and maximum boundary. I'll lower the maximum

benhammondmusic:
Updating font sizing logic for the MADLIB. Here are shots at 500px (small mobile) 1200px (small desktop) 2400 (large desktop)

joshzarrabi:
@benhammondmusic the text is still bigger for me but the top bar takes up the same amount of space so 🤷

joshzarrabi (approved):
This is great, thanks you!

PR #1368, opened 2021-12-22
Removes "Copy to Clipboard" UI and package dependency

benhammondmusic:
Previously "copy to clipboard" was already being used in the "Share" modal, so it was a fun touch to add it to the citation. I never loved the layout, and honestly it's not super needed. I have removed it to reduce clutter on the screen, and more importantly remove a package (and its subsequent dependencies) from our project (since we no longer have the Share Modal or clipboard function there either)

Screen Shot 2021-12-22 at 12 04 38 AM


joshzarrabi:
@benhammondmusic I kind of like it in the box, even without the copy button, but do whatever you think is better.

joshzarrabi (approved):

benhammondmusic:
> @benhammondmusic I kind of like it in the box, even without the copy button, but do whatever you think is better.

Ya, I think with the clipboard icon removed and with the missing `max-width` reinstated (in #1367 ) it'll look nice in the raised card. I also think it'd be very easy to make clicking the raised card copy to clipboard automatically... without needing an npm package and extra UI.

PR #1370, opened 2021-12-22
Fixes NaN% on Unknown Maps and Improves Missing Data Alert 

### LINKED ISSUE #1267: NaN% in legend for suppressed states (some?)

benhammondmusic:
### improves `<MissingDataAlert>` on all MAP cards:
- by specifying correct level-down region thats missing and required to draw maps
- by cleaning up the conditional rendering and making it easier to understand the code


joshzarrabi (approved):
Nice

PR #1371, opened 2021-12-23
Adds Determinant Definitions

benhammondmusic:
- [x] adds fields to metric query to store a condition's DEFINITION, FRIENDLY SOURCE NAME and SOURCE URL
- [x] renders ALL condition definitions (if they exist) on the METHODOLOGY page
- [x] renders the SELECTED condition(s) on the EXPLORE DATA page
- [x] creates a "scroll down" link for the first occurrence of the defined term in the 100k map card


benhammondmusic:
https://user-images.githubusercontent.com/41567007/147190343-94f41f67-2073-4048-8d9f-e35d1c698618.mov

joshzarrabi:
Thanks for this Ben, I think it is a big UI improvement that people will like alot.

Did we you clear these definitions with Ebony or someone else from SHLI? That would be my main concern.

joshzarrabi:
Also, I think we should add definitions for every condition, and link to the definition like we do with vaccinations. Does that sound reasonable?

benhammondmusic:
> Thanks for this Ben, I think it is a big UI improvement that people will like alot.
> 
> Did we you clear these definitions with Ebony or someone else from SHLI? That would be my main concern.

Thanks! For the UHC variables the definitions come directly from the Americas Health Ranking site (that's why I thought we should include the source link right after the definition so people could easily click through). 

I'll work on finding the official source definitions for our other conditions, and making them a "scroll to bottom" button like the vaccine is in the 100k map. We've been talking about adding a "condition card" which would sort of build off from this feature and dynamically display information about the selected condition(s); we'll need to chat as a team about where that will live and how it works with both our existing data and also the new PDOH data (where she has the extra explainer text she'd like featured)

benhammondmusic:
All terms in map card are hyperlinked to the definition on the bottom of the page
https://user-images.githubusercontent.com/41567007/148263570-a5a92594-48a5-42ac-889b-2bb666d12d32.mov

joshzarrabi (approved):

PR #1373, opened 2021-12-28
Strengthens Typing; allows correct groups in MultiMap and Group Dropdown

### LINKED ISSUE #1350: Fix suppressed blank maps on small multiples view
Currently the code prevents blank maps from rendering; generally this is what we want but on the small multiples view the blank demographic groups are still taking up their assigned space, leaving big gaps.

Screen Shot 2021-12-09 at 1 40 01 PM

benhammondmusic:
- [x] confirm ALL demographic groups are included (age groups for vaccines, etc, all race groups for vaccine at all levels)

Fixes #1350

This PR extracts most of its logic from #1355 to allow filtering of demographic groups that don't contain data at that level

- properly filters available demographic group options into `withData` and `noData`
- sends `noData` into multimap to be displayed in an alert box; better visually than a blank map box, but still highlighting the issues re: missing data and inconsistent terminology / data collection
- updates `.getFieldValues()` and associated tests
- adds stronger typing for breakdown variables / groups


benhammondmusic:
Screen Shot 2021-12-28 at 11 23 18 AM

joshzarrabi (changes requested):
This warning should say "state level". Those demographics do exist on the national level, but no state reports them so we cant do breakdowns.

Screen Shot 2021-12-28 at 2 06 52 PM

benhammondmusic:
Screen Shot 2021-12-29 at 10 12 31 AM
Screen Shot 2021-12-29 at 10 12 56 AM

frontend/src/data/utils/Constants.ts
```diff
@@ -22,5 +22,83 @@ export const UNKNOWN_RACE = "Unknown race";
 export const UNKNOWN_HL = "Unknown Hispanic or Latino";
 export const UNKNOWN_ETHNICITY = "Unknown ethnicity";
 
+export const SEX_GROUPS = [MALE, FEMALE, UNKNOWN, ALL] as const;
```
joshzarrabi, line 25: We haven't had to define demographic groups on the frontened before and I'm curious what the gain is to adding that extra barrier?

benhammondmusic: It's basically just creating the typescript restricted options to prevent errors in coding going forward; similar to how we define all of the `MetricId`s:
```
export type MetricId =
  | "acs_vaccine_population_pct"
  | "brfss_population_pct"
...
  ```
  
We still rely on the actual property name coming from the backend data for frontend presentation, but just confirm that the property is in our expected range of properties. Typescript gets complied away so it's mostly just preventing us from trying to assign the wrong string to these variables. Rather than define them as string unions like they do for `MetricId`, I'm defining the arrays first so we can use them for iterations / logic checks, and then generating the types from the arrays. 

I think they just didn't get a chance to strongly type everything... the code is littered with "TODO" comments to strongly type everything just to prevent headaches when trying to decipher some of the variable names. 

We can definitely leave it as strings but it seems to defeat the purpose of typescript. Is there a concern with strongly typing them?

File: frontend/src/data/utils/Constants.ts [hunk 2]
```diff
@@ -1,26 +1,131 @@
...
 export const HISPANIC = "Hispanic or Latino";
 export const NON_HISPANIC = "Not Hispanic or Latino";
 export const TWO_OR_MORE = "Two or more races";
+export const UNKNOWN_RACE = "Unknown race";
+export const UNKNOWN_HL = "Unknown Hispanic or Latino";
+export const UNKNOWN_ETHNICITY = "Unknown ethnicity";
 
+// GROUPED RACE DEMOGRAPHICS
+export const STANDARD_RACES = [
+  "American Indian and Alaska Native (Non-Hispanic)",
```
joshzarrabi, line 22: It feels a little bit weird that only some of these are constants. Maybe we should take the time to come up with a better way of defining all of the demographic groups on the frontend and stick to it. I'm fine merging this for now so we dont have to keep thinking about merge conflicts but this would be a good thing to think about I think

benhammondmusic: Those old individually declared ones were there already for tests; I guess we should take those out and then just type check our tests? 

```
export const WHITE_NH = "White (Non-Hispanic)";
export const WHITE = "White";
export const ASIAN_NH = "Asian (Non-Hispanic)";
export const HISPANIC = "Hispanic or Latino";
export const NON_HISPANIC = "Not Hispanic or Latino";
```

benhammondmusic: I just fixed it up a little; basically declare all race options as consts and then combine then into arrays; this allows us to use the single consts in tests / checks throughout the app but also keeps them organized into arrays for the filtering we do in the menu, etc. I didn't bother to export all of the age groups since we don't use any individual age group options in the code. The one that had been declared for a test i moved into the test file and used the new `AgeBucket` type

joshzarrabi (approved):
This is excellent, thank you so much for putting so much detail into this

Issue #1383, opened 2021-12-29
Idea: hover box showing outline of tracker cards and user position on the page


benhammondmusic:
We have discussed a feature that has a hover box on the side of the tracker to outline all cards available, and showing the user where they are and what other cards (off screen) are available. This would be the ideal behavioral, and this would track which specific card was "in view" and re-scroll the user to the equivalent scroll position with new tracker position.

benhammondmusic:
I think we could use the MUI `Stepper` for this and stay in line with the MUI components we are already using. Look at the right hand sidebar of this video:


https://user-images.githubusercontent.com/41567007/147702679-5d4b9547-5c54-4bbd-b114-da4c3a6969db.mov

benhammondmusic:
https://css-tricks.com/sticky-table-of-contents-with-scrolling-active-states/

https://css-tricks.com/sticky-smooth-active-nav/

jgonzalezmsm:
#1733

PR #1384, opened 2021-12-29
Smoothly scrolls user to top of page on changes to tracker settings

### LINKED ISSUE #1195: Screen jumps to bottom when select new location
When scrolled down on Explore Data (say viewing the disparity bar chart), if you enter a new location into the madlib the screen jumps down to the "What Data Is Missing" rather than either the top of the page or staying on the same chart

https://user-images.githubusercontent.com/41567007/136134098-64235499-545a-4a84-b098-f9197d020b8c.mov

benhammondmusic:
fixes #1195

## Issue: 
When the user adjusts the madlib settings (condition, geo, or carousel setting) the page appears to jump around at times. This is because the browser is attempting to maintain position, but some cards are expanding / contracting and that causes the "same" scroll position to look different (and often random). 

## Solution: 
Whenever the user selects a new setting (condition, geo, or carousel setting), smoothly scroll them back to the top of the page. This reinforces to the user that they are seeing an entirely new report as well as prevents the "random" jumping of scroll position that happens when some charts/maps/alerts appear and disappear

## Future Idea: #1383
We have discussed a feature that has a hover box on the side of the tracker to outline all cards available, and showing the user where they are and what other cards (off screen) are available. This would be the ideal behavioral, and this would track which specific card was "in view" and re-scroll the user to the equivalent scroll position with new tracker position. 

 
[
smooth scroll to top if user adjusts madlib vars, geos or tracker mode](https://user-images.githubusercontent.com/41567007/147702288-81537d85-dec1-41f4-99cc-81387c5fa4d2.mp4
)

### DIFF
--- a/frontend/src/pages/ExploreData/ExploreDataPage.tsx
+++ b/frontend/src/pages/ExploreData/ExploreDataPage.tsx
@@ -92,6 +92,11 @@ function ExploreDataPage() {
     };
   }, []);
 
+  /* scroll to top on any changes to the madlib settings */
+  useEffect(() => {
+    window.scrollTo({ top: 0, behavior: "smooth" });
+  }, [madLib]);
+
   const setMadLibWithParam = (ml: MadLib) => {
     setParameter(MADLIB_SELECTIONS_PARAM, stringifyMls(ml.activeSelections));
     setMadLib(ml);


joshzarrabi:
Looks great, thanks ben

joshzarrabi (approved):

PR #1397, opened 2022-01-14
Adds verbose units on table chart

### LINKED ISSUE #1396: Consider identifying % unit in the table card
# Background

Right now the table card shows two percentages

* The percent of people with the shown condition out of all the people with the condition who are a certain demographic 
* The percent of the population that demographic group makes up

See below

Screen Shot 2022-01-14 at 4 19 57 PM

We have heard from user interviews (with the kff) that this confuses people, and we have seen people come up with the wrong conclusions from this chart (mixing up % of population and % of covid cases, for example)

# Proposed Solution

We should specify the meaning of the `%` next to every metric on the table, to make it less likely that people will misinterpret. 

For example

| Share of Total COVID-19 Cases      | Population Share |
| ----------- | ----------- |
| 8.7% of affected      |  6.1% of population       |
| 0.2% of affected   | 0.8% of population        |

Considered using `of affected` or similar if some condition names would be too long, however it appears using the shortened metric config `shortVegaLabel` looks ok across devices

benhammondmusic:
fixes #1396


benhammondmusic:
Screen Shot 2022-01-14 at 2 36 51 PM

benhammondmusic:
`per 100k` is force onto it's own line on mobile to prevent weird alignment issues

Screen Shot 2022-01-14 at 3 21 54 PM
Screen Shot 2022-01-14 at 3 22 03 PM
Screen Shot 2022-01-14 at 3 22 11 PM
Screen Shot 2022-01-14 at 3 22 21 PM

joshzarrabi (approved):

PR #1422, opened 2022-01-28
Adds an #anchorId to each card type; sets us up for later internal nav features

### LINKED ISSUE #1420: Add #anchorLinks to each card on the Explore Data reports
This will be helpful for both #1383 and also for potentially internally linking from a blog post to display a particular chart rather than the entire report

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1422--health-equity-tracker.netlify.app/" target="_blank">
  <br />
Click to demo updates </a> 

## Description

adds `id="mapCard"` or equivalent to each card type, allowing for later internal navigation features like #1383 

## Motivation and Context

closes #1420 

## How has this been tested?
Manually Chrome MacOS for single and compare variable modes


joshzarrabi:
@benhammondmusic looks good but two issues:

1. The page scrolls too far down when going to an anchor, I think its because we are ignoring the top bar. See below

Screen Shot 2022-02-04 at 10 38 12 AM

2. I know we have the plan to add internal navigation, but maybe it would be good to somehow provide a way to easily generate a link to each card? Not sure how exactly I would want that to look though.

benhammondmusic:
> 1. The page scrolls too far down when going to an anchor

Good call; there are several ways to account for this but I'd rather merge as is and then solve the offset once we implement the actual navigation. The PR was just sort of setting up the landmarks for later 

> 2. I know we have the plan to add internal navigation, but maybe it would be good to somehow provide a way to easily generate a link to each card? Not sure how exactly I would want that to look though.

I think this is a good idea; simplest way to start would be to intercept the `#mapId` and then smooth scroll the user down. As i mentioned the blocking that the data loading causes sometimes interferes with this. We also considered hiding all of the other cards offering a "show full report" button, similar to what we do on the data sources page

For both items, I'd still like to merge as is just with the id labels in place, and then account for the issues you mentioned in the implementation

PR #1448, opened 2022-02-10
remove all instances of the pandas dataframe append function

joshzarrabi:
* it is depreciated and will be removed in a future version

## Preview Link
No frontend changes

## Description

Removes the pandas `.append` function and replaces it with the `.concat` function. 

## Motivation and Context

Pandas has depreciated `.append` because it is not very performant and throws a very annoying error every time we use it.

## How has this been tested?

Ran our unit tests which should be good.

## Types of changes


benhammondmusic (approved):
🐼

Issue #1471, opened 2022-02-18
Consider changing how we calculate national population numbers


joshzarrabi:
**Describe the bug**
Currently, when calculating the per 100k measures for covid cases, deaths, and hospitalizations at the national level, the denominator, which is the population used to calculate the per 100k number, includes the population from every state that we include *cases* from. 

The code calculating the population is here: https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/CdcCovidProvider.ts#L101-L114

This is unintuitive, because we restrict data from additional states that don't report enough hospitalizations or deaths to meet our threshold against the nytimes numbers. So, for example. even though we show `no data` for Hawaii COVID deaths, the population numbers from Hawaii are still included in the national per 100k population numbers for COVID deaths. 

We updates the methodology page here https://github.com/SatcherInstitute/health-equity-tracker/pull/1472 to specifically call this out for now. 

**Expected behavior**
If a state is grey on the national map, its population should not be included in the national population for that metric. 

**Additional context**

We discussed this with Peter and Feng and they agreed it should be changed.

**Implementation notes**
This should probably wait until we move the calculation to the backend.

PR #1472, opened 2022-02-18
clarify that we currently only suppress pop from states that dont report cases

joshzarrabi:
- we should consider changing this: see #1471

## Preview Link

<a href="https://deploy-preview-1472--health-equity-tracker.netlify.app/" target="_blank">
  <br />

## Description

Add a more clear description of how we calculate national population numbers

## Motivation and Context

The current explanation is confusing and doesn't exactly describe how we calculate national population numbers

## Types of changes


benhammondmusic (approved):
good clarification. I would still like to do #1263 and keep the tracker for data, and the methodology page(s) for context and clarification but that's a future thing.

PR #1503, opened 2022-03-03
Adds download script

joshzarrabi:
## Description

Adds a script to download data files from test gcp projects in the same format that the dataserver will serve them to frontend in a normal flow.

Run `./scripts/download_data_file -f <data-set-name>` to get the file written to the `frontend/public/tmp` directory.

## Motivation and Context

It is annoying to have to download the files by hand via big query, and when we do it writes the file differently than the data server would.

Right now when downloading directly from big query items with the value `null` in the dataset are added to the json while our frontend expects those not to be there. This fixes that.

## Types of changes

- Chore (A developer productivity fix that doesn't effect the user)


benhammondmusic (approved):
This will be way easier thanks!

PR #1508, opened 2022-03-07
adds chore to pr template

joshzarrabi:
- a chore is a developer productivity fix that adds no direct user value


benhammondmusic (approved):
🧺

benhammondmusic (changes requested):
Should be "affect" not "effect"

benhammondmusic (approved):

Issue #1475, opened 2022-02-21
Generate national ACS population numbers on the backend


joshzarrabi:
## Background

This is the first step in service of https://github.com/SatcherInstitute/health-equity-tracker/issues/1474, and it needs to be done first because moving the national calculation to the backend for every condition will rely on it.

Currently, whenever we want to generate national population numbers, we download the state level population dataset and then do a [national aggregation using the dataforge library in typescript.](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/AcsPopulationProvider.ts#L66-L74)

## Proposal

Because we need to generate national numbers on the fly depending on what states data we include in national aggregates, we should include a function in the ACSPopulation class that generates a national dataset depending on the given states we want to include. 

We also need a static national dataset that lives in Big Query, in order  for the frontend to pull down when we need to display national population numbers.

## End State

- [x] There is a function in https://github.com/SatcherInstitute/health-equity-tracker/blob/main/python/datasources/acs_population.py with the interface 
`def generate_national_dataset(state_population_numbers, states_to_include) -> df: national_pop_df`
- [x]  acs_population.by_sex_national
- [x]  acs_population.by_race_national
- [x]  acs_population.by_age_national
- [x] [This code](https://github.com/SatcherInstitute/health-equity-tracker/blob/5d6f4948cce9407f69a3ef657e4a163609a8d433/frontend/src/data/variables/AcsPopulationProvider.ts#L8-L16) and [this code](https://github.com/SatcherInstitute/health-equity-tracker/blob/5d6f4948cce9407f69a3ef657e4a163609a8d433/frontend/src/data/variables/AcsPopulationProvider.ts#L66-L74) are deleted from the frontend.

joshzarrabi:
Finished with this commit: https://github.com/SatcherInstitute/health-equity-tracker/commit/76010fbc58f3e60dcfc3ee14c442dd2143198c4b

Issue #1436, opened 2022-02-04
Pandas not testing for correct column names


benhammondmusic:
``` python
assert_frame_equal(mock_bq.call_args_list[i].args[0], expected_df, check_like=True)
```
seems to only test for deep equality of matching columns, however the tests pass if the expected df has different column names than the mocked df. This can accidentally allow malformed data to pass the test phase

PR #1511, opened 2022-03-09
use normal pattern of importing std_col as a package in acs_population.py

joshzarrabi:
- instead of imprting everything individually

## Preview Link

<a href="https://deploy-preview-1234--health-equity-tracker.netlify.app/" target="_blank">
  <br />

## Motivation and Context

see https://github.com/SatcherInstitute/health-equity-tracker/issues/1475

## Types of changes

### DIFF
--- a/python/datasources/acs_population.py
+++ b/python/datasources/acs_population.py
@@ -1,13 +1,8 @@
 import pandas as pd
 
-from ingestion.standardized_columns import (HISPANIC_COL, RACE_COL,
-                                            STATE_FIPS_COL, COUNTY_FIPS_COL,
-                                            STATE_NAME_COL, COUNTY_NAME_COL,
-                                            POPULATION_COL, AGE_COL, SEX_COL,
-                                            Race, RACE_CATEGORY_ID_COL,
-                                            RACE_INCLUDES_HISPANIC_COL,
-                                            TOTAL_VALUE, POPULATION_PCT_COL,
-                                            add_race_columns_from_category_id)
+import ingestion.standardized_columns as std_col
+
+from ingestion.standardized_columns import Race
 from ingestion import url_file_to_gcs, gcs_to_bq_util, census
 from datasources.data_source import DataSource
 from ingestion.census import (get_census_params, parse_acs_metadata,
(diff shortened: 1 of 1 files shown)


benhammondmusic (commented):
I like the idea of matching the std_col pattern across files, but wouldn't it be more readable to have all files import the specific properties rather than the whole class and avoid the extra std_col each time? I guess it's a preference thing but I'd vote for shorter variable references

python/datasources/acs_population.py
```diff
@@ -291,15 +287,22 @@ def write_to_bq(self, dataset, gcs_bucket):
         frames['by_sex_%s' % self.get_geo_name()] = self.get_by_sex(
             frames[self.get_table_name_by_sex_age_race()])
 
+        # Generate national level datasets based on the state datasets
+        if not self.county_level:
+            for demo in ['age', 'race', 'sex']:
+                state_table_name = 'by_race_state_std' if demo == 'race' else 'by_%s_state' % demo
```
benhammondmusic, line 293: would it make sense for us to coordinate `race` vs `race_ethnicity` throughout the backend? maybe just use race? It just seems to make for more complicated code that sometimes we name things w/ the word ethnicity attached and sometimes we don't

python/datasources/acs_population.py
```diff
@@ -574,3 +577,44 @@ def _create_ingesters(self):
             ACSPopulationIngester(False, BASE_ACS_URL),
             ACSPopulationIngester(True, BASE_ACS_URL)
         ]
+
+
+def generate_national_dataset_with_all_states(state_df, demographic_breakdown_category):
+    all_state_fips = set(state_df[std_col.STATE_FIPS_COL].drop_duplicates().to_list())
```
benhammondmusic, line 583: do you need to both cast it to a set and use drop_duplicates() ? isn't that doing the same thing?

python/tests/datasources/test_acs_population.py
```diff
@@ -56,6 +60,34 @@ def get_sex_by_age_county_value_as_df(concept):
...
+    assert_frame_equal(national_df, expected_df, check_like=True)
+
+
+def testGenerateNationalDatasetSex():
+    state_df = pd.read_csv(os.path.join(TEST_DIR, 'national', 'state_by_sex.csv'), dtype={'state_fips': str})
+    expected_df = pd.read_csv(os.path.join(TEST_DIR, 'national', 'national_by_sex.csv'), dtype={'state_fips': str})
+    states_to_include = {'01', '06'}
+
+    national_df = generate_national_dataset(state_df, states_to_include, 'sex')
+    assert_frame_equal(national_df, expected_df, check_like=True)
```
benhammondmusic, line 79: in some of the other files we addressed #1436 by using
```
assert set(national_df.columns) == set(expected_df.columns)
```

not sure it that would be helpful here?

benhammondmusic (approved):
looks great

python/datasources/acs_population.py
joshzarrabi: Good catch, I don't need to drop_duplicates() because turning the list into a set will do that. 

But it *is* important to use a set because the lookup time is O(1), which is what we are using it for.

python/datasources/acs_population.py
joshzarrabi: Yeah its definitely an issue throughout the codebase. For now it probably doesnt matter that much but if/when the cdc actually does split out race/ethnicity its something we will need to tackle.

PR #1514, opened 2022-03-10
Use Unique VariableIds

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1514--health-equity-tracker.netlify.app/" target="_blank">
  <br />
Click to demo updates </a>  

## Description
Currently several of our data types use the same `variableId` value of `cases`; this makes it more difficult to filter or otherwise detect the correct data type when needed. This PR re-assigns more accurate variable ids to all data types

Edit: also more work to distinguish between `DropdownVarId` which are used for the parent level, selectable health topics, and the `VariableId` which is the data type. In some cases they are the same (like `depression`) and other cases they are nested (`covid` contains `covid_deaths` `covid_cases` etc)

## How has this been tested?
manually and also frontend tests are passing

## Types of changes


joshzarrabi (approved):

benhammondmusic:
@joshzarrabi made some more changes; realized we needed more definition between the parent level `DropdownVarId` and the data type level `VariableId`. Maybe we should change `VariableId` to `DataTypeId`; might make more sense

joshzarrabi:
> @joshzarrabi made some more changes; realized we needed more definition between the parent level `DropdownVarId` and the data type level `VariableId`. Maybe we should change `VariableId` to `DataTypeId`; might make more sense

Yeah I was always confused by that. I think changing the name could be a good call.

joshzarrabi (approved):

PR #1521, opened 2022-03-17
Load ACS2010 from /data 

benhammondmusic:
- rather than needing to manually upload acs2010 files, this adds new functions that allow reading directly from the `/data` directory which are then pushed to BigQuery
- brings all functions in line to use `df` instead of some being `dataframe`
- use convention of `pd` for pandas

### DIFF
--- a/python/ingestion/gcs_to_bq_util.py
+++ b/python/ingestion/gcs_to_bq_util.py
@@ -1,9 +1,9 @@
 from datetime import datetime
 from datetime import timezone
+import pathlib
 import requests
 import json
 import os
-from pathlib import Path
 import pandas as pd
 from google.cloud import bigquery, storage
 
@@ -240,8 +240,7 @@ def load_csv_as_df_from_data_dir(directory, filename, dtype=None):
     directory: directory within data to load from
     filename: file to load the csv file from"""
 
-    home = str(Path.home())
-    file_path = os.path.join(home, "data", directory, filename)
+    file_path = os.path.join("data", directory, filename)
 
     return pd.read_csv(file_path, dtype=dtype)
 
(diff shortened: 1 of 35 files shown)


python/ingestion/gcs_to_bq_util.py
```diff
@@ -221,15 +221,49 @@ def load_json_as_dataframe(gcs_bucket, filename, dtype=None):
...
+
+    #  go up a level
+    path_parent = os.path.dirname(os.getcwd())
+    os.chdir(path_parent)
+
+    # go into /data
+    os.chdir("data")
+
+    # combine sub-dir and filename
+    file_path = os.path.join(directory, filename)
```
joshzarrabi, line 264: Why change directories, why not just generate the entire filepath?

benhammondmusic: I couldn't get that way working.... and this did...  and I was tired of debugging it by running the entire process lol. If you have a quick fix I'm happy to use it

joshzarrabi: I'll give it a shot

joshzarrabi (approved):

python/ingestion/gcs_to_bq_util.py
```diff
@@ -7,6 +7,9 @@
 from google.cloud import bigquery, storage
 
 
+DATA_DIR = os.path.join(os.sep, 'app', 'data')
```
benhammondmusic, line 10: will it always be `app` regardless of where this is running (docker locally, gcp, etc) ? That's what I wasn't sure

joshzarrabi: This code will only run in the docker environment

PR #1526, opened 2022-03-22
checks in state fips file for testing

joshzarrabi:
## Description

Adds a checked in state fips file so we can mock out the call to big query to merge in state fips codes.

To use it add the following decoration ontop of every test function that merges in state fips codes using [the merge_fips_codes function](https://github.com/SatcherInstitute/health-equity-tracker/blob/3ce95d2637c56bda34c1127fd29830dbddaf08dd/python/ingestion/dataset_utils.py#L111-L126)
```
@mock.patch('ingestion.gcs_to_bq_util.load_public_dataset_from_bigquery_as_df',
            return_value=test_utils.get_state_fips_codes_as_df())
```

## Types of changes

### DIFF
--- a/python/tests/data/utils/state_fips.json
+++ b/python/tests/data/utils/state_fips.json
@@ -1,0 +1,344 @@
+[
+  {
+    "state_fips_code": "01",
+    "state_postal_abbreviation": "AL",
+    "state_name": "Alabama",
+    "state_gnisid": "01779775"
+  },
+  {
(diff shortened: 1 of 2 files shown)


benhammondmusic:
Is there a place we should store instructions like the comment above about how to use it? Maybe in the readme for this repo? I'm just worried these kind of things get lost quickly

benhammondmusic (approved):
🗺

Issue #966, opened 2021-06-03
Age adjusting COVID data


vanshkumar:
We should likely do this, question is how much work is it & how we should prioritize

vanshkumar:
To evaluate how important this is, we can compare the below table from the CDC (which age-adjusts) to what the equivalent numbers in the tracker say

![Screen Shot 2021-06-03 at 2 05 56 PM](https://user-images.githubusercontent.com/3394642/120711927-d1cc2700-c474-11eb-9439-2f8f6f23fe99.png)

Source: https://www.cdc.gov/coronavirus/2019-ncov/covid-data/investigations-discovery/hospitalization-death-by-race-ethnicity.html

vanshkumar:
What the current rate ratios in the tracker (i.e. non age-adjusted) look like:

![Screen Shot 2021-06-09 at 11 33 59 AM](https://user-images.githubusercontent.com/3394642/121409648-96b77100-c916-11eb-8350-01a242a9accf.png)

jgonzalezmsm:
Proposal is to add age adjustment data as a demographic toggle for race and ethnicity. Would also need to add alert defining age adjustment need & formula(?). JC to present to Ebony for timing and next steps.

benhammondmusic:
Make sure we update all website copy re: our numbers not being age-adjusted. 

![age adjust](https://user-images.githubusercontent.com/41567007/129941963-88766829-c642-4c9c-9a20-f58dd229125e.png)

Also, does it make more sense to only display age-adjusted numbers in relevant situations? Or do we want the user to be able to toggle between adjusted/non-adjusted?

vanshkumar:
FYI here is the design doc for age adjustment that I wrote before the end of my fellowship: https://docs.google.com/document/d/16ULlfOQ1bIrQz4PD4yu4MPuxHRCgC4W3asl9rBXCHMc/edit?usp=sharing&resourcekey=0-O0cib8eWUbbLZ_QV90VIPw

benhammondmusic:
This is excellent @vanshkumar , thank you for that.

joshzarrabi:
Some updates on this.

* PR with ongoing work: https://github.com/SatcherInstitute/health-equity-tracker/pull/1179

[...]

PR #1531, opened 2022-03-24
switch to using TOTAL as the standard population for age adjustment 

joshzarrabi:
## Description

Uses the Total population breakdown of each state as the standard population for age adjustment.

## Motivation and Context

For #966, and in response to expert feedback we had on using `White (Non-Hispanic)` as the standard population.

## How has this been tested?

Yes, testes were updated. 

## Types of changes


benhammondmusic (approved):
awesome

Issue #1474, opened 2022-02-21
Backend Calculation Mega Issue


joshzarrabi:
## Background

Currently we do lots of calculations shown to the user on the frontend. In most cases, the user downloads condition and population metrics separately, and then their browser aggregates the numbers on the frontend in typescript.

The reason we do this is because of the decision to use the Google Big Query SQL function as our method to combine datasets. 

Our current process for data analysis is the following 
1. Download data from an external data source, and use python and pandas to coerce the numbers to be ready to be merged with population numbers.  Then upload those numbers to a big query table. This is represented by the [ingest to bq](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/airflow/dags/cdc_vaccination_county.py#L20-L23) operator on airflow.

2. Merge that dataset with the population dataset, and any other external big query datasets needed. This is represented by the [aggregator step](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/airflow/dags/cdc_vaccination_county.py#L25-L27) on airflow.

3. Download the merged dataset, calculate the per100k, and any other metrics that require population numbers.

4. If its a national dataset, also download the standalone population dataset, and aggregate both the condition and population dataset to national numbers. Then merge and calculate all metrics.

## Problems with the current approach

1. This is *slow*. 

The [dataforge library](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/CdcCovidProvider.ts#L1) we use as a pandas stand-in on the frontend is horribly inefficient. Earlier we found that merging the ACS population percentage to the covid data [causes some browsers to crash](https://github.com/SatcherInstitute/health-equity-tracker/issues/1202).  While we fixed that, we still see many operations using this library take a number of seconds on the frontend.

2. It makes it basically impossible to test our numbers end to end. 

Because we break our data analysis into three steps which need to be tested individually, and because the SQL aggregation steps CAN NOT be tested, we can not actually end to end test the per 100k numbers the user ends up actually seeing on the page. This is self evidently bad and makes adding any new metric to the page take longer and be scarier.

## Proposed Solution

We should go through conditions, one by one, and move every single calculation to the backend. 

## End State
This track of work will be complete when there is no more math being done on the frontend, and every number the user sees can be seen in big query.

## How will this move the needle on health equity?

[...]

PR #1534, opened 2022-03-30
adds merge_pop_numbers function

joshzarrabi:
* merges `population` and `population_pct` columns into an existing df
* for #1474 issues

## Motivation and Context

So we can have a consistent function to easily merge in population numbers.

## How has this been tested?

Yes

## Types of changes

### DIFF
--- a/python/ingestion/dataset_utils.py
+++ b/python/ingestion/dataset_utils.py
@@ -124,3 +124,30 @@ def merge_fips_codes(df):
     df = df.rename(columns={'state_fips_code': std_col.STATE_FIPS_COL}).reset_index(drop=True)
 
     return df
+
+
+def merge_pop_numbers(df, demo, loc):
+    """Merges the corresponding `population` and `population_pct` column into the given df
+
+      df: a pandas with demogrpahic and a `state_fips` column
+      demo: the demographic in the df, either `age`, `race`, or `sex`
+      loc: the location level for the df, either `state` or `national`"""
+
+    on_col_map = {
+        'age': std_col.AGE_COL,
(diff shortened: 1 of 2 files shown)


python/ingestion/dataset_utils.py
```diff
@@ -124,3 +124,30 @@ def merge_fips_codes(df):
     df = df.rename(columns={'state_fips_code': std_col.STATE_FIPS_COL}).reset_index(drop=True)
 
     return df
+
+
+def merge_pop_numbers(df, demo, loc):
+    """Merges the corresponding `population` and `population_pct` column into the given df
+
+      df: a pandas with demogrpahic and a `state_fips` column
```
benhammondmusic, line 132: typo

benhammondmusic: We should probably specify that the df is expected to have the items in the `demo` column be a race code (like `WHITE_NH`)

benhammondmusic (approved):
super helpful going forward. thanks

python/ingestion/dataset_utils.py
benhammondmusic: Sure. But I mean we should specify the exact format expected for that column's data. For race it needs to be race code, for sex it needs to be a string `Male` or `Female`, for age is has to be `10-14` or whatever. Like where we someone using that fn know to look for what was expected as a value in that col

joshzarrabi: I see. Maybe we should make like a demographic enum or something to enforce it. I think its a bit of a bigger issue than can be solved in the docstring of this function TBH

PR #1504, opened 2022-03-03
Scroll to Specific Highlighted Card from External #HashLink

benhammondmusic:
Example Links (make sure you don't have the tracker already open anywhere):
- https://deploy-preview-1504--health-equity-tracker.netlify.app/exploredata#map
- https://deploy-preview-1504--health-equity-tracker.netlify.app/exploredata#bar
- https://deploy-preview-1504--health-equity-tracker.netlify.app/exploredata#unknowns
- https://deploy-preview-1504--health-equity-tracker.netlify.app/exploredata#disparity
- https://deploy-preview-1504--health-equity-tracker.netlify.app/exploredata#table

# To Do

- [ ] consider how to bypass warm welcome modal if `#hash` is detected and scrolled to? maybe it's ok to leave it since we'll scroll and the welcome will stay up top? 
- [ ] works when opening a new tab / window; if an existing HET window is open though it doesn't properly detect the #card. I think this is a browser/React router thing because React router doesn't treat # changes as needed renders 

## Description
Another way of handling externally linking to a specific card; in this case appended `#table` or another hash will pause and let the full report load, and then smoothly scroll the user down until the card in question is vertically centered. It also adds highlighting to the specific card, which slowly fades away. The hash is automatically removed from the url

## Screenshots (if appropriate):
https://user-images.githubusercontent.com/41567007/156811673-9c7277eb-41f1-4522-a496-d3ab266f55ba.mp4

## Types of changes

- New feature (non-breaking change which adds functionality)

PR #1542, opened 2022-04-01
Manually revert scroll-to-id

benhammondmusic:
Scroll to ID #1504  was buggy particularly on Safari / Safari iOS. Will hold on this feature for now and hopefully figure out a reliable way to link external users directly to a specific card


joshzarrabi (approved):
Thanks

PR #1556, opened 2022-04-11
Disable localhost google tag tracking

benhammondmusic:
We don't want to pollute out Google Analytics with development usage. This PR will restrict the embedded Google Analytics to tracking pages served to `healthequitytracker.org`. This will specifically ignore:
- local dev `localhost`
- test on netlify
- dev on `dev.healthequitytracker.org`

## Motivation and Context

https://stackoverflow.com/questions/40297763/how-to-disable-google-analytics-on-localhost

## How has this been tested?
Tested manually by including logs based on the `window.location.hostname` matching to confirm condition application, and tested that tracking is suppressed by comparing active users on Google Analytics page and visiting various little used pages like Terms of Service

We will need to manually test this upon merging to dev, and also release to prod as it's impossible to fully test those without actual deployment

## Types of changes


joshzarrabi:
@benhammondmusic You can test this by running it locally, clicking around, and making sure google analytics doesn't register anything you are doing.

benhammondmusic:
> @benhammondmusic You can test this by running it locally, clicking around, and making sure google analytics doesn't register anything you are doing.

Cool, yep this works. I'll restrict tracking only to PROD unless you want to keep DEV tracking as well?

joshzarrabi (approved):

PR #1557, opened 2022-04-12
Update `merge_pop_numbers` to include territories

benhammondmusic:
## Description

Updates the `merge_pop_numbers()` fn to also retrieve the ACS2010 table when fetching state-level populations. 

- updated the `merge_pop_numbers` fn to merge ACS2010 
- included mock ACS2010 tables in the UHC tests
- updated the UHC test mocks to account for the extra calls to fetch ACS2010
- updated the assertions to match the number and names of specific calls

Split this change into its own PR to keep the size of #1512 smaller

## How has this been tested?

all backend tests pass

## Types of changes


python/ingestion/dataset_utils.py
```diff
@@ -139,15 +142,41 @@ def merge_pop_numbers(df, demo, loc):
...
+    # other territories from ACS 2010 (VI, GU, AS, MP)
+    if loc == 'state':
+        verbose_demo = "race_and_ethnicity" if demo == 'race' else demo
+        pop_2010_table_name = 'by_%s_territory' % (
+            verbose_demo)
+
+        pop_2010_df = gcs_to_bq_util.load_df_from_bigquery(
+            'acs_2010_population', pop_2010_table_name, pop_dtype)
+
+        if len(pop_2010_df.index) > 0:
```
joshzarrabi, line 173: Why is this check here? In what situation would we not want to merge it. It should always be there right?

benhammondmusic: when it mocks the fetch call for ACS2010 in BRFSS it returns an empty df, so the column names were causing an error. Maybe would be better to generate fake territory data with the correct columns names for the UHC tetss, even though the territory data wont be merged?

joshzarrabi: yeah, I would avoid adding more code to production just to get tests that are mocked differently than what we would expect to pass.

benhammondmusic: That makes sense. I added the test mock for fetching ACS2010 data with just Guam so that it is more like the regular production flow.

joshzarrabi (approved):

PR #1558, opened 2022-04-13
rename Total to All on the backend

joshzarrabi:
## Description

* Renames `Total` to `All` for all demographic categories on the backend. 
* The backend now has no mention of `Total`
* After we cut a release and update the prod data we can remove the `renameTotalToAll` function on the frontend everywhere.

## Motivation and Context

In service of: #963 and #1474

## How has this been tested?

Changed all test data. Manually tested to make sure it doesnt break anything. 

## Types of changes


python/datasources/acs_population.py
```diff
@@ -587,10 +587,10 @@ def _create_ingesters(self):
 
 def generate_national_dataset_with_all_states(state_df, demographic_breakdown_category):
     all_state_fips = set(state_df[std_col.STATE_FIPS_COL].to_list())
-    return generate_national_dataset(state_df, all_state_fips, demographic_breakdown_category)
+    return GENERATE_NATIONAL_DATASET(state_df, all_state_fips, demographic_breakdown_category)
```
benhammondmusic, line 590: why did you decide to capitalize this fn ?

python/ingestion/standardized_columns.py
```diff
@@ -32,7 +32,6 @@
 INCOME_COL = "income"
 POPULATION_PCT_COL = "population_pct"
 
-TOTAL_VALUE = "Total"
```
benhammondmusic, line 35: I'm not super clear how we will now access rows in our data sources that use the string `TOTAL` or `Total`. Doesn't the ACS data use total and not all? I know with the 2010 data we aren't pinging the API directly and just are getting it from `/data` but for regular ACS won't it need "Total"? Or is it always summing the groups to create an all/total row?

benhammondmusic (approved):

python/datasources/acs_population.py
joshzarrabi: Yeah I should change this and come up with a better naming scheme. We will probably have one of these functions for each datasource, or at least every datasource that needs to generate national totals on the fly based on states data quality. 

I capitalized it because after I ran `brew upgrade` pytest was giving me an error saying it wasn't named correctly to be exported. I think it wanted it to start with a capital letter, but I don't really understand why that is the only one it had an issue with.

python/ingestion/standardized_columns.py
joshzarrabi: Right, if you notice [here](https://github.com/SatcherInstitute/health-equity-tracker/pull/1558/files?diff=unified&w=0#diff-47d9e5580ee35db4b25d4a9f0b6a25a9334ef7c31e9257062178a013418352bfL39-R44) in `python/datasources/acs_population.py`, we map the value of whatever represents the "Total" value from the ACS API to whatever our `Total/All` value is. So we are simply changing that map from `Total` to `All`

python/datasources/acs_population.py
joshzarrabi: Fixed, IDK what I was doing wrong before. Just using the lower case version and . importing it.

PR #1561, opened 2022-04-20
Removes casting of TYPE and KEY for BRFSS UHC data

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1561--health-equity-tracker.netlify.app/" target="_blank">
  <br />

## Description

Previously we needed to cast data coming from BigQuery on the frontend to normalize both the type (everything was previously stored/exported as STRING) and also often to rename the KEY referenced by the frontend code. We no longer need to perform this step, so long as the backend is factored properly to put the data in the correct type and key

## How has this been tested?
Manually

## Types of changes


joshzarrabi (approved):
Beautiful, thank you

Issue #1529, opened 2022-03-22
Implement a way to reduce demographic toggle options


benhammondmusic:
**Is your feature request related to a problem? Please describe.**
Presently some data sets are not available for all three filter options `SEX` `AGE` `RACE`. For example, #1518 will only allow a view with breakdowns by race. Other data sets are missing some of these breakdown options at certain geographic levels (county, state, etc)

**Describe the solution you'd like**
To start, a clean and extensible solution at the `VariableId` level that will reduce the toggle options or simply hide the demographic toggle entirely. Eventually we might consider hiding demographic options per geographic level but that would be more difficult

**Describe alternatives you've considered**
The alternative is to leave all three options `AGE` `SEX` and `RACE` even though race is the only selection that will populate data into our visualizations. This means a screen full of "missing data" message. Is it helpful to allow the user to get to this type of page?

**Additional context**

benhammondmusic:
Follow up, we are going to proceed with keeping the demogrpahic toggle in place but disabling the unavailable options. This serves two purposes: 
- allows the user to see what other demographic toggles might be available given different topics/settings
- prevents the user from clicking to a screen with no information available
- allows a user who got to an unavailable report "by mistake" to click back to an available option (like, choosing AGE on COVID and then clicking over to WOMEN IN LEGISLATIVE OFFICE where AGE isn't an option) 
<img width="842" alt="Screen Shot 2022-04-21 at 11 27 43 AM" src="https://user-images.githubusercontent.com/41567007/164522695-d1ae9318-0787-4735-a445-cbeaa0f2555a.png">

PR #1562, opened 2022-04-21
Suppress Demographic UI

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1562--health-equity-tracker.netlify.app/" target="_blank">
  <br />
Click to demo updates </a> 

## Description

- Hardcodes "data gaps" of VariableIds by Demographic by Geographic Level toggle where we don't expect any data would prefer to not let the user select that filter option. 
- Suppresses the demographic group dropdown section of the `<MapCard />` when the only available option is `All`

## Motivation and Context

#1529 

## How has this been tested?

Manually

Screen Shot 2022-04-22 at 9 22 30 AM

## Types of changes


benhammondmusic:
Screen Shot 2022-04-21 at 12 37 58 PM

joshzarrabi:
@benhammondmusic We should implement this for state level vaccines, we only have data for race

frontend/src/data/utils/datasetutils.ts
```diff
@@ -178,6 +178,41 @@ export function shouldShowAltPopCompare(fromProps: ShouldShowAltPopCompareI) {
   return (
     fromProps.fips.isState() &&
     fromProps.breakdownVar === RACE &&
-    fromProps.variableConfig.variableId === COVID_VAXX
+    fromProps.variableConfig.variableId === "covid_vaccinations"
```
joshzarrabi, line 181: Is this a regression?

joshzarrabi (approved):
Nice work

frontend/src/data/utils/datasetutils.ts
benhammondmusic: No, I eliminated that CONST, since the only reason we were using so much was the previous hard-coding of UI suppression for COVID VAXX at county level.

PR #1567, opened 2022-04-22
Adds util function to load json from `/data` into a dataframe based on a list of keys

benhammondmusic:


joshzarrabi (approved):
Very fancy pandas usage, I didn't realize that function exists\

PR #1572, opened 2022-04-27
Fix issues with datasets including only `0` and `undefined`

### LINKED ISSUE #1391: "0" conditional rendering issue for no data in Vega
Looks like somewhere in the legend generation there is a condition using short circuit `&&` or `||` that needs to use a ternary instead `x ? y : z`

Screen Shot 2022-01-10 at 10 57 25 AM

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1572--health-equity-tracker.netlify.app/" target="_blank">
  <br />
Click to demo updates </a>   

## Description

- Improves the demographic group list filter to allow groups that contain `0` as data (as opposed to `undefined`). Now it only excludes groups that contain entirely undefined data for that metric
- switches from a gradient bar legend to a dot legend if the data set's upper and lower bound are the same (so for example, as in the bug issue, the dataset contains only a mix of `0`s and `undefined`s, which was causing the weird looking legend with an invisible legend gradient bar

## Motivation and Context

fixes #1391

## How has this been tested?

manually across several settings; frontend tests pass

Normal Legend:
Screen Shot 2022-04-27 at 10 11 55 AM

Legend for "No Range" data
Screen Shot 2022-04-27 at 10 11 45 AM

## Types of changes

### DIFF
--- a/frontend/src/charts/ChoroplethMap.tsx
+++ b/frontend/src/charts/ChoroplethMap.tsx
@@ -103,6 +103,13 @@ export function ChoroplethMap(props: ChoroplethMapProps) {
   // Dataset to use for computing the legend
   const legendData = props.legendData || props.data;
 
+  const legendLowerBound = Math.min(
+    ...legendData.map((row) => row[props.metric.metricId])
+  );
+  const legendUpperBound = Math.max(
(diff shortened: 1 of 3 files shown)


joshzarrabi (approved):
Awesome, great catch

PR #1584, opened 2022-05-09
Adds buckets for BJS Prisoners 2020 in ACS routine

benhammondmusic:
We need the ACS routine to generate buckets that match up with the BJS Prisoner 2020 Table in #1581

### DIFF
--- a/python/datasources/acs_population.py
+++ b/python/datasources/acs_population.py
@@ -154,7 +154,7 @@ def get_uhc_voter_age_bucket(age_range):
         return '55-64'
 
 
-# buckets for BJS prisoners 2020
+# buckets for BJS prisoners 2020 - National
 def get_bjs_age_bucket(age_range):
     if age_range in {'0-4', '5-9', '10-14', '15-17'}:
         return '0-17'
@@ -175,7 +175,7 @@ def get_bjs_age_bucket(age_range):
     else:
         return 'Unknown'
 
-# buckets for juvenile/adult
+# buckets for BJS prisoners 2020 - State
 
 
 def get_juv_adult_age_bucket(age_range):


python/datasources/acs_population.py
```diff
@@ -272,17 +310,25 @@ def write_to_bq(self, dataset, gcs_bucket):
...
+
         if not self.county_level:
             by_sex_standard_age_uhc = self.get_by_sex_age(
                 frames[self.get_table_name_by_sex_age_race()], get_uhc_standard_age_bucket)
             by_sex_decade_plus_5_age_uhc = self.get_by_sex_age(
                 frames[self.get_table_name_by_sex_age_race()], get_uhc_decade_plus_5_age_bucket)
             by_sex_voter_age_uhc = self.get_by_sex_age(
                 frames[self.get_table_name_by_sex_age_race()], get_uhc_voter_age_bucket)
+            by_sex_age_bjs = self.get_by_sex_age(
+                frames[self.get_table_name_by_sex_age_race()], get_bjs_age_bucket)
```
benhammondmusic, line 324: @joshzarrabi , any thoughts on how to get `get_bjs_age_bucket` to only run on national and not on state? I guess it doesn't hurt to have the extra buckets available at the state level, but they aren't needed for incarceration and take extra time to compute. I only see the flag above for `if not self.county_level` and didn't see any other class properties for like `self.national_level`

joshzarrabi: @benhammondmusic Here https://github.com/SatcherInstitute/health-equity-tracker/blob/0aa7406e2f2320e5a2d5b4fa0b7762999f2e03ad/python/datasources/acs_population.py#L290-L295, we automatically generate national level datasets based on the state ones. We do this because the lowest level the acs data goes to is state, so we need to generate the national based off of it. I would just leave it in the state dataset.

benhammondmusic: ahh gotcha. thanks!

PR #1595, opened 2022-05-19
Cdc refactor state county frontend

### LINKED ISSUE #1478: Calculate state and county covid per_100k and pct_share metrics on the backend
## Background

This is in service of https://github.com/SatcherInstitute/health-equity-tracker/issues/1474

Currently, in order to calculate the state and county `per_100k` and `pct_share` metrics we show to users we do the following

1. Generate the raw count of `cases`, `deaths`, and `hospitalizations` for each demographic group in each locality using the [cdc_restricted_local.py script](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/python/datasources/cdc_restricted_local.py).

2. Merge with population date in the [big query sql](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/config/data_sources/AGG_cdc_restricted.sql).

3. Calculate the `per_100k` metric on the frontend [using dataforge](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/CdcCovidProvider.ts#L195-L204).

4. Calculate the `pct_share` metric on the frontend [using dataforge](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/CdcCovidProvider.ts#L206-L266).

This is problematic because:

* It is impossible to test the entire dataflow end to end, we have to split up testing into two parts, leaving room for error.
* It requires the user to download much more data than necessary.
* The calculations are inefficient to do on the frontend.

## Proposed Implementation 

Merge the population and COVID data in python and create tables for each demographic with the following columns

* `cases_per_100k`
* `deaths_per_100k`
* `hosp_per_100k`
* `covid_cases_share_of_known`
* `covid_hosp_share_of_known`
* `covid_deaths_share_of_known`

## End state

I can look at big query and see the following tables filled out with the above columns

- [x] `cdc_restricted_data.by_race_county`
- [x] `cdc_restricted_data.by_sex_county`
- [x] `cdc_restricted_data.by_age_county`
- [x] `cdc_restricted_data.by_race_state`
- [x] `cdc_restricted_data.by_sex_state`
- [x] `cdc_restricted_data.by_age_state`

This code is delete from the frontend

- [x] https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/CdcCovidProvider.ts#L87-L268

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1595--health-equity-tracker.netlify.app/" target="_blank">
  <br />

## Description

Removes all unnecessary processing code for the state and county level COVID code. Assumes state and county level numbers are ready to go on the backend,

## Motivation and Context

Closes #1478

## How has this been tested?

Manually.

## Types of changes


frontend/src/data/variables/CdcCovidProvider.test.ts
```diff
@@ -34,6 +24,38 @@ const METRIC_IDS: MetricId[] = [
...
+  acsDatasetIds: string[],
+  baseBreakdown: Breakdowns,
+  breakdownVar: BreakdownVar,
+  rawCovidData: any,
+  finalCovidData: any
+) {
+  const acsProvider = new AcsPopulationProvider();
+  const cdcCovidProvider = new CdcCovidProvider(acsProvider);
+
+  var consumedDatasetIds = [];
```
benhammondmusic, line 38: is there a specific reason for `var` here? For React I think the general preference (and the way most of the code base is currently) is to prefer `const` or `let` over `var` as they are block scoped. `const` is usually the go to, and using `let` only if you need to mutate the variable. In this case `const` is probably preferred since we aren't re-assigning the array itself, only adding items to it.

frontend/src/data/variables/CdcCovidProvider.test.ts
```diff
@@ -34,6 +24,38 @@ const METRIC_IDS: MetricId[] = [
...
+
+  dataFetcher.setFakeDatasetLoaded(cdcDatasetId, [rawCovidData]);
+  consumedDatasetIds.push(cdcDatasetId);
+
+  for (var id of acsDatasetIds) {
```
benhammondmusic, line 43: same here; just usually default to `const` for these

frontend/src/data/variables/CdcCovidProvider.ts
```diff
@@ -170,124 +166,106 @@ class CdcCovidProvider extends VariableProvider {
...
+        "hosp_n",
+        "hosp_unknown",
+      ]);
+
+      df = df
+        .generateSeries({
+          // Calculate per100k
+          covid_cases_per_100k: (row) =>
+            this.calculations.per100k(row.covid_cases, row.population),
+          covid_deaths_per_100k: (row) =>
```
benhammondmusic, line 194: all this is only still here for national level frontend calculations, right?

frontend/src/data/variables/CdcCovidProvider.test.ts
joshzarrabi: sounds good

frontend/src/data/variables/CdcCovidProvider.ts
joshzarrabi: yup! Will delete it in the next story

PR #1598, opened 2022-05-20
Use wrapper fn in CAWP to prevent division by zero = None

benhammondmusic:
## Description

The util fn should normally return `null` if we get a percent of anything divided by zero. However, for CAWP data, we do want to represent 0 women of color divided by 0 women at all as `0.0%` rather than `null`

this PR wraps the normal util fn and returns early with 0.0 if both numerator and denominator are 0

## How has this been tested?

unit tests pass. Need to uncomment one more once util is fixed

## Types of changes


python/tests/datasources/test_cawp_women_legislators.py
```diff
@@ -17,6 +18,13 @@
 
 # UNIT TESTS
 
+def test_pct_never_null():
+    assert pct_never_null(1, 2) == 50
+    assert pct_never_null(0, 0) == 0
+    # TODO uncomment next test once util fn is fixed
```
benhammondmusic, line 24: @joshzarrabi when you make your changes to the util, can you also rm this comment and uncomment the last test?

joshzarrabi: Sure!

joshzarrabi (approved):
Thanks!

PR #1607, opened 2022-05-25
Add util fn to read a zip file from a remote URL as df

benhammondmusic:
## Motivation and Context

Used for #1581 

## How has this been tested?

tests pass

## Types of changes


joshzarrabi:
Does this need to be a separate PR? I know we want to keep our prs small but I don't necessarily think this makes sense on its own. I would like to see an example of it doing something.

benhammondmusic:
> Does this need to be a separate PR? I know we want to keep our prs small but I don't necessarily think this makes sense on its own. I would like to see an example of it doing something.

i thought about adding a test; I think I could use that zip utility to generate a zip file, and then pull it apart with this fn and confirm that the extracted file names match what is expected.... would that be a good idea?

PR #1610, opened 2022-05-27
gets the county names from big query

joshzarrabi:
## Description

Throws out the county names from the cdc restricted files and replaces them with county names from the census utilities big query bucket based on the fips code.

## Motivation and Context

In service of #1478 

We didn't do this originally and we were losing some county names. 

## How has this been tested?

tested manually 

## Types of changes


benhammondmusic (approved):
Nice

PR #1621, opened 2022-06-06
Safelist flaky URL

benhammondmusic:
## Description

Recently this url has been failing. Safelisting to avoid wasting time. Eventually #1533 should fix this... 

## Types of changes


eriwarr (approved):
Looking good 👍

PR #1623, opened 2022-06-10
Manually combine the newly disparate race/ethnicity fields

joshzarrabi:
## Description

Combines the `race` and `ethnicity` fields into the legacy `race_ethnicity_combined` field based on the given cdc logic.

## Motivation and Context
In the latest CDC data update they separated out the race and ethnicity fields. While this is a great update that has the potential to both lower the `UNKNOWN` race rate and help us  gain more insight into inequalities in the Hispanic community, we are not yet ready to reap those benefits without a significant design and engineering effort. So, for now, we simply recombine them to how they were before. 

The matrix that describes the combination logic from the CDC is shown here:

Screen Shot 2022-06-13 at 5 03 12 PM

## How has this been tested?
Yes, I added a unit test that tests every single case of logic based on the conversion matrix the CDC provided.

## Types of changes


python/datasources/cdc_restricted_local.py
```diff
@@ -117,6 +117,26 @@
...
+       We will keep ths in place until we can figure out a plan on how to display
+       the race and ethnicty to our users disagregted."""
+
+    def get_combined_value(row):
+        if row[ETH_COL] == 'Hispanic/Latino':
+            return std_col.Race.HISP.value
+
+        elif row[RACE_COL] in {'NA', 'Missing', 'Unknown'} or row[ETH_COL] in {'NA', 'Missing', 'Unknown'}:
+            return std_col.Race.UNKNOWN.value
+
```
benhammondmusic, line 131: shouldn't `race_col == NA` AND `eth_col == NA` result in "missing" which would be `NA` ?

benhammondmusic (approved):
sweet

python/datasources/cdc_restricted_local.py
joshzarrabi: That all equates to `UNKNOWN` to us

PR #1626, opened 2022-06-15
removes unneeded `dataset_utils.replace_state_abbr_with_names` function

joshzarrabi:
## Description

Removes the unneeded `dataset_utils.replace_state_abbr_with_names` as the `dataset_utils.merge_fips_codes` function fulfills the same functionality. 

## How has this been tested?

I re-ran the data pipeline and saw that the output is the same as the current output. 

## Types of changes


benhammondmusic (approved):
Oh ya; thanks totally forgot about this. Thanks!

PR #1629, opened 2022-06-15
separates merging utility functions into their own file

joshzarrabi:
## Description

Adds a `merge_utils.py` file that contains all of the functionality to merge datasets with numbers from big query.

## Motivation and Context

The `dataset_utils.py` file was getting too big and complicated.

## How has this been tested?

Yes

## Types of changes


benhammondmusic (approved):
nice. way cleaner importing the fns directly as well

PR #1630, opened 2022-06-15
Svi frontend

### LINKED ISSUE #75: CDC: Social Vulnerability Index (SVI) at the county level
https://www.atsdr.cdc.gov/placeandhealth/svi/data_documentation_download.html
https://www.atsdr.cdc.gov/placeandhealth/svi/documentation/SVI_documentation_2018.html
https://www.atsdr.cdc.gov/placeandhealth/svi/documentation/pdf/SVI2018Documentation-H.pdf

The spreadsheet also contains 15 underlying factors that go into the SVI, which we should also ingest to use if we want.

eriwarr:
## Preview Link
<a href="https://deploy-preview-1630--health-equity-tracker.netlify.app/exploredata?mls=1.covid-3.36081" target="_blank">
  <br />
Click to demo updates </a>  

## Description

This PR introduces the implementation of the social vulnerability index (SVI); which resides on the population card. The variable provider and methodology pages were also updated in this PR. 

## Motivation and Context
- closes #75

These changes allow the user the view the SVI at the county level. By using the SVI, communities that rank lower on the index can receive more resources such as food, water, medicine, etc. to help them recover or mitigate suffering and economic loss.

## How has this been tested?

Local tests were run initially to ensure that the data populated as expected by updating the public/tmp directory, then the tests were run on the cloud pipeline to ensure that data worked as it should from the actual API. 
Screen Shot 2022-07-13 at 1 26 18 PM

## Types of changes


benhammondmusic:
Looking great! I was just testing it on mobile and I think we need to ensure the "full profile" button goes below the new alert. This isn't an issue on desktop because they're side by side but we should try and vertically swap them for mobile 

 image

frontend/src/cards/ui/SviAlert.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return "low";
+  }
+  if (svi > 0.67) {
+    return "high";
+  } else return "medium";
+};
+
+const findColor = (rating: string) => {
+  if (rating === "high") {
+    return "#d32f2f";
```
benhammondmusic, line 24: If possible I'd love to stay consistent with our styling. Specifically:
- make a scss module for this new SviAlert component
- import scss variables
- only use sass variables like `sass.altGreen`; add new entries to the variable files if we don't have an existing color that works

frontend/src/cards/ui/SviAlert.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const color = findColor(rating);
+
+  return (
+    <>
+      {props.svi > 0 ? (
+        <Alert severity="info" style={{ margin: "10px" }}>
+          This county has a social vulnerability index of <b>{props.svi}</b>;
+          which indicates a{" "}
+          <a href="testing" style={{ textDecorationColor: color }}>
+            <span style={{ color: color, fontWeight: "bold" }}>
```
benhammondmusic, line 42: I know there are haters out there, but we generally just use `<b>` tags for bold text. Perhaps down the line we should create a global class to handle emphasized text, but for now let's just stay consistent with `<b>`. In general I think we should avoid inline CSS as well, except rare places where it's needed

frontend/src/cards/PopulationCard.tsx
```diff
@@ -67,9 +69,23 @@ export function PopulationCard(props: PopulationCardProps) {
     Breakdowns.forFips(props.fips).andAge(onlyIncludeDecadeAgeBrackets())
   );
 
+  const queries = [raceQuery, ageQuery];
+
+  if (props.fips.isCounty()) {
+    const sviQuery = new MetricQuery(
+      "svi",
+      Breakdowns.forFips(props.fips).andAge(onlyInclude("All"))
```
benhammondmusic, line 77: Can you file a new issue to switch this logic from using `Age` when we should really just use a new breakdown type for data that applies to the entire geography ?

frontend/src/cards/Card.module.scss
```diff
@@ -102,3 +102,15 @@
...
+  padding-left: 0px;
+  padding-right: 0px;
+}
+
+.SviContainer {
+  @media only screen and (min-width: $md) {
+    display: flex;
+    align-items: center;
+  }
+}
```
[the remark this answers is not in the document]
benhammondmusic: I think the Profile button should be left aligned when it's bumped to a new line on narrow screens, right? 
Screen Shot 2022-07-13 at 12 40 59 PM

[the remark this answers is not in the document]
benhammondmusic: Like this instead? I think it looks better on the left but open to other ideas
Screen Shot 2022-07-13 at 1 21 10 PM

benhammondmusic (changes requested):

joshzarrabi:
@eriwarr @benhammondmusic , For now I think its confusing that clicking the SVI takes you to the top of the methodology page. I know we want to redesign that whole flow but as it stands right now I don't think it works super well. How hard would it be to add an anchor to the SVI section of the methodology page?

benhammondmusic (approved):
Awesome!! Thanks so much man; I appreciate the attention to detail and the extra effort to continuously streamline and standardize the code base

PR #1631, opened 2022-06-17
properly update cdc_restricted-by_race_age dataset

joshzarrabi:
## Description

Update the by_race_age dataset in python, previously we were doing it SQL

## Motivation and Context

In service of #1474

## How has this been tested?

Ran it on my cloud environment.

## Types of changes

### DIFF
--- a/python/tests/datasources/test_cdc_restricted.py
+++ b/python/tests/datasources/test_cdc_restricted.py
@@ -250,7 +250,7 @@ def testWriteToBq(
     assert mock_bq.call_args_list[3].args[2] == 'by_sex_county_processed'
     assert mock_bq.call_args_list[4].args[2] == 'by_race_county_processed'
     assert mock_bq.call_args_list[5].args[2] == 'by_age_county_processed'
-    assert mock_bq.call_args_list[6].args[2] == 'cdc_restricted_by_race_and_age_state'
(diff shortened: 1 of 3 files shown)


python/tests/datasources/test_cdc_restricted.py
```diff
@@ -250,7 +250,7 @@ def testWriteToBq(
     assert mock_bq.call_args_list[3].args[2] == 'by_sex_county_processed'
     assert mock_bq.call_args_list[4].args[2] == 'by_race_county_processed'
     assert mock_bq.call_args_list[5].args[2] == 'by_age_county_processed'
-    assert mock_bq.call_args_list[6].args[2] == 'cdc_restricted_by_race_and_age_state'
+    assert mock_bq.call_args_list[6].args[2] == 'by_race_age_state'
```
benhammondmusic, line 253: assuming this is intentional to bring our table naming scheme more into alignment? I've noticed it's inconsistent between data sources / levels

benhammondmusic (approved):
cool

python/tests/datasources/test_cdc_restricted.py
joshzarrabi: This is intentional! the tables that start with `cdc_restricted` are the old ones and will go away completely after we do national calculations on the backend!

Issue #1477, opened 2022-02-22
Create CDC restricted national dataset on the backend


joshzarrabi:
## Background 

This is in service of: https://github.com/SatcherInstitute/health-equity-tracker/issues/1474

Currently, when we want to calculate the numbers for the national covid breakdown, we do it on the front end in [this block of code](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/CdcCovidProvider.ts#L100-L114). 

This is an issue because
* It requires us to do calculations on the frontend, which makes testing the data end to end impossible
* It requires us to also download the raw population dataset and do an inefficient dataforge merge

## Proposed Implementation

Aggregate the cdc covid national data on the backend, merge with the national acs data, and create tables with all of the information we want to show to the user.

The table should have the following columns, in addition to location and demographic data

* cases_per_100k
* cases_pct_share
* hosp_per_100k
* hosp_pct_share
* deaths_per_100k
* deaths_pct_share

## End State

This is finished when I can log onto big query and see the following tables with the above columns correctly filled out

- [ ] cdc_restricted_data.by_race_national
- [ ] cdc_restricted_data.by_age_national
- [ ] cdc_restricted_data.by_sex_national

## Notes

While working on this, we should also address this issue: https://github.com/SatcherInstitute/health-equity-tracker/issues/1471

benhammondmusic:
Should we pull #1426 into this as well ?

joshzarrabi:
> Should we pull #1426 into this as well ?

@benhammondmusic Yeah, we wont really need any of those columns anymore. I think it makes sense just to close that issue in favor of this one.

PR #1636, opened 2022-06-21
Cdc national backend refactor

joshzarrabi:
## Description

* Generates all cdc covid numbers on the backend in python.
* Removes all usage of cloud sql aggregation function for covid data

NOTE:

* Addresses #1471 so the death and hospitalizations numbers will be a little bit different than what we currently have.

## Motivation and Context

Fixes: #1471 
In service of: #1477, #1474

## How has this been tested?

Lots of tests, also checked it out manually

## Types of changes


benhammondmusic (approved):
looks great!

eriwarr (approved):
Looking good! 😊

PR #1639, opened 2022-06-28
Split `merge_fips_codes()` into `merge_state_fips_codes` and `merge_county_names()`

benhammondmusic:
## Description

We were overloading the `merge_fips_codes()` fn by having it:
- add state level fips codes based on the state names or postal codes
- add/overwrite county level names based on the county level fips codes

Since they are really two different things, I broke them into two different functions. 

## How has this been tested?

Tests have been updated and are passing

## Types of changes


joshzarrabi (approved):
Nice, thanks for this.

PR #1641, opened 2022-06-29
Fix dag

benhammondmusic:
- run merge_county_names once at the beginning rather than on each output breakdown table
- fix mismatched naming
- properly type columns that contain a mixture of ints and `null`


eriwarr (commented):
Looking good! 😊

eriwarr (approved):
Looking good! 😊

PR #1652, opened 2022-07-06
upgrade gcs_to_bq_util to work with pandas 1.4.x

joshzarrabi:
- pandas.read_json needs to take in a decoded string
- previously it was being read in from gcs as bytes
- see https://github.com/pandas-dev/pandas/issues/46935

## Motivation and Context

ACS pipelines broke after upgrading to pandas 1.4.3 in #1648

## How has this been tested?

Ran the pipelines

## Types of changes


benhammondmusic (approved):
Cool

PR #1653, opened 2022-07-07
remove usage of pd.append

joshzarrabi:
## Description

Its going away in future pandas versions

## Types of changes


benhammondmusic (approved):
Thanks! There might be a few more in the CAWP and other stuff I've done; I'll check

PR #1654, opened 2022-07-07
Rollback/Fix Flaky Playwright Tests

benhammondmusic:
## Description

Tests were behaving very differently on CI/CD versus locally; many of the tests were not properly running and were wrongly passing. 

This PR reduces the scope of testing but ensures they are being run properly; we can continue to build out the e2e testing now that the baseline is functioning properly. 

- Tests were importing exported consts from other test files; this seems to have been causing issues with the tests "passing" when they should have been failing
- Config file was relying on `process.env` to choose `headless` flag, which seems to have been problematic
- removed some extra UI interactions that were flaky
- Denver "flow" test was loading urls in the wrong order for what was being tested

## How has this been tested?

Playwright tests pass; and fail when i intentionally, temporarily change expected stuff to fail

## Types of changes


joshzarrabi (approved):
Thank you ben!

Issue #1588, opened 2022-05-12
UHC conditions aren't populating footer link to AHR


benhammondmusic:
**Describe the bug**
Each card has a footer listing the stat sources used in that card's calculations. Currently the UHC conditions like Vote Participation are only showing the footer link to ACS but not Americas Health Rankings 

**To Reproduce**
1. Go to "Voter Participation"
2. Look at the footer of each card

**Expected behavior**
Footer should show both ACS and AhR links

**Screenshots**

![image](https://user-images.githubusercontent.com/41567007/167977732-c2a6bad8-d760-4810-96b8-ddfc6df93711.png)

joshzarrabi:
Good catch. I could have sworn I fixed this a while ago. I guess it broke again.

PR #1655, opened 2022-07-08
Adds test confirming dataset ids stay in sync

### LINKED ISSUE #658: Create tests to verify that all dataset IDs are mapped to DataSourceMetadata
DataSourceMetadata matches the consumed dataset IDs from a MetricQueryResponse (populated in Variable Providers).
There is no mechanism to ensure eng are keeping the DataSourceMetadata and VariableProviders in sync.

benhammondmusic:
## Description

It was easy to let the ids drift apart between `MetadataMap` and `DatasetMeta`; this test ensures that there are no ids in one file that aren't also in the other. This would have prevented #1588 and will prevent similar future errors

## Motivation and Context

closes #658 

## How has this been tested?

manually breaking this test causes it to fail

## Types of changes


joshzarrabi (approved):
Nice

PR #1660, opened 2022-07-08
Fixes missing columns in `.csv` files

### LINKED ISSUE #1647: Our `.csv` files lack columns whose first rows contain empty values
**Describe the bug**

It appears the BQ csv export function drops any columns whose row1 value is empty (or `null`)

**To Reproduce**
1. go to "Downloads and Methodology"
2. scroll down and click "download" for BJS of KFF Vaccinations
3. download RACE at STATE

downloaded BJS csv is missing `race_contains_ethnicity` and `jail_per_100k`
downloaded KFF vaccination csv is missing `vaccinated_pct`, `population_pct` and others

**Expected behavior**
Columns would remain despite some of the row values (including the first) being `null`

**Screenshots**
Screen Shot 2022-06-30 at 11 42 38 AM
Screen Shot 2022-06-30 at 11 43 33 AM

**Additional context**
Known issue? https://stackoverflow.com/questions/69660855/how-to-export-a-null-field-on-bigquery

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1660--health-equity-tracker.netlify.app/" target="_blank">
  <br />
Click to demo updates </a>  

## Description

Previously they had extracted the column headers from the first object of the json, assuming all keys would be present. Then they had looped through the rows applying the format util fn and then eventually recombined everything manually.

Instead, this PR:
- removes some old code that was merging state fips from the frontend map
- uses a custom function to retrieve ALL keys throughout the array of objects and uses them as a header. The built in headers methods all use the same strange behavior of retrieving column names as the keys of the first object (and those keys don't register if their value is `null`)
- extends the `convertSpecialCharactersForCsv()` to convert explicit `null` into `""`
- applies the `convertSpecialCharactersForCsv()` function to all columns
- manually builds a CSV style string, joining values with a `,` and joining rows with a `"\r\n"` newline

This change should remove the CSV's reliance on the `null` check / `Number()` casting performed in `DataFetcher`. Then, once #1474 is complete moving all calculations into the backend and #1662 where all numbers are sent consistently as numbers, we should be able to remove the block of `null` check / `Number()` casting altogether

Note: Ideally we should be shipping this as meta data I think per data source? They mention that in their comment

## Motivation and Context

fixes #1647 

## How has this been tested?

I've added automated tests to ensure a fake df gets converted properly, including values that are strings which include commas inside of quotes (which is apparently problematic and why they needed to apply that fn across all columns

BEFORE, MISSING COLUMNS:
Screen Shot 2022-07-08 at 7 41 51 PM

AFTER, NO MISSING COLUMNS, NULLS AS BLANK, NO MALFORMED COMMAN-STRINGS:
Screen Shot 2022-07-15 at 1 41 55 PM

## Types of changes

-


frontend/src/data/loading/DataFetcher.ts
```diff
@@ -96,32 +96,48 @@ export class ApiDataFetcher implements DataFetcher {
...
+          death_n: row["death_n"] == null ? null : Number(row["death_n"]),
+          death_unknown:
+            row["death_unknown"] == null ? null : Number(row["death_unknown"]),
           population:
-            row["population"] == null ? null : Number(row["population"]),
+            row["population"] == null
+              ? null
+              : row["population"] == null
+              ? null
+              : Number(row["population"]),
```
eriwarr, line 123: I think that this nested ternary is checking the same value twice `row["population"] == null`

benhammondmusic: good catch! i think we can eliminate this whole weird null checking stuff because we are sending it in the correct type from BQ now (instead of strings to ships numbers)

frontend/src/data/utils/DatasetTypes.ts
```diff
@@ -59,32 +60,45 @@ export class Dataset {
...
+    const headersSet = new Set<string>();
+    for (const row of this.rows) {
+      for (const header of Object.keys(row)) {
+        headersSet.add(header);
+      }
+    }
+    return Array.from(headersSet);
+  }
+
+  toCsvString(): string {
```
joshzarrabi, line 76: Super interesting, I'm surprised there isn't a built in way to do this.

frontend/src/data/loading/DataFetcher.ts
```diff
@@ -96,17 +96,27 @@ export class ApiDataFetcher implements DataFetcher {
       datasetId.startsWith("acs_2010_population")
     ) {
       result = result.map((row: any) => {
-        return { ...row, population: Number(row["population"]) };
+        return {
```
joshzarrabi, line 99: Just making sure I understand that we can still get rid of these when everything is done on the backend?

joshzarrabi (approved):
Nice Ben

frontend/src/data/utils/DatasetTypes.ts
benhammondmusic: There is but it defaults to inferring the column names from the first json object/row. The API of the library they're extending in Dataforge seems to allow for passing in the column names explicitly as an argument but I wasn't able to get it working

frontend/src/data/loading/DataFetcher.ts
benhammondmusic: Yes I believe so; I've outlined what i think the overall solution should be here: #1662

PR #1661, opened 2022-07-09
Fix issue with Vera non-All rates being lost

benhammondmusic:
## Description

One of the last changes I made previously had been some INT64 conversion to allow a nullable int column for the rates. This seemed to be working and the tests were still passing, but actually many (all?) rows with demographics other than `All` were losing their `per_100k` rates. I believe this is because I was using a `.dropna()` but still not entirely sure. 

This PR:
- casts all numbers as floats, (so the previously int columns just end in `.0`). We will need to format on the frontend possibly, which we already do for other datasources
- 

## Motivation and Context

#1622 

## How has this been tested?

tests are still passing. although they were passing when the data was going missing in BQ as well so... 

## Types of changes


joshzarrabi (approved):
Do these per100ks still show up as `ints` on the frontend?

benhammondmusic:
> Do these per100ks still show up as `ints` on the frontend?

I'll confirm in the incarceration tweaks Pr. Just need to make sure we cast them as numbers and it should work

PR #1675, opened 2022-07-13
allow users to download covid data

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1675--health-equity-tracker.netlify.app/datacatalog?dpf=cdc_restricted" target="_blank">
  <br />

## Description

Allows users to download the the cdc_restricted datasets.

## Motivation and Context

* After finishing https://github.com/SatcherInstitute/health-equity-tracker/issues/1478 and https://github.com/SatcherInstitute/health-equity-tracker/issues/1477, all data sent to the frontend is exactly what the user sees on their browser.
* Previously, as we did some calculations on the frontend, the data contained more granular data than the cdc was comfortable with users seeing. 
* Since now the data is only exactly what the users see anyway, there is no more justification for stopping the user from downloading the raw data.
* Some users have asked for this.

## Types of changes

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -23,12 +23,7 @@ A clear and concise description of what you expected to happen.
 **Screenshots**
 If applicable, add screenshots to help explain your problem.
 
-**Desktop (please complete the following information):**
- - OS: [e.g. iOS]
- - Browser [e.g. chrome, safari]
- - Version [e.g. 22]
-
-**Smartphone (please complete the following information):**
+**Device (please complete the following information):**
  - Device: [e.g. iPhone6]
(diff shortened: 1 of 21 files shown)


benhammondmusic (changes requested):
Need to update the text to remove the "this isn't available for download" part. Also, is it worth putting in a sentence explaining its only rates and not raw numbers? For privacy etc? Lastly; should it be: national, state, county for the geographic level and corresponding text?

joshzarrabi:
> Need to update the text to remove the "this isn't available for download" part. Also, is it worth putting in a sentence explaining its only rates and not raw numbers? For privacy etc? Lastly; should it be: national, state, county for the geographic level and corresponding text?

Good catches, thanks

benhammondmusic (approved):
nice

PR #1677, opened 2022-07-14
update covid, county/national vaccines to july, kff vaccines back to april

joshzarrabi:
* KFF hasn't updated their state level dataset since april, better to be
  more accurate

## Preview Link

<a href="https://deploy-preview-1677--health-equity-tracker.netlify.app/" target="_blank">
  <br />
Click to demo updates </a>


benhammondmusic (approved):

joshzarrabi:
> This came up with CAWP from their review as well; we should clarify the "update frequency" as to whether it's when the source updates itself, when we re-pull from the source, or some lowest common denominator of both

@benhammondmusic Yeah idk, we need to think more about that.

PR #1678, opened 2022-07-14
React router hash link

eriwarr:
## Description

This PR introduces the React Router Hash Link library. This is a solution to React Router’s issue of not scrolling to #hash-fragments when using the <Link> component to navigating [view here](https://github.com/remix-run/react-router/issues/394). When you click on a link created with react-router-hash-link it will scroll to the element on the page with the id that matches the #hash-fragment in the link. This will also work for elements that are created after an asynchronous data load. 

## Motivation and Context

This is being introduced to resolve the navigation issue. When a user clicks on the link to the methodology page, the link should direct the user to the section of the page that is relevant to the variable they are inquiring about. 

## How has this been tested?

I've tested this library locally to ensure that the routes are working as they should. We currently have the react-router-dom library, but this library extends the react-router-dom library and adds further functionality. 

```
import { HashLink as Link } from "react-router-hash-link";

          <Link to={`${METHODOLOGY_TAB_LINK}#svi`} className={color}>
            <b>{rating} level of vulernability.</b>
          </Link>
```

## Types of changes


benhammondmusic:
@eriwarr i don't see where the package.json was updated with the new package? did you get this working yet? Honestly, the fact that our methodology page is a mess shouldn't hold up this PR; I think we should either have the SVI alert link directly to CDC or just link to the top of the Methodology and then treat this as a distinct issue

frontend/src/cards/ui/SviAlert.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import Alert from "@material-ui/lab/Alert";
+import { MetricQueryResponse } from "../../data/query/MetricQuery";
+import { Fips } from "../../data/utils/Fips";
+import { METHODOLOGY_TAB_LINK } from "../../utils/internalRoutes";
+import styles from "./SviAlert.module.scss";
+
+const cdcLink =
+  "https://www.atsdr.cdc.gov/placeandhealth/svi/documentation/pdf/SVI2018Documentation_01192022_1.pdf";
```
benhammondmusic, line 9: Toss this into `frontend/src/utils/externalUrls.ts`

benhammondmusic: into the `urlMap` object

PR #1679, opened 2022-07-15
Refactor Map Card UI Components

### LINKED ISSUE #1620: Refactor Map Info Box (with current rate and link to multimap) into its own component
**Is your feature request related to a problem? Please describe.**
The code and logic have grown significantly and it would be much easier to work with if it was an isolated component in its own file

**Describe the solution you'd like**
refactor the logic and content into a component

this alert should appear:
- when the current view (geographic level + demographic selection, including "All") contains a value. 
 
Meaning if it's on default `All` and the current view doesn't contain any demographic breakdowns and only totals, this alert should still appear. It should also still appear even when the lower level geographic view does not contain any data at all. For example, when viewing a report on `California`, if we don't have county data the map itself should not display, but we DO want the info box to show the rate for California as a whol

Screen Shot 2022-06-06 at 9 13 16 AM

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1679--health-equity-tracker.netlify.app/" target="_blank">
  <br />
Click to demo updates </a> 

## Description

extracts multiple smaller UI from the `<MapCard />` into their own components to improve code readability
- `<MultiMapLink />`
- `<RateInfoAlert />`

## Motivation and Context

closes #1620 

## How has this been tested?

e2e and unit tests pass

## Types of changes


joshzarrabi (approved):
Nice

PR #1681, opened 2022-07-15
specifies that the age-adjusted ratios are cloud sql FLOATS

joshzarrabi:
## Description

previously the age-adjusted ratios were STRINGS, now they are FLOATS

## Motivation and Context

A potential solution for #1647

## Types of changes


benhammondmusic (approved):
🎸

PR #1685, opened 2022-07-16
Generate time series/longitudinal COVID datasets

joshzarrabi:
## Description

Generates `_time_series` versions of each COVID dataset. These datasets have all the same fields as the originals but are also broken down on a `time_period` variable. 

The `time_period` variable is the year and month fields of the `cdc_case_earliest_dt` variable for the case from the original data. So this means that each `case`, `death` and `hospitalization` is being counted as part of the year and month the *case* was first reported. It is imported to note then that the deaths ARE NOT reported in the month the death occurred, but rather the month that that case was first reported.

## Motivation and Context

Address the backend component of: https://github.com/SatcherInstitute/health-equity-tracker/issues/1703

## How has this been tested?

I wrote some tests for the `time_series` versions of the dataset, but more manually testing needs to be done on them. This should be okay to merge though, as I have verified that it produces the exact same output for the non time series versions of the dataset. 

```
 ‣ python e2e_tests/scripts/ensure_datasets_equal.py -ep jzarrabi-het-infra-test-f4 -tp het-infra-test-05 -d cdc_restricted_data
checking het-infra-test-05.cdc_restricted_data.by_age_county_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_age_county_processed
checking het-infra-test-05.cdc_restricted_data.by_age_national_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_age_national_processed
checking het-infra-test-05.cdc_restricted_data.by_age_state_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_age_state_processed
checking het-infra-test-05.cdc_restricted_data.by_race_age_state against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_race_age_state
checking het-infra-test-05.cdc_restricted_data.by_race_county_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_race_county_processed
checking het-infra-test-05.cdc_restricted_data.by_race_national_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_race_national_processed
checking het-infra-test-05.cdc_restricted_data.by_race_national_processed-with_age_adjust against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_race_national_processed-with_age_adjust
checking het-infra-test-05.cdc_restricted_data.by_race_state_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_race_state_processed
checking het-infra-test-05.cdc_restricted_data.by_race_state_processed-with_age_adjust against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_race_state_processed-with_age_adjust
checking het-infra-test-05.cdc_restricted_data.by_sex_county_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_sex_county_processed
checking het-infra-test-05.cdc_restricted_data.by_sex_national_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_sex_national_processed
checking het-infra-test-05.cdc_restricted_data.by_sex_state_processed against jzarrabi-het-infra-test-f4.cdc_restricted_data.by_sex_state_processed
%%%%%%%%%%%%%%%%%%%%%% SUCCESS %%%%%%%%%%%%%%%%%%%%%%%%
```

## Types of changes


python/datasources/cdc_restricted.py
```diff
@@ -54,52 +58,62 @@ def upload_to_gcs(self, _, **attrs):
...
                 table_name = f'by_{demo}_{geo}_processed'
+                if not cumulative:
+                    table_name += '_time_series'
+
                 gcs_to_bq_util.add_df_to_bq(
                     df, dataset, table_name, column_types=column_types)
 
-        for filename, table_name in ONLY_FIPS_FILES.items():
-            df = gcs_to_bq_util.load_csv_as_df(gcs_bucket, filename)
+        # Only do this once, open to a less weird way of doing this
```
benhammondmusic, line 83: remind me why these "only fips" files are processed differently ?

python/datasources/cdc_restricted.py
```diff
@@ -113,13 +127,35 @@ def generate_breakdown(self, df, demo, geo):
             std_col.COVID_POPULATION_PCT,
         ]
 
+        if geo == 'national' or cumulative:
+            # We don't want the missing values for the time series data
+            # but we do need them to correctly calculate the national
+            # ALL per_100k rates
```
benhammondmusic, line 133: I think i need you to explain this in more detail; i still don't totally understand what's happening here

benhammondmusic (approved):
Thanks !

python/datasources/cdc_restricted.py
joshzarrabi: They are files that we dont want to do processing on so we can use them in the age-adjustment. We do the fips merging logic here so all we have to worry about in the age adjustment is age-adjusting.

python/datasources/cdc_restricted.py
joshzarrabi: When a racial group doesn't have any COVID cases for a state or county we still want them to show up on the on the table card. Also, when we calculate the national COVID numbers we want racial groups that don't have any COVID cases in a state count in that state's population when we calculate the national per_100k rates. However I don't want to inflate the size of the time_series state and county files, where we don't need the missing data as you do the calculation on the frontend to fill in nulls.

Issue #1689, opened 2022-07-18
Speed up frontend test by reusing React server


benhammondmusic:
**Is your feature request related to a problem? Please describe.**
As we improve test coverage of the frontend, the tests take an increasing amount of time

**Describe the solution you'd like**
- [x] since we are starting a server with the built react app for the `npm run test`, we should be able to reuse that running server for `npm run e2e`
- [x] there may be a way to cache node modules so that GitHub actions doesn't have to redownload these on every new push/test. Unless this is happening by default?

PR #1690, opened 2022-07-18
Reuse React Server on CI Tests

benhammondmusic:
## Description

- Rather than restart a new server for testing, it uses the existing server generated in the CI flow that is used by Jest
- More workers!

## Motivation and Context

Part of #1689

## How has this been tested?

Seems to finish ~30 seconds faster

## Types of changes


eriwarr (approved):
👌🏿

Issue #1611, opened 2022-06-01
Need County SVI on Population Card and Map Hover Tooltips


joshzarrabi:
## Is your feature request related to a problem? Please describe.

The [Social Vulnerability Index](https://www.atsdr.cdc.gov/placeandhealth/svi/index.html) is a ratio generated by the CDC that shows how vulnerable a county is to disasters and disease outbreaks. Closer to 0 means less socially vulnerable and closer to 1 means more socially vulnerable.

This gives the user context on the overall level of environmental and economic suffering in a county when looking at any of our other metrics.

Multiple users have asked us for this and it is something we have been discussing for a while.

## Describe the solution you'd like
When the user looks at the county level data they should be able to see the SVI of the county. 

I believe we should use the `RPL_THEMES` key from the SVI csv.

- [x] The user should be able to see it in the population card. Done in #1630 
- [x] Somewhere on the page the user should be able to see an explanation on what the SVI means. Done in #1630 
- [x] Methodology should be updated with SVI stuff
- [x] Updating `frontend/src/data/config/DatasetMetadata.ts` with the new datasetid so that the Sources footer populates on the PopulationCard whenever SVI is loaded (and eventually on the Map card(s) as well when we pull in the SVI info for the hover box)
- [x] Updating `frontend/src/data/config/MetadataMap.ts` so that the Data source appears and users could download a csv with our data
- [x] Adding SVI to the hover tooltip on STATE level report maps that are displaying COUNTIES
- [x] Linking from the Population card directly to the SVI section of the methodology using HashRouter
- [x] Ensure colors used for SVI are > 4.5 contrast (medium is currently 3.59)

<img width="1101" alt="Screen Shot 2022-07-15 at 4 36 04 PM" src="https://user-images.githubusercontent.com/41567007/179321457-a5675e39-736d-41d5-a264-802b366c1a55.png">
<img width="1352" alt="Screen Shot 2022-07-15 at 4 37 29 PM" src="https://user-images.githubusercontent.com/41567007/179321459-462456b9-82e6-4cd7-95c9-98af992982cb.png">
<img width="777" alt="Screen Shot 2022-07-15 at 4 48 43 PM" src="https://user-images.githubusercontent.com/41567007/179321576-e095160a-a651-4a36-88f7-aadfb5fc7b5c.png">

jgonzalezmsm:
[...]

PR #1702, opened 2022-07-20
Adds SVI to tooltip

eriwarr:
## Preview Link

<a href="https://deploy-preview-1702--health-equity-tracker.netlify.app/exploredata?mls=1.covid-3.36" target="_blank">
  <br />
Click to demo updates </a> 

## Description

This PR includes changes includes the following updates:
- Adds SVI to the hover tooltip on STATE level report maps that are displaying counties. 
- Updates the color used to display medium vulnerability to ensure the color is > 4.5 contrast. 

## Motivation and Context

This PR solves two of the remaining issues on issue #1611. 

Screen Shot 2022-07-21 at 4 25 31 PM

## Types of changes


eriwarr:
I added the rating to the tooltip as well. Please let me know if this is helpful or if it should only include the SVI.

benhammondmusic:
> I added the rating to the tooltip as well. Please let me know if this is helpful or if it should only include the SVI.

@eriwarr I think it's helpful; I wonder about putting it on the same line? So it would say:
```
County: Sarasota
Cases per 100k: 35,000
SVI: 0.23 (low)
```

or maybe even JUST the word?? Isn't that more helpful "at a glance" ? And then if someone wants to know more they click on the county and can see the full number on the lower level report. Hmm, I think I like that actually

benhammondmusic:
Like:
```
County: Sarasota
Cases per 100k: 20,043
County SVI: High Vulnerability
```

then even if a user has never heard of SVI they know 1) it's about vulnerability 2) whether it's really good/bad without needed to know the scale

eriwarr:
> 
I like this idea. The other idea you mentioned won't work as I'm unable to put the rating on the same line as the SVI for some reason. 


> Like:
> 
> ```
> County: Sarasota
> Cases per 100k: 20,043
> County SVI: High Vulnerability
> ```
> 
> then even if a user has never heard of SVI they know 1) it's about vulnerability 2) whether it's really good/bad without needed to know the scale

benhammondmusic:
> > 
> 
> I like this idea!!
> 
> > Like:
> > ```
> > County: Sarasota
> > Cases per 100k: 20,043
> > County SVI: High Vulnerability
> > ```

> > then even if a user has never heard of SVI they know 1) it's about vulnerability 2) whether it's really good/bad without needed to know the scale

Awesome, looks great. I would lowercase the work `vulnerability` (but not the "High" or "Low" etc). It's not 100% consistent but we should be defaulting towards sentence case rather than title case for most things

benhammondmusic:
@eriwarr Noticed something interesting; if the 100K isn't available it doesn't display the SVI. I think that's fine for now, but should we try and make it so it shows the SVI regardless of the presence of the other metric? 

Screen Shot 2022-07-21 at 3 50 19 PM
Screen Shot 2022-07-21 at 3 50 22 PM

frontend/src/cards/MapCard.tsx
```diff
@@ -156,10 +164,28 @@ function MapCardWithKey(props: MapCardProps) {
...
+          dataForActiveBreakdownFilter.map((row) => {
+            const thisCountySviRow = dataForSvi.find(
+              (sviRow) => sviRow.fips === row.fips
+            );
+            return {
+              ...row,
+              svi: thisCountySviRow?.svi,
+              rating: findRating(thisCountySviRow?.svi, true),
+            };
+          });
```
benhammondmusic, line 183: should we only do this kind of expensive nested loop filter if it's state or county level? like, 

```
let dataForActiveBreakdownFilter = ...

if (!props.fips.isUsa()) dataForActiveBreakdownFilter = mergeSviRows...
```

frontend/src/cards/MapCard.tsx
```diff
@@ -362,7 +388,9 @@ function MapCardWithKey(props: MapCardProps) {
                     data={
                       listExpanded
                         ? highestRatesList.concat(lowestRatesList)
-                        : dataForActiveBreakdownFilter
+                        : props.fips.isUsa()
+                        ? dataForActiveBreakdownFilter
+                        : dataForActiveBreakdownFilterWithSvi
```
benhammondmusic, line 393: I don't super love nested ternaries.... if we do what i mentioned above and just keep the `dataForActiveBreakdownFilter` array as the real data, merging SVI into it as needed, we could avoid the nested ternary here. Otherwise, I think adding a comment might help for future confused code readers

frontend/src/cards/ui/SviAlert.tsx
```diff
@@ -11,13 +11,22 @@ interface SviAlertProps {
...
   }
   if (svi > 0.67) {
+    if (map) {
+      return "High Vulnerability";
+    }
     return "high";
   }
+  if (map) {
+    return "Medium Vulnerability";
+  }
```
benhammondmusic, line 29: I think this makes it too confusing... can you leave this fn as it was but then just `uppercase()` and use a string template to add in vulnerability?

benhammondmusic: or make a wrapper function

```
findVerboseRating(svi: number){

    return `${findRating(row.svi).toUppercase()} vulnerability`

}

```

Just trying to avoid all those extra if checks and making single functions do too many things based on boolean flags

frontend/src/cards/ui/SviAlert.tsx
```diff
@@ -23,6 +23,11 @@ export const findRating = (svi: number) => {
   return "medium";
 };
 
+export const findVerboseRating = (svi: number) => {
+  const rating = findRating(svi);
+  return `${rating[0].toUpperCase() + rating.slice(1)} vulnerability`;
```
benhammondmusic, line 28: nice. i always forget JS doesn't have a built in string method to capitalize just the first letter... seems weird they didn't include it. Even CSS has it lol

benhammondmusic (approved):
awesome!

joshzarrabi (approved):
Nice job!

PR #1705, opened 2022-07-21
Added React Router Hash Link

eriwarr:
## Description

This PR introduces the React Router Hash Link library. This is a solution to React Router’s issue of not scrolling to #hash-fragments when using the <Link> component to navigating [view here](https://github.com/remix-run/react-router/issues/394). When you click on a link created with react-router-hash-link it will scroll to the element on the page with the id that matches the #hash-fragment in the link. This will also work for elements that are created after an asynchronous data load. 

## Motivation and Context

This is being introduced to resolve the navigation issue. When a user clicks on the link to the methodology page, the link should direct the user to the section of the page that is relevant to the variable they are inquiring about. 

## How has this been tested?

I've tested this library locally to ensure that the routes are working as they should. We currently have the react-router-dom library, but this library extends the react-router-dom library and adds further functionality. 

```
import { HashLink as Link } from "react-router-hash-link";

          <Link to={`${METHODOLOGY_TAB_LINK}#svi`} className={color}>
            <b>{rating} level of vulernability.</b>
          </Link>
```

## Types of changes


frontend/src/cards/ui/SviAlert.tsx
```diff
@@ -4,6 +4,8 @@ import { MetricQueryResponse } from "../../data/query/MetricQuery";
 import { Fips } from "../../data/utils/Fips";
 import { urlMap } from "../../utils/externalUrls";
 import styles from "./SviAlert.module.scss";
+import { HashLink as Link } from "react-router-hash-link";
```
benhammondmusic, line 7: What do you think about importing it as just `HashLink` ? Because use `Link` in other components and I anticipate using this for more interconnections in other components and it might be nice to just keep them obviously distinct?

frontend/src/cards/ui/SviAlert.tsx
```diff
@@ -47,9 +49,9 @@ function SviAlert(props: SviAlertProps) {
         <Alert severity="info" className={styles.Alert}>
           This county has a social vulnerability index of <b>{props.svi}</b>;
           which indicates a{" "}
-          <a href={urlMap.cdcSvi} className={color}>
+          <Link to={`${METHODOLOGY_TAB_LINK}#svi`} className={color}>
```
benhammondmusic, line 52: This is great man; so glad you got it working. I noticed that just visiting `/methodolgy#svi` directly in the url bar doesn't work... so that's why I wasn't able to use this package to solve the problem a couple months ago I mentioned. But this works great here and I think we can use it lots of other places on the site; especially once our methodology is cleaned up

benhammondmusic (approved):
awesome; nice work man

frontend/src/cards/ui/SviAlert.tsx
eriwarr: Makes perfect sense.

PR #1751, opened 2022-08-15
UHC Voter Bug Fix

### LINKED ISSUE #1732: uhc pipeline generating null `voter_participation_per_100k` for ages `65+`
**Describe the bug**
When I run the `uhc_data` pipeline, the `voter_participation_per_100k` column for age `65+` comes up as null. 

It should be the average of the Presidential and Midterm rows, however it seems that the America's Health Rankings people have gotten rid of the `"Voter Participation - Ages 65+ (Midterm)"` [row that we rely on to make the estimation](https://github.com/SatcherInstitute/health-equity-tracker/blob/a14f99dd6451da00a830a3f2e04eb5a2e48c7f2e/python/datasources/uhc.py#L255), and instead only have `Voter Participation (Presidential) - Ages 65+`. 

**To Reproduce**
1. Run the `uhc_data` Airflow pipeline
2. Look at `uhc_data.age_national` table in bug query
3. See that the `voter_participation_per_100k`  column for age `65+` is `null`.

**Expected behavior**
We should correctly calculate the voter participation rate for 65+ using the `Voter Participation (Presidential) - Ages 65+` Row.

eriwarr:
## Description

This PR fixes the bug introduced when AHR updated its dataset, and it refactors and removes unneeded code. 

- AHR's dataset has a typo, where the measure name for the 65+ category has an extra space at the end.

- This caused a bug as our code looks for the `65+` string without extra spacing. This is specific only to the 65+ voter participation line. The other measure names we use from AHR's dataset do not have this issue.

- This PR will replace the measure name on the data frame with the correct format. 

- Removed `get_average_determinate_value` function. Suggest that we only display breakdowns for presidential elections so that our data is more uniform since AHR no longer provides demographic breakdowns for midterm elections. 

Overall Issue:
When AHR updates their dataset, sometimes they change the spacing in the names of the measures. When we check for matched rows, extra spacing in the measure name will cause the data frame to leave out those rows because we check for string literals. This will create unexpected null values. We should introduce checks to counter this. 

## Motivation and Context

fixes #1732

## How has this been tested?

I tested this locally using pytests and on glcloud. 

This created a null value specifically for the `65+` category:
Screen Shot 2022-08-16 at 10 13 10 AM

No more demographic breakdowns for the Midterm section:
Screen Shot 2022-08-16 at 10 16 40 AM

## Types of changes


joshzarrabi:
@eriwarr Good work Eric.

I think we still need the `get_average_determinate_value` function as it is used to generate a single `voter_participation` value for each demographic group. The way the function works is that if there are two values it averages them, otherwise, like in the case for `65+` it only uses the one value that we have.

eriwarr:
> @eriwarr Good work Eric.
> 
> I think we still need the `get_average_determinate_value` function as it is used to generate a single `voter_participation` value for each demographic group. The way the function works is that if there are two values it averages them, otherwise, like in the case for `65+` it only uses the one value that we have.
> 
> The thing is, the tests should be failing, I am not really sure why they still passed without the function

Thanks, Josh! I think the tests are still passing because the calculation isn't done until after the row is matched with the demographic group. The function was already bypassing missing midterm values, so I just removed the function altogether and only used the presidential values.  I can add the function back to the All portion of the code, but I think it could still be left off the demographic breakdown section. However, I could see the argument for keeping it there if AHR  decides to include demographic midterms breakdowns in the future.

joshzarrabi:
> > @eriwarr Good work Eric.
> > I think we still need the `get_average_determinate_value` function as it is used to generate a single `voter_participation` value for each demographic group. The way the function works is that if there are two values it averages them, otherwise, like in the case for `65+` it only uses the one value that we have.
> > The thing is, the tests should be failing, I am not really sure why they still passed without the function
> 
> Thanks, Josh! I think the tests are still passing because the calculation isn't done until after the row is matched with the demographic group. The function was already bypassing missing midterm values, so I just removed the function altogether and only used the presidential values. I can add the function back to the All portion of the code, but I think it could still be left off the demographic breakdown section. However, I could see the argument for keeping it there if AHR decides to include demographic midterms breakdowns in the future.

I see what you mean, Yeah I think we should add it back for the ALL calculation.

python/datasources/uhc.py
```diff
@@ -244,22 +246,8 @@ def parse_raw_data(df, breakdown):
 
                     # BY AGE voter participation is avg of pres and midterm
```
benhammondmusic, line 247: update this comment if we are ignoring midterms now

python/datasources/uhc.py
```diff
@@ -204,18 +202,22 @@ def parse_raw_data(df, breakdown):
...
+
                     matched_row = df.loc[
                         (df[std_col.STATE_NAME_COL] == state) &
                         (df['Measure Name'] ==
                          ALT_ROWS_ALL.get(determinant, determinant))
                     ]
 
-                    # TOTAL voter_participation is avg of pres and midterm data
+                    # TOTAL voter_participation is taken from pres value
                     if determinant in AVERAGED_DETERMINANTS:
```
benhammondmusic, line 219: i think we can remove this `AVERAGED_DETERMINANTS` list now, since voting was the only one being averaged and we aren't averaging it anymore. Just treat it as a normal condition, making sure you deal with % vs 100k appropriately

benhammondmusic (approved):
This looks great. Can you spot check several different states for voting % and makes sure our data matches theirs?

Issue #1703, opened 2022-07-20
Add a time dimension for all COVID metrics


joshzarrabi:
**Is your feature request related to a problem? Please describe.**

PRD Here: https://docs.google.com/document/d/1WSdPSvww4njifeSR0Y2niZkwZIQ3InTv8wl9aaTGZt8/edit#heading=h.5h2374q35fo4

As described in #1699, we plan to add time based metrics to every health equity metric on the site.

We will do the COVID dataset first because, due to its line item origin, it provides the richest and most customizable data. Furthermore, each line has a specific symptom onset date, and, crucially, the latest version of the data contains *all* of the data. Thus, we can mold the data in any way that we want without having to tackle the complexity of storing multiple versions of the same dataset, which we will need to do to add time to any other dataset. 

We are working with the 2N design agency to design new visualizations for the `per_100k`, `pct_share`, and `unknown_share` metrics that show change over time, and we will adjust the schema based on the outcome of the design.

**Describe the solution you'd like**

So far the currently proposed flow will be to change the outputted data schema so that there is an extra column representing the month in which a covid patient first reported symptoms. 

Currently the data looks like:

| sex | state_fips | covid_cases |
| --- | ---- | --- |
| Male | 06 | 10
| Female | 06 | 5
| Unknown | 06 | 1

But after this I propose it would look like:

| sex | state_fips | covid_cases | time_period |
| --- | ---- | --- | --- |
| Male | 06 | 5 | 2021-02 |
| Female | 06 | 2 | 2021-02 |
| Unknown | 06 | 0 | 2021-02 |
| Male | 06 | 5 | 2021-03 |
| Female | 06 | 3 | 2021-03 |
| Unknown | 06 | 1 | 2021-03 |

--------------------------------------------------

### To transform into a time series dataset

Then, as we need a time series dataset, we need to transform the tables into a time dataset where each demographic category would be a key, with a value that is a list of (timestamp, value) tuples. ~We need to decide if this transformation happens on the backend or with d3 on the frontend.~ 

As we do not have a good way to store time series data in big query (I think its possible but does not match with our current workflow), I think we should try to do the transform on the frontend using d3.

So for example the per_100k series would look something like:

| Asian | Black |
| --- | ---- |
| (2021-02, 10,000), (2021-03, 11,000) | (2021-02, 5,000), (2021-03, 7,000) |

[...]

PR #1718, opened 2022-07-26
2N - Time Tracking Cards

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1234--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

- [x] Adds cards to OneVariableReport that will contain new 2N visualizations
- [ ] Adds to TwoVariableReport

## Motivation and Context

closes #1717 

## Types of changes

- New feature (non-breaking change which adds functionality or content)

PR #1754, opened 2022-08-16
Specify in card names that COVID and COVID vaccination rates are cumulative

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1754--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

Rename COVID and COVID Vaccination cards to specify that the metrics are cumulative for the pandemic, as opposed to our other metrics which show the most recent time period we have data for. For example Poverty shows the 2019 ACS numbers, 

## Motivation and Context

We believe it is important to differentiate between this and our other metrics which are only for the latest year. 
Also, this will help differentiate between the cumulative and time based metrics are planning on adding in #1703 

Screen Shot 2022-08-16 at 12 55 52 PM

Screen Shot 2022-08-16 at 12 56 01 PM
Screen Shot 2022-08-16 at 12 56 07 PM
Screen Shot 2022-08-16 at 12 56 14 PM
Screen Shot 2022-08-16 at 12 56 22 PM

## Types of changes


benhammondmusic (commented):
I don't know if cumulative is the best word here. I think it makes me associate too much with a case count. I wonder if "all time" or similar is more clear?

joshzarrabi:
@benhammondmusic I'm not a huge fan of "All Time", sounds too colloquial. I'm surprised you only associate cumulative with case counts, I think its the right word here.

benhammondmusic:
> @benhammondmusic I'm not a huge fan of "All Time", sounds too colloquial. I'm surprised you only associate cumulative with case counts, I think its the right word here.

I guess it's fine because we are explicitly describing the RATE as `COUNT / POPULATION`, so it works to write out `"CUMULATIVE" COUNT / POPULATION`. But I don't think we can call it a "cumulative rate" (not that you have), because the rates don't accumulate, only the counts do. The only title change you made here that I'm not sure fits this is the Age-Adjusted ratio, because again we aren't accumulating ratios, we are accumulating counts and then calculating the ratios. 

I kind of feel like we need to be specifying the time range of the cumulative counts or something, to make it clear that it's 3 years worth of case counts (and growing) divided by a population.

joshzarrabi:
Closing this in favor of #1718 Where the work will be done.

Issue #1757, opened 2022-08-18
County level views should only download data for the needed state


joshzarrabi:
**Is your feature request related to a problem? Please describe.**
Our county level time series datasets are very large, as they have data for every single county in the country. The `by_age_county` time series dataset is 11MB compressed, 258 MB uncompressed, and takes over ten seconds to download on a relatively fast internet connection and macbook pro. If I throttle the network to "Fast 3G" this download takes well over a minute, essentially rendering the card unusable. 

Also, if the users clicks through `age`, `race` and `sex`, then we will be storing over half a gig of data in their react cache. This is wasteful and unnecessary, as there is no time on the site where can see data from two different states on the same card.

We have also been thinking about this for a while, see https://github.com/SatcherInstitute/health-equity-tracker/issues/305, for a discussion about this from before the site was launched. 

**Describe the solution you'd like**
When a user looks at county level information on the UI, the site only downloads the county level information for the state the county is in. 
 
**Suggested Implementation**
Currently, our data files are split by `demographic` and `geographic level`, so when a user requests time series covid data for race in any county, they download the `cdc-restricted-data_by_race_county_processed_time_series.json' datafile. 

We should split the county level files by state as well, so if a user is looking at a county in California (FIPS 06) then the user would only need to download `cdc-restricted-data_by_race_county_processed_time_series-06.json`

We should do this for every county level file we have.

A good place to do the splitting would be in our [exporter/mian.py file.](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/exporter/main.py). We can loop through every state fips code there, download only those from the table, and then write the file with the correct name to s3.

**Describe alternatives you've considered**
We could, instead of splitting on the `exporter` step, we could split while generating the big query tables, and write a different table per state. This is problematic because it would make it much harder and more complicated to use those tables for debugging and finding insights. 

Also, we could split by county, and only download the data for the specific county a user wants. This is problematic because looking at the state level map would require us to hit 50 API endpoints, as we need data for every county.

**Additional context**
As noted [here](https://github.com/SatcherInstitute/health-equity-tracker/issues/305#issuecomment-767118239) we should not change the current flow of the downloads page. It will be a better user experience to only have to download all of the county information for the county instead of having to download 50 different files and stitch them back together.

PR #1758, opened 2022-08-18
Backend: Split county-level datasets by state fips

benhammondmusic:
## Description

Flow:
Python -> Big Query -> Exporter Step -> File in GCP -> React fetches from File

This PR adds a loop to the Exporter step which:
- grabs only rows matching the current iterated state FIPS
- writes that subset query to a new file which will be used by the frontend
- continues to write the entire file as normal, which will be used on the frontend only for the CSV downloads page

## Motivation and Context

#1757 

## Types of changes

### DIFF
--- a/exporter/main.py
+++ b/exporter/main.py
@@ -23,6 +23,7 @@ def export_dataset_tables():
     dataset_id = "{}.{}".format(project_id, dataset_name)
 
     bq_client = bigquery.Client()
+
     dataset = bq_client.get_dataset(dataset_id)
     tables = list(bq_client.list_tables(dataset))
 
@@ -32,16 +33,20 @@ def export_dataset_tables():
         return ('Dataset has no tables.', 500)
 
     for table in tables:
-        dest_uri = "gs://{}/{}-{}.json".format(export_bucket, dataset_name, table.table_id)
+        dest_uri = "gs://{}/{}-{}.json".format(
+            export_bucket, dataset_name, table.table_id)
         table_ref = dataset.table(table.table_id)
+
         try:
-            export_table(bq_client, table_ref, dest_uri, 'NEWLINE_DELIMITED_JSON')
+            export_table(bq_client, table_ref, dest_uri,
+                         'NEWLINE_DELIMITED_JSON')
 
             std_table_suffix = "_std"
             if not table.table_id.endswith(std_table_suffix):
                 continue
 
-            dest_uri = "gs://{}/{}-{}.csv".format(export_bucket, dataset_name, table.table_id)
+            dest_uri = "gs://{}/{}-{}.csv".format(
+                export_bucket, dataset_name, table.table_id)
             export_table(bq_client, table_ref, dest_uri, 'CSV')
         except Exception as err:
             logging.error(err)
@@ -53,7 +58,17 @@ def export_dataset_tables():
 def export_table(bq_client, table_ref, dest_uri, dest_fmt):
     """ Run the extract job to export the give table to the given destination and wait for completion"""
     job_config = bigquery.ExtractJobConfig(destination_format=dest_fmt)
-    extract_job = bq_client.extract_table(table_ref, dest_uri, location='US', job_config=job_config)
+
+    state_fips = "01"
+    state_fips_matcher = f'{state_fips}%'
+
+    query_job = bq_client.query(
+        f'SELECT * FROM {table_ref} WHERE county_fips LIKE {state_fips_matcher};',
+        job_config=job_config,
+    )
+
+    extract_job = bq_client.extract_table(
+        table_ref, dest_uri, location='US', job_config=job_config, query_job=query_job)
     extract_job.result()
     logging.info("Exported %s to %s", table_ref.table_id, dest_uri)
 


config/run.tf
```diff
@@ -115,6 +115,7 @@ resource "google_cloud_run_service" "exporter_service" {
 
   template {
     spec {
+      timeout_seconds = 60 * 10
```
joshzarrabi, line 118: Fancy

exporter/main.py
```diff
@@ -32,16 +33,23 @@ def export_dataset_tables():
         return ('Dataset has no tables.', 500)
 
     for table in tables:
-        dest_uri = "gs://{}/{}-{}.json".format(export_bucket, dataset_name, table.table_id)
+        # split up county-level tables by state and export those individually
+        export_split_county_tables(bq_client, table, export_bucket)
+
+        # export the full table
+        dest_uri = "gs://{}/{}-{}.json".format(
```
joshzarrabi, line 40: Should we change all these weird `.format` strings to `fstrings` while we are here?

exporter/main.py
```diff
@@ -51,12 +59,68 @@ def export_dataset_tables():
...
+            export_nd_json_to_blob(blob, nd_json)
+
+        except Exception as err:
+            print("ERR!", err)
+            logging.error(err)
+            return ('Error splitting county-level table, {}, into state-specific file: {}: {}'.format(table_name, state_file_name, err), 500)
+
+
+def get_table_name(table):
+    return "{}.{}.{}".format(
```
joshzarrabi, line 101: Same here. we should use `fstring`s

exporter/test_exporter.py
```diff
@@ -25,46 +29,141 @@ def client():
...
+
+
+# TEST ADDITIONAL COUNTY-LEVEL DATASET SPLIT FUNCTIONS
+_test_query_results_df = pd.DataFrame({
+    'county_fips': ["01001", "01002", "01003"],
+    'some_condition_per_100k': [None, 1, 2],
+})
+
+
+def _get_query_results_as_df(*args):
```
joshzarrabi, line 118: This is the right idea but I think

* All of the assertions should happen inside the test function. You should take the assertions out of the side effect function, and instead inspect the args list to make sure that the query is correct.
* You should make sure that the query string is exactly what you would expect. 

Looks great though.

exporter/main.py
```diff
@@ -17,46 +18,141 @@ def export_dataset_tables():
...
+        return
+
+    bucket = prepare_bucket(export_bucket)
+
+    print("Splitting county-level table by state-level fips")
+
+    for fips in STATE_LEVEL_FIPS_LIST:
+
+        state_file_name = f'{table.dataset_id}-{table.table_id}-{fips}.json'
+        print(state_file_name)
```
joshzarrabi, line 99: We should remove this print statement or make it more useful.

exporter/test_exporter.py
```diff
@@ -25,46 +29,156 @@ def client():
...
+    payload = {
+        'dataset_name': 'my-dataset',
+        'demo_breakdown': 'age'
+    }
+    response = client.post('/', json=payload)
+
+    assert response.status_code == 500
+    assert mock_split_county.call_count == 1
+
+#
```
joshzarrabi, line 115: Remove these

exporter/test_exporter.py
```diff
@@ -25,46 +29,156 @@ def client():
...
+
+    # ensure initial call to bq client and county-level calls per state/terr
+    assert mock_bq_client.call_count == 1
+    assert mock_query_df.call_count == NUM_STATES_AND_TERRITORIES
+    assert mock_prepare_blob.call_count == NUM_STATES_AND_TERRITORIES
+    assert mock_export.call_count == NUM_STATES_AND_TERRITORIES
+
+    # ensure generated ndjson for bq.storage matches expected ndjson
+    generated_nd_json = mock_export.call_args[0][1]
+    assert (generated_nd_json ==
```
joshzarrabi, line 158: I'm assuming we don't care about order here so lets sort these so we don't end up with annoying errors in the future. 
```
    assert (sorted(generated_nd_json) ==
 sorted(_test_query_results_df.to_json(orient="records",
                                           lines=True)))

File: exporter/test_exporter.py [hunk 8]
```diff
@@ -1,17 +1,21 @@
...
+
+from main import app, STATE_LEVEL_FIPS_LIST, get_table_name
 
-from main import app
+NUM_STATES_AND_TERRITORIES = len(STATE_LEVEL_FIPS_LIST)
 
-test_tables = [bigquery.Table("my-project.my-dataset.t1"),
-               bigquery.Table("my-project.my-dataset.t2_std"),
-               bigquery.Table("my-project.my-dataset.t3")]
+test_tables = [bigquery.Table("my-project.my-dataset.t1-sex"),
```
joshzarrabi, line 14: Global constants should be ALL_CAPS. Otherwise I'm assuming its defined in the function.

joshzarrabi (approved):
Great job, lets just clean up the print statements and then we can merge

airflow/dags/acs_2010_population.py
```diff
@@ -28,15 +28,35 @@
 acs_2010_bq_op = util.create_bq_ingest_operator(
     'acs_2010_gcs_to_bq', acs_2010_bq_payload, data_ingestion_dag)
 
-acs_2010_aggregator_payload = {'dataset_name': _ACS_2010_POPULATION_DATASET}
-acs_2010_aggregator_operator = util.create_aggregator_operator(
-    'acs_2010_aggregator', acs_2010_aggregator_payload,
+acs_2010_exporter_payload_race = {
+    'dataset_name': _ACS_2010_POPULATION_DATASET,
+    'demo_breakdown': "race"
```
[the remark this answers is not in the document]
benhammondmusic: i have been avoiding `demographic` because to me that seems like it'd expect the value to be `black` or `male`, e/g/ the actual demographic. in this case it's the type of demographic disaggregation so it'd used `demo_breakdown` but i agree we should standardize. open to whatever but i think better to reserve `demographic` for the actual groups?

exporter/main.py
```diff
@@ -17,46 +18,141 @@ def export_dataset_tables():
...
+        tables = [
+            table for table in tables if (
+                demo_breakdown in table.table_id
+            )
+        ]
+
     # If there are no tables in the dataset, return an error so the pipeline will alert
     # and a human can look into any potential issues.
     if not tables:
         return ('Dataset has no tables.', 500)
```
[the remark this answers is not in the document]
benhammondmusic: good idea

exporter/test_exporter.py
benhammondmusic: cool didt know sorted() was a thing

airflow/dags/acs_2010_population.py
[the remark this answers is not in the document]
benhammondmusic: switched to "demographic" since it's easier to make it work. but still don't love conflating the groups themselves with the method of breakdown

airflow/dags/util.py
```diff
@@ -55,7 +58,14 @@ def generate_bq_payload(workflow_id: str, dataset: str, filename: str = None,
                 var.
     url: The URL used for ingestion. This should be deprecated in favor of
          writing any metadata to GCS during the GCS step. It's temporarily
-         necessary since ACS directly requests metadata during BQ upload."""
+         necessary since ACS directly requests metadata during BQ upload.
+    cumulative: boolean flag used to direct the creation of either
+        the cumulative or the time-based dataset
+    demo_breakdown: string "race_and_ethnicity" | "age" | "sex".
```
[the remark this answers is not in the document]
benhammondmusic: no there are a bunch that need `_and_ethnicity` still

airflow/dags/acs_2010_population.py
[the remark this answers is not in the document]
joshzarrabi: That makes sense. I'm open to doing a mass renaming.

PR #1759, opened 2022-08-19
Fixes sticky URL params

### LINKED ISSUE #1551: Datatype URL param sticking when changing to different dropdown topic
**Describe the bug**
The tracker represents the currently selected datatype (if there is more than one option) in the URL like this `?dt1=covid_hospitalizations`. Because most of our health topics only have one sub-datatype, the old topic's datatype selection is still present in the URL even after selecting a new topic.

**To Reproduce**
1. open the tracker and selected the "covid hospitalizations" data type
2. click on the madlib dropdown and select "voter participation"
3. observe "covid hospitalizations" still in the URL as a datatype param

**Expected behavior**
The URL should reflect the current state of the generated report. I'm not sure if this erroneous datatype param will actually cause issues or if it will always be ignored, but it should probably just be overwritten with the single "data type" from the new dropdown topic

**Screenshots**
Screen Shot 2022-04-08 at 1 39 15 PM

eriwarr:
## Preview Link

<a href="https://deploy-preview-1759--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

This PR fixes the sticky datatype and demo param issue when selecting new MadLib topics:

- Currently, when a datatype is selected, it is generated in the URL param. When the user selects a new MadLib topic, the datatype from the previous MadLib topic remains in the URL.

- This PR fixes this issue, by removing the previously selected datatype when a new topic is selected, and populating a new datatype when one is selected from the datatype and demo buttons.

## Motivation and Context

Fixes #1551

## Types of changes

### DIFF
--- a/frontend/src/pages/ExploreData/ExploreDataPage.tsx
+++ b/frontend/src/pages/ExploreData/ExploreDataPage.tsx
@@ -18,12 +18,13 @@ import {
   PhraseSelections,
 } from "../../utils/MadLibs";
 import {
+  DATA_TYPE_1_PARAM,
+  DEMOGRAPHIC_PARAM,
   getParameter,
   MADLIB_PHRASE_PARAM,
   MADLIB_SELECTIONS_PARAM,
   parseMls,
   psSubscribe,
-  setParameter,
   setParameters,
   SHOW_ONBOARDING_PARAM,
   stringifyMls,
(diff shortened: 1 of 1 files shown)


joshzarrabi (approved):
Looks great

benhammondmusic (approved):
Nice

PR #1767, opened 2022-08-25
Adjust preload heights responsively

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1767--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

Makes some basic adjustments to card's preload heights, which will help in reducing "cumulative layout shift" and the annoying effect of content jumping around on the screen as things continue loading (like our maps, etc). 

Basically we know that the cards are generally longer on wider screens. We also know that the charts/tables (not maps) are about half as tall when the number of demographic groups is only 2/3 (for sex)

These are obviously not precise but in context do help quite a bit with reducing the reflowing of content

Also:
- put logic into a reuable hook
- reorganizes existing hooks into their own sub dir

## How has this been tested?

different screen sizes

## Screenshots

LEFT (BEFORE): maps cards have incorrect preload-height causing cards to jump
RIGHT (THIS PR): cards are approximately the size they will end up being; reducing the CLS

https://user-images.githubusercontent.com/41567007/187977099-50ba159f-53ea-45f1-a500-131e180d661f.mp4

## Types of changes


joshzarrabi (approved):
Nice

eriwarr (approved):
Looks good!

PR #1769, opened 2022-08-26
Change COVID zero'ing out rules for time based data

joshzarrabi:
## Description

Set most instances of a lack of COVID data of `cases`, `deaths`, and `hospitalizations` to `0` instead of `NULL`.

A detailed explanation can be found [here](https://docs.google.com/document/d/1LtWyd8lxZ4S-Zg3UNqhzniQ8DML_ZYH9tSw-IQ2HJiQ/edit?resourcekey=0-3eq4gC6UBOys9TPofCTEEQ#heading=h.972apg45bz2j).

In summary:

* If a jurisdiction (state or county level) has at any point recorded a case for a racial group, it is wrong to assume that they are simply not tracking that race in months where they reported zero cases, deaths, or hospitalizations, so now we will mark those are zero instead of null.
* Stop assuming deaths and hospitalizations should be null if all of the outcomes are marked as `Unknown`. It is a little weird we are doing this at all because otherwise we only look at `hosp_y` and `death_y`, and ignore the other outcomes.

NOTE: We are waiting on comment from the CDC before going forward with this.

## Motivation and Context

Needed for #1703.

Our current nulling strategy gives us weird looking chats like this:
Screen Shot 2022-08-29 at 4 36 29 PM

This is because, whenever there are zero reported deaths for a racial group, they are marked as `Null` and d3 just doesn't draw anything for that time period.

With the new strategy done in this PR we see a more reasonable looking chart:
Screen Shot 2022-08-29 at 4 38 35 PM

## How has this been tested?

Tested with unit tests and on gcp.

## Types of changes


eriwarr (approved):
Looks so good! Great job Josh!

benhammondmusic:
Is this a duplicate of #1787 ? @joshzarrabi

joshzarrabi:
@benhammondmusic and @eriwarr yes sorry this is out of date and should be closed in favor of https://github.com/SatcherInstitute/health-equity-tracker/pull/1787

PR #1778, opened 2022-09-07
explain how we calculate AHR numbers

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1778--health-equity-tracker.netlify.app/methodology" target="_blank">
  <br />
Click to demo updates </a> 

## Description

Explain how we calculate the `per_100k` and `pct_share` metrics for AHR variables.

## Motivation and Context

This is missing and is an oversight, it came up in a user interview.

## Types of changes

### DIFF
--- a/frontend/src/pages/DataCatalog/MethodologyTab.tsx
+++ b/frontend/src/pages/DataCatalog/MethodologyTab.tsx
@@ -277,7 +277,7 @@ function MethodologyTab() {
                     display these conditions at the county level either.
                   </li>
                   <li>
-                    All metrics sourced from Americas Health Rankings are
+                    All metrics sourced from America's Health Rankings are
                     calculated based on rates provided from their data API.
                     <ul>
                       <li>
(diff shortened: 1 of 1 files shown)


frontend/src/pages/DataCatalog/MethodologyTab.tsx
```diff
@@ -276,6 +276,22 @@ function MethodologyTab() {
...
+                    Rankings are taken directly from the datasource. If the
+                    metrics are presented as percentages rather than{" "}
+                    <b>per 100k</b> metrics, we simply multiply the percentage
+                    by 1000 to get a <b>per 100k</b> metric.
+                  </li>
+                  <li>
+                    To calculate the <b>percent share</b> number for each
+                    metric, we must estimate the total number of people with
+                    each condition. Because AHR only presents data as
+                    percentages, we do this by multiplying the percentage by the
```
benhammondmusic, line 290: maybe clarify that they only provide "incidence rates" since some (suicide) are as per 100k, and we don't want to conflate the two types of percentages (rates vs shares).

`Because AHR only presents data as prevalence rates, we do this by multiplying the rate by the`

benhammondmusic (approved):
awesome thanks for the quick turnaround. left a possible comment to consider clarifying the language since not all rates from AHR are as %

frontend/src/pages/DataCatalog/MethodologyTab.tsx
joshzarrabi: Sure, thanks

Issue #1699, opened 2022-07-19
Time tracking


joshzarrabi:
## Background
Many users have asked us about the ability to track health equity topics over time. Some have even gone as far as to describe our current iteration as a "health equity snapshot" as opposed to a health equity tracker. 

Furthermore, we got a grant to implement this and accepted it so we are going to do it. 

## What's Missing / Problems with the current approach
Currently there is no way to track changes in health equity topics over time. Every metric is simply a snapshot of the most recently updated version. 

## Proposed Solution
Work with 2N to design a way to view both the `per100k` and the `pct_share` metrics for each demographic in a specific geography over time, and then do the work to get all the needed old versions of the data on the backend.

We will use our COVID dataset to start off with because it is the most rich dataset, and the simplest to work with as the latest version of the COVID data has all historical data, and we can choose the time frame as we have the data down to the day.

After that we will go through each dataset one by one and pull in old versions of the data. [Here](https://docs.google.com/document/d/1-4-neD0EAmtu0hJ65InB5u0_D9ETi4LYLdJ1m4IbMVs/edit#) we document how much historical data there is for each datasource.

## Screenshots / Mockups
Coming soon

#### Stage 1 

We have working visualizations tracking the `pct_share` `per_100k`, and `unknown_pct` metrics for all COVID geographic and demographic pairs over the length of the pandemic (starting Jan 2020).

#### Stage 2

All datasources are plugged into the new visualizations and users can see them over time. 

## How will this move the needle on health equity?
Specifically how this will help advance health equity, and is possible who the target audience might be (policy influencers, non-technical users, etc)

Users can see the impact that policy has on health equity outcomes.

## Tasks:

- [x] https://github.com/SatcherInstitute/health-equity-tracker/issues/1703
- [x] #1717

jgonzalezmsm:
close this, open new issue for each topic

PR #1811, opened 2022-09-23
label Non-Hispanic races as (NH) instead of (Non-Hispanic)

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1811--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

* Rename the label on the backend for all data sources.
* Adds a footnote defining (NH) as Non-Hispanic

Note that the info message explaining why races don't automatically add up to 100% will be broken temporarily and shown on the wrong views. This will be fixed once we run the pipelines on prod.

## Motivation and Context

We decided to do this to make more space for #1699 and we want to do the renaming on the backend instead of the frontend.

## How has this been tested?

Locally

## Types of changes


benhammondmusic (approved):
sweet. we'll want to do LOTS of clicking around and confirm everything makes sense one the changes are in place

eriwarr (approved):

PR #1812, opened 2022-09-27
Improves loading

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1812--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

Lazy load components so they aren't included in the initial bundle size and make for a snappier page load

- FeedbackBox
- Onboarding (still loads part)
- All tab routes when possible

Also switches the blog to use localStorage instead of sessionStorage; this means the blog content will persist per device across multiple days instead of being deleted upon closing the tab. Since we update the blog infrequently this means that when a user loads the site, it will initially load the cached blog content, and then if it's been more than 30 seconds since it's refreshed the content it will re-fetch. 

## Motivation and Context

Speed: user experience and google ranking

## How has this been tested?

manually; tests still pass

## Types of changes

- Improvement


joshzarrabi (approved):
Nice

PR #1820, opened 2022-09-28
Fix for missing county sources

eriwarr:
## Preview Link

<a href="https://deploy-preview-1820--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

-This PR includes a fix to the missing sources on the cards. 

## How has this been tested?

-Tested locally w/unit tests
Screen Shot 2022-09-28 at 1 09 13 PM

## Types of changes


frontend/src/cards/ui/Sources.test.ts
```diff
@@ -0,0 +1,20 @@
+import { stripCountyFips } from "./Sources";
+
+describe("stripCountyFips", () => {
+  it("Should return string with fips code", () => {
```
joshzarrabi, line 4: It should return a list of strings withOUT fips codes

frontend/src/cards/ui/Sources.tsx
```diff
@@ -32,7 +32,20 @@ export function getDatasetIdsFromResponses(
   );
 }
 
-function getDataSourceMapFromDatasetIds(
+export const stripCountyFips = (datasetIds: string[]) => {
+  datasetIds.map((id) => {
+    if (/\d/.test(id)) {
```
benhammondmusic, line 37: can we add a comment here just quickly explaining what it's doing and why? I hate trying to read REGEX lol

frontend/src/cards/ui/Sources.tsx
```diff
@@ -72,6 +85,9 @@ export function Sources(props: {
 
   let datasetIds = getDatasetIdsFromResponses(props.queryResponses);
 
+  datasetIds = getDatasetIdsFromResponses(props.queryResponses);
```
benhammondmusic, line 88: delete this duplicate line?  it's already doing this on line 86 right?

frontend/src/cards/ui/Sources.tsx
```diff
@@ -72,6 +85,9 @@ export function Sources(props: {
 
   let datasetIds = getDatasetIdsFromResponses(props.queryResponses);
```
benhammondmusic, line 86: might consider `unstrippedDatasetIds` or something... in general i like avoiding mutating the variables just because it's easier to see what should be there at each step by the variable name alone

frontend/src/cards/ui/Sources.tsx
```diff
@@ -32,7 +32,17 @@ export function getDatasetIdsFromResponses(
   );
 }
 
-function getDataSourceMapFromDatasetIds(
+export const stripCountyFips = (datasetIds: string[]) => {
+  const strippedData = datasetIds.map((id) => {
+    //Remove number from county datasetId
```
benhammondmusic, line 37: I'd love to be more explicit; what exactly is this regex doing? is it searching the entire string for any digit? or is it specifically detecting a dash followed by 2 digits only when occurring at the end of a string? (that would be my preference to avoid other potential issues down the line).

frontend/src/cards/ui/Sources.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  });
+  it("Should return input when county level data isn't being displayed", () => {
+    const output = [
+      "cdc_restricted_data-by_race_county_processed_time_series-with_age_adjust",
+    ];
+    const input = [
+      "cdc_restricted_data-by_race_county_processed_time_series-with_age_adjust",
+    ];
+    expect(stripCountyFips(input)).toEqual(output);
+  });
```
benhammondmusic, line 19: can we add another test that sends in an array with more than one id, a mix of needing to be stripped / not needing to be stripped? and then expect the stripped array? I think that test would have failed and exposed the nested mapping issue

frontend/src/cards/ui/Sources.tsx
eriwarr: I updated the regex to check if the string contains a hyphen followed by any two digits, i.e. `-23`. I think checking the last 3 digits may be unnecessary as the backed end will only ever ship strings with the fips code attached to the end.

benhammondmusic (approved):
Looks great; thanks man!

frontend/src/cards/ui/Sources.tsx
joshzarrabi: Good call Ben!

joshzarrabi (approved):
Beautiful

PR #1821, opened 2022-09-28
parallelizes the export stages of all data pipelines

### LINKED ISSUE #1816: Run `exporter` Airflow tasks in parallel
Currently, we have three separate Airflow tasks to export our data to s3. One for `sex`, `race`, and `age`.

For our larger datasets these can take a decent amount of time, for example for COVID they each take five minutes, and there is no reason why they can't all happen in parallel.

We should run all `export` tasks in parallel so our pipelines finish faster.
Instructions on how to do this are here: https://betterdatascience.com/apache-airflow-run-tasks-in-parallel/

## Additional context
* I have not tested this out I just skimmed the article. If this doesn't work as I would assume then we can regroup.
* The main technical issue I can think of is that our exporter image could potentially get overloaded. Feel free to make it beefier if you need to.

joshzarrabi:
## Description

All export steps happen in parallel. 

## Motivation and Context

Closes #1816

## How has this been tested?

On my airflow

Note that the export steps are on the same level.
Screen Shot 2022-09-28 at 4 59 29 PM

## Types of changes

### DIFF
--- a/airflow/dags/acs_2010_population.py
+++ b/airflow/dags/acs_2010_population.py
@@ -55,8 +55,9 @@ acs_2010_exporter_operator_sex = util.create_exporter_operator(
 
 # Data Ingestion DAG
 (
-    acs_2010_bq_op >>
-    acs_2010_exporter_operator_race >>
-    acs_2010_exporter_operator_age >>
-    acs_2010_exporter_operator_sex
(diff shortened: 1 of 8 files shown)


benhammondmusic (approved):
this is awesome. nice find

PR #1829, opened 2022-10-04
calculate inequitable shares on the backend

### LINKED ISSUE #1827: Time series percent shares of deaths and hospitalizations shouldnt be zero when there are zero TOTAL deaths/hospitalizations
**Describe the bug**
When calculating the `percent share` metric for time series COVID data, we set deaths and hospitalizations to `zero` if the jurisdiction has ever reported a case for that race, as done and explained in #1787.

This is fine for the `per100k` metric, because it makes sense to say that, in cases where there were zero total deaths, each demographic group has `0 per100k` deaths.

However, it is problematic for the `pct_share` metric, because it is just wrong to say that each group had `0%` of the share of deaths. In that case it should be reported as `NULL` (division by zero)

**Steps to reproduce**
* Go here: https://deploy-preview-1718--health-equity-tracker.netlify.app/exploredata?onboard=false&mls=1.covid-3.02#inequities-over-time, (or to any county view)
* Notice how both sexes have their `inequitable share` drop to `-100%`. This is because they are both reporting a `0%` share, which is `0%` of their population: `(0% share of deaths - x% share of population) / x% share of population = -100%`

**Expected Behavior**
When there are zero deaths or hospitalizations at all, we can say that each demographic group "equally shares" the zero. So we should set the inequitable share in these cases to zero.

**Screenshots**
Screen Shot 2022-10-04 at 12 16 05 PM

**Additional context**
This will be complex to calculate on the frontend so we should add another column of data called `inequitable_share` that is calculated on the backend and is correctly set based on these rules.

joshzarrabi:
## Description

Calculates the inequitable share metric on the backend. 
Puts it into the following columns

* `covid_cases_inequitable_share`
* `covid_deaths_inequitable_share`
* `covid_hosp_inequitable_share`

Also sets the inequitable share to `0` if the `per_100k` of all known demographics in the jurisdiction add up to zero.

Note, this is slightly different behavior than we had before, where, in some cases the `per_100k` would round down to zero but we would still calculate the inequitable share from raw numbers. Now we are manually setting those to zero. 

## Motivation and Context

Fixes #1827.

## How has this been tested?

Testing locally and added automated data tests.

BEFORE:
Screen Shot 2022-10-07 at 5 17 58 PM

AFTER:
Screen Shot 2022-10-07 at 5 17 51 PM

Please note, however, the following other changes:

BEFORE:
Screen Shot 2022-10-07 at 5 18 58 PM

AFTER:
Screen Shot 2022-10-07 at 5 19 12 PM

This discrepancy is because, as you can see in the per 100k chart, those dates reported a `per_100k` of `0`
Screen Shot 2022-10-07 at 5 20 23 PM

## Types of changes


benhammondmusic (approved):
Merging to main, as the main bug of -100% is fixed. they may be some remaining weirdness but we can either fix in another release or improve documentation to explain.

PR #1832, opened 2022-10-05
Fixes chart titles on mobile

eriwarr:
## Preview Link

<a href="https://deploy-preview-1832--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

-Changed the size of font when on mobile to fit more of the text in on the Vega charts since there is not text wrapping property on SVGs.
-Broke the population vs share chart titles into more lines to prevent long titles from expanding outside the chart.

## Motivation and Context

-Currently when on mobile, some of the very long titles on the Vega charts, expand beyond the card view. 

## How has this been tested?

Tested locally on mobile view. 

## Types of changes

### DIFF
--- a/frontend/src/charts/utils.ts
+++ b/frontend/src/charts/utils.ts
@@ -121,6 +121,11 @@ export function createTitles({
   const time = "since Jan 2020";
   const isMobile = containerWidth < 700;
   const isComparing = window.location.href.includes("compare");
+  const timeTrackingData = [
+    "covid_cases",
+    "covid_deaths",
+    "covid_hospitalizations",
+  ].includes(variableConfig.variableId);
 
   if (trend) {
     return {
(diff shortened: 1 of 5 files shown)


frontend/src/charts/utils.ts
```diff
@@ -189,24 +195,28 @@ export function createTitles({
       chartTitle = population
         ? [
             `Population vs distrbution of`,
-            `total ${metric}`,
+            `total ${metric.toLocaleLowerCase()}`,
```
joshzarrabi, line 198: curious about this, was it an oversight from before?

joshzarrabi (approved):
Nice

frontend/src/charts/utils.ts
eriwarr: Yeah. I think Mahia suggested that we have sentencing casing for our titles and I just noticed that I missed this from before.

joshzarrabi: Cool

Issue #1438, opened 2022-02-07
Rework labeling process in Metric config


benhammondmusic:
Our metric config was designed with the pattern of providing phrase units like `% share of cases` or `asthma per 100k`, and then the React components programmatically stitch these units together to create the phrases needed for alerts, titles, labels, etc. 

As the number of health topics grows, along with "edge-case" reports (e.g. intersectional presentations like #1518 and #1554), we should consider shifting the pattern to the metric config providing an explicit string for each phrase, label, and title needed on the report. 

We should also couple the dropdown categories and dropdown variables with their associated metric configs so all these wordings and categorizations are in a single source.

Other things to look at:
- `.variableDisplayName` seems to be only uses as aria label and data attribute , but should likely be used in the definitions. - There is an issue with using card title in the info box as well. 
- all properties
- all use cases and required wordings

We should also be describing the CARD in the title of the CARD (like `Incidence Map`) and then providing the specific within the Vega viz itself in the title and legend (like the geography, the condition, the demographic filter applied, etc)

benhammondmusic:
added comments to see where each property is being used

```
      variableId: "voter_participation_pres",
      variableDisplayName: "Voter Participation", // aria label and data attr
      variableFullDisplayName: "Voter Participation (Presidential Election)", // definition boxes
      variableDefinition: `U.S. citizens ages 18 and older who voted in the last presidential national election.`, // definition box
      metrics: {
        per100k: {
          metricId: "voter_participation_pres_per_100k",
          fullCardTitleName:
            "Rates of Voter Participation (Presidential) Per 100k People", // map card header, map card info box, simple bar chart card header, data table column name
          shortVegaLabel:
            "voters in the last presidential election per 100k", // aria, map legend title, simple bar axis label, 
          type: "per100k",
        },
        pct_share: {
          metricId: "voter_participation_pres_pct_share",
          fullCardTitleName: "Share Of All Voter Participation (Presidential)", // part of unknown map card header, part of disparity bar chart card header, data table column name
          shortVegaLabel: "% of voters", // disparity bar labels and axis title, data table unit

```

PR #1838, opened 2022-10-07
Inequitable share frontend

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1838--health-equity-tracker.netlify.app/exploredata?onboard=false&mls=1.covid-3.02&demo=sex&dt1=covid_deaths#inequities-over-time" target="_blank">
  <br />
Click to demo updates </a> 

## Description

frontend for #1829 

## To Do
- [x] fix high/low default on ShareTrends
- [x] the metric isn't really a share... it's like a "pct_deviation" not sure if it's worth renaming the metric on the backend now? I just don't want to confuse ourselves thinking it represents a share of total cases, it's really "by what percent is the result off from the expected"
- [x] fine-tooth comb to confirm all labels, etc are transferred. We really still need to do #1438 because it's growing more and more confusing and the new title generating functions are kind of crazy. There should be more 1:1 of strings of text in the metric config and the text that appears on the page, and less string manipulation to slice and dice and end up with the right phrasing

Screen Shot 2022-10-11 at 9 37 01 PM

## Types of changes


benhammondmusic:
Screen Shot 2022-10-11 at 9 50 05 PM

frontend/src/data/config/MetricConfig.ts
```diff
@@ -333,11 +359,8 @@ export const METRIC_CONFIG: Record<DropdownVarId, VariableConfig[]> = {
         pct_share: {
           metricId: "covid_cases_share",
           fullCardTitleName: "Share of total COVID-19 cases",
-          chartTitle: "Share of total COVID-19 cases with unknown",
```
eriwarr, line 336: Removing the _chartTitle_  prop will cause the chart title on the Unknown maps to be partially blank.

eriwarr (changes requested):
This is the only thing I noticed. Everything else looks great!!

Unknown Map

joshzarrabi:
Looks good to me! Just update the col name to https://github.com/SatcherInstitute/health-equity-tracker/pull/1857 and we should be good to go.

eriwarr (approved):
Great job Ben!!!

PR #1842, opened 2022-10-10
Updated responsive view when comparing

### LINKED ISSUE #1841: Our responsive design needs to calculate based on card width, not page width
**Describe the bug**
Much of our responsive design is based on measuring the width of the browser window. This doesn't work for compare views though, and everything falls apart. I've start sending in a "isCompareCard" prop when something is the right hand card in compare; we should likely do something similar that gets passed around (or stored in context) and then every view can immediately know if it's showing the card on wide or skinny space. 

**To Reproduce**

any compare mode, as skinny as the browser can go before collapsing into the mobile single column stacked view

**Expected behavior**

Cards shouldn't be squished. In this example below the labels should collapse into a single column like they would on mobile

**Screenshots**

Screen Shot 2022-10-08 at 2 40 43 PM

eriwarr:
## Preview Link

<a href="https://deploy-preview-1842--health-equity-tracker.netlify.app/exploredata" target="_blank">
  <br />
Click to demo updates </a> 

## Description

- Updated the CSS to apply the proper styling to the menu component that contains the demographic legends. 
- Applied the grid to the demographic legend and removed the padding-left when on mobile view. 

## Motivation and Context

- Fixes #1841

## How has this been tested?

- Tested locally

- Previous:
Previous

- Fix:
fix

## Types of changes

### DIFF
--- a/frontend/src/charts/trendsChart/FilterLegend.tsx
+++ b/frontend/src/charts/trendsChart/FilterLegend.tsx
@@ -109,7 +109,6 @@ export function FilterLegend({
                 style={{
                   opacity: !selectedGroups?.length || groupEnabled ? 1 : 0.2, // failing a11y; need minimum opacity .55 ?
                 }}
-                name={getDataView()}
               >
                 {/* Legend Item color swatch */}
                 <div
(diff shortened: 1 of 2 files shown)


benhammondmusic (approved):
thanks! good catch with the div vs. menu

PR #1843, opened 2022-10-11
Title Function Refactor

### LINKED ISSUE #1390: 100k Map should better reflect selected demographic group
### PROBLEM: 
Currently the 100k map only displays the selected demographic group in the dropdown menu (and the info box if available). The dropdown box is somewhat easy to miss, and more importantly the demographic group is not reflected in the Card Title or the Vega Legend (in RED), potentially leading to a misrepresentation of the data. 

### SOLUTION:
Using the logic from the info box (in BLUE) we can add the optional selected demographic group (eg "Black or African American", "15-24") to at least the legend, and potentially also the card title. We also need to add to the A11y descriptions being created in #1389
Screen Shot 2022-01-10 at 10 23 44 AM

### LINKED ISSUE #1840: COVID needs to be all-caps
**Describe the bug**

We have decided to always have COVID-19 spelled as such, not "Covid" or "covid"; recently using `.toLowerCase()` looks like it's lowercased the word in the title of the pop vs share chart title

Example https://healthequitytracker.org/exploredata#population-vs-distribution

**Expected behavior**

COVID should always be CAPS, even though we are sticking with sentence-casing in general

rather than a sea of if statements, we should consider adding new properties to each condition's MetricConfig entry with the exact text we want on each part of the page. The type `VariableConfig` will always have to be updated

**Screenshots**

Screen Shot 2022-10-08 at 2 27 00 PM

### LINKED ISSUE #1850: Women in Leg. Office is % not 100k
**Describe the bug**

Titles need to be different for CAWP charts

**Expected behavior**

CAWP is the only data we show the RATE as PERCENT instead of 100K; so we need to fix this

**Screenshots**

Screen Shot 2022-10-13 at 9 39 37 PM

eriwarr:
## Preview Link

<a href="https://deploy-preview-1843--health-equity-tracker.netlify.app/exploredata?onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

This PR includes a refactor to the _createTitles_ function. 
-`createTitles()` has been removed, and now only  `createSubTitle()` is used in the util folder to determine the subtitles for specific charts. 
- `titleTextArray()` is now used to determine each chart's title. 
- `useMediaQuery()` has been added to help simulate text wrapping for the chart using the SVG components. 
- The chart titles have been added to the _MetricConfig_ component. 

## Motivation and Context

- Closes #1850
- Close #1840
- Closes #1390
- Currently, the _createTitles_  function uses many if checks to determine each chart's title and the title array. This can become confusing and possibly difficult to maintain. 

## How has this been tested?

- Tested locally

## Types of changes

- Refactor


frontend/src/utils/hooks/useCreateTitle.tsx
```diff
@@ -0,0 +1,21 @@
...
+  location: string
+) {
+  const isMobile = useMediaQuery("(max-width:800px)");
+  const isLarge = useMediaQuery("(max-width:1850px)");
+  const isTiny = useMediaQuery("(max-width:500px)");
+  const isComparing = window.location.href.includes("compare");
+  const titleTextArray = [metricConfig.chartTitle || "", location];
+  console.log(metricConfig.chartTitle);
+
+  if ((isComparing && isLarge) || isTiny) {
```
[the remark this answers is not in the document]
eriwarr: Excellent feedback. I didn't notice this at all!

benhammondmusic:
Looks like there's some linting issues preventing the build @eriwarr er

eriwarr:
@benhammondmusic any A11y Issues with this?

frontend/src/cards/DisparityBarChartCard.tsx
```diff
@@ -71,19 +72,14 @@ function DisparityBarChartCardWithKey(props: DisparityBarChartCardProps) {
...
-      metricConfig.fullCardTitleName
-    } in ${props.fips.getSentenceDisplayName()}`;
-  }
+  const chartTitle = useCreateChartTitle(
+    metricConfig.populationComparisonMetric as MetricConfig,
+    locationName
+  );
 
-  const HASH_ID: ScrollableHashId = "population-vs-distribution";
+  const filename = `${metricConfig.populationComparisonMetric?.chartTitle}${locationName}`;
```
benhammondmusic, line 80: I've run into issues with using optional chaining inside string template literals, I think if this metricConfig doesn't have a `populationComparisonMetric.chartTitle` value then the filename would end up being`undefinedFlorida` or something like that. it might be worth declaring a chartTitle variabel first with a fallback, and then using that in the templated filename string? like `const chartTitle = metricConfig.populationComparisonMetric?.chartTitle || "Some generic title"` just a consideration

frontend/src/charts/utils.ts
```diff
@@ -86,151 +78,21 @@ export function addMetricDisplayColumn(
...
-  variableConfig: VariableConfig;
-  fips: Fips;
-  demographic?: string;
-  breakdown?: string;
-  trend?: boolean;
-  share?: boolean;
-  population?: boolean;
-  unknown?: boolean;
+type subtitleProps = {
+  activeBreakdownFilter: string;
```
benhammondmusic, line 82: i think these should be custom types scoped tighter than just "string" right? 

```
activeBreakdownFilter: DemographicGroup; // can be "White" or "All" or "Male"
currentBreakdown: BreakdownVar; // can be "age" or "sex" or "race_and_ethnicity"
```

frontend/src/utils/hooks/useCreateChartTitle.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { MetricConfig } from "../../data/config/MetricConfig";
+
+const theme = createTheme({
+  breakpoints: {
+    values: {
+      xs: 0,
+      sm: 800,
+      md: 900,
+      lg: 1500,
+      xl: 1850,
```
benhammondmusic, line 11: i dont think these are the material breakpoints we use elsewhere. These are what i had found and added to the variable files a while ago : https://github.com/SatcherInstitute/health-equity-tracker/blob/main/frontend/src/styles/variables.module.scss#L105-L109

benhammondmusic: Maybe these were intentional choices for px values to make the layout work? if so that's totally fine but we should indicate that these breakpoints are different than the others we use for the rest of the layour responsiveness

benhammondmusic (approved):
awesome work; thanks for making all these tiny changes that add up to make the end result look much more professional and responsive

frontend/src/utils/hooks/useCreateChartTitle.tsx
eriwarr: The default breakpoints don't do well for mocking the text wrapping of the chart titles. I'll add some comments for them.

PR #1857, opened 2022-10-24
renames 'inequitable_share' col to 'pct_relative_inequity`

joshzarrabi:
## Description

We think `pct_relative_inequity` is a better name, because these are *not* absolute shares but rather percents of inequity.

## Motivation and Context

for #1827 and #1829

## How has this been tested?

Tested locally

## Types of changes

### DIFF
--- a/python/tests/ingestion/test_dataset_utils.py
+++ b/python/tests/ingestion/test_dataset_utils.py
@@ -261,7 +261,7 @@ def test_ensure_leading_zeros():
     assert_frame_equal(df, expected_df, check_like=True)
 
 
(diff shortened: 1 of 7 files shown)


python/tests/ingestion/test_dataset_utils.py
```diff
@@ -263,11 +263,11 @@ def test_ensure_leading_zeros():
 
 def testGenerateInequitableShareCol():
```
benhammondmusic, line 264: Worth renaming this test?

benhammondmusic (approved):
Sweet thanks Josh

Issue #1852, opened 2022-10-17
Assorted issues from watching UX videos


benhammondmusic:
## Tasks:
- [x] hover tooltip doesn’t appear on multimaps for some reason. 
- [x] Need to swap some 0 per 00k for <1 per 100k on Vega too
- [x] #1907
- [x] we should organize our external data sources, perhaps promoting higher reputation sites higher (e.g. CDC, Census, etc to the top, and ProPublica lower down)
- [x] issues with the charts loading on the population card; I have personally experienced this as well
- [x] blog thumbnails still weren't loading on one person's machine after quite a while
- [x] #1908
- [x] phrase in missing data section "there is no" data implies there's no data ANYWHERE, when really we mean there is no data from the sources we have found. We could address that it's available for specific counties but not NATIONALLY at the county level for every county

PR #1864, opened 2022-10-27
Don't make Vega show "0 per 100k"

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1864--health-equity-tracker.netlify.app/exploredata?mls=1.covid-3.13239&dt1=covid_deaths&onboard=false" target="_blank">
  <br />
Click to demo updates </a> 

## Description

Relatively small case counts will round down to "0 per 100k", but this can mask the very real impact to those small number of people. Rather than display 0, it's better to use "<1" which conveys the same information but makes it clear the value wasn't necessarily exactly 0.  

- changes in the vega map hover tooltip formatting
- changes in the `formatFieldValue()` util which changes it everywhere else like the bar charts, tables, highest lowest, etc. 

## Motivation and Context

item from #1852 

Screen Shot 2022-10-27 at 8 30 36 AM
Screen Shot 2022-10-27 at 8 32 28 AM
Screen Shot 2022-10-27 at 8 33 54 AM

## Types of changes

### DIFF
--- a/frontend/src/charts/ChoroplethMap.tsx
+++ b/frontend/src/charts/ChoroplethMap.tsx
@@ -174,7 +174,10 @@ export function ChoroplethMap(props: ChoroplethMapProps) {
     const geographyName = props.showCounties
       ? countyOrEquivalent
       : stateOrTerritory;
-    const tooltipDatum = `format(datum.${props.metric.metricId}, ',')`;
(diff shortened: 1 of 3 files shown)


joshzarrabi (approved):
Nice

Issue #1786, opened 2022-09-09
Figure out how to calculate `pct_shares` for AHR conditions


benhammondmusic:
**Describe the bug**

Previous to #1792  many pct_share values displayed for AHR sum not to 100%. They are now nulled until we figure out how to proceed.

**Expected behavior**
We expect pct_shares to sum to 100% *ONLY IF* we are using demographic groups that cover the entire population. There are quite a few times when this is not the case:
- Hispanic-inclusive race groups (hispanic people identifying as a particular race could be double counted)
- Conditions where we only have data for some of the groups but there may be cases occurring outside of those groups. For example VOTER PARTICIPATION is lacking by race 100k values for NHPI, AIAN, 2+, and Unrepresented. Can we detect this by comparing our inferred "All" estimated total with the sum of each group's estimated total? 

Additionally, our calculations may require INTERSECTIONAL population figures to calculate accurately. Previously we were taking AHR's GROUP 100K and estimating the total case count using that group's population. However, many of the conditions from AHR only measure a subset of the population, meaning to proceed this way we would likely need to ensure the population figure matches the exact demographic breakdown. This might not be possible for conditions. 

### AHR rate denominators:

- "Chronic Obstructive Pulmonary Disease": adults 
- "Diabetes": adults
- "Frequent Mental Distress": adults
- "Depression": adults
- "Excessive Drinking": adults
- "Non-medical Drug Use": adults
- "Illicit Opioid Use": adults
- "Non-medical Use of Prescription Opioids": adults
- "Asthma": adults
- "Cardiovascular Diseases": adults
- "Chronic Kidney Disease": adults
- "Avoided Care Due to Cost": adults
- "Suicide": population (All ages? or only 15+? no group presented for Ages 0-14)
- "Preventable Hospitalizations": Medicare beneficiaries ages 18 and older enrolled in the fee-for-service program
- "Voter Participation": U.S. citizens ages 18 and older. ([original Census tables 4b, 4c](https://www.census.gov/data/tables/time-series/demo/voting-and-registration/p20-585.html) I think)

**More Info**

**To Do**
- [x] hide current pct_share metrics that we can't be confident in
- [x] reach out to AHR and determine best path forward (matching their exact population source?)
- [ ] implement new method for determining `raw count` and thereby `pct_share`
- [ ] add some type of testing to all sources using non-overlapping demographic groups that ensures pct_share metrics sum to 100% (with some level of tolerance for rounding errors) #1805

[...]

PR #1877, opened 2022-11-01
BACKEND: Adds 18+ pop for each sex and race group, for use in AHR `pct_share` calcs

benhammondmusic:
## Description

- [x] creates a second census population estimates datasource
- [x] processes the csv files from census by selecting unique combos of state/group/year and sums populations for 18+ only
- [x] writes to new BQ tables: `sex_age` and `race_age` for `state` and `national`

## Motivation and Context

#1786 

## How has this been tested?

- Adds automated tests 

Manually compared the results a few different ways
- using Numbers I cropped out all rows that weren't applicable to MALES 18+ in ALABAMA of ANY RACE, specifically HISPANIC or NON-HISPANIC and summing the 2021 column was exactly the same as my output file's corresponding values
Screen Shot 2022-11-03 at 3 22 31 PM

- using a Census file that only has 18+ populations by state (not broken down by race) I compared the state 18+ totals to my calculated race/eth == "ALL". The numbers were in the right ballpark, but this comparison census file is from 2019 and my file calculates 2020 and 2021 values. 
Screen Shot 2022-11-03 at 3 28 48 PM

Screenshot from my BQ showing 4 tables with correct columns
Screen Shot 2022-11-08 at 5 15 02 PM

## Types of changes


eriwarr (approved):
Looks good 👍🏿

python/datasources/census_pop_estimates_sc.py
```diff
@@ -0,0 +1,195 @@
+from ingestion.standardized_columns import Race
+from datasources.data_source import DataSource
+from ingestion import gcs_to_bq_util
+import ingestion.standardized_columns as s
```
joshzarrabi, line 4: we import this as `std_col` everywhere else, can we keep that pattern for now

python/tests/datasources/test_census_pop_estimates_sc.py
```diff
@@ -0,0 +1,134 @@
...
+    assert mock_csv.call_count == 1
+
+    # 4 = 2 demographic breakdowns X 2 geographic breakdowns
+    assert mock_bq.call_count == 4
+    assert mock_gen.call_count == 4
+
+# TEST INNER FUNCTION - RACE BY STATE
+
+
+def test18PlusByRace():
```
joshzarrabi, line 68: Can you specify these are by state in the function title.

joshzarrabi (approved):
Looks great, thank you ben

python/datasources/census_pop_estimates_sc.py
benhammondmusic: sure. I got tired of typing lol

python/tests/datasources/test_census_pop_estimates_sc.py
benhammondmusic: good call

PR #1889, opened 2022-11-08
Adds chart title to age-adjust table

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1889--health-equity-tracker.netlify.app/exploredata?dt1=covid_deaths#age-adjusted-risk" target="_blank">
  <br />
Click to demo updates </a> 

## Description

Adds a chart level title to the age-adjusted ratios table.

"Risk of COVID-19 `<death/hospitalization>` compared to White (NH) in `<location>`"

## Motivation and Context

We need to make sure it is clear that this is comparing to white NH 

## How has this been tested?

Locally

## Screenshots (if appropri
Screen Shot 2022-11-08 at 11 54 51 AM
ate):

## Types of changes


frontend/src/charts/AgeAdjustedTableChart.tsx
```diff
@@ -128,28 +130,42 @@ export function AgeAdjustedTableChart(props: AgeAdjustedTableChartProps) {
...
-              {rows.map((row: Row<any>, index) => (
-                <TableDataRow row={row} key={index} />
-              ))}
-            </TableBody>
-            <TableFooter>
-              <TableRow />
-            </TableFooter>
-          </Table>
-        </TableContainer>
+        <>
```
benhammondmusic, line 148: this is a good idea using <figcaption> for the table header, but to be accessible I think it meeds to be wrapped in a <figure> component. So you can just swap these React fragments `<>` for `<figure>`s and i think it'll still work well and be accessible. 

should be:
```
<figure>
<figcaption>some title</figcaption>
<table>...</table>
</figure>
```

frontend/src/data/config/MetricConfig.ts
```diff
@@ -459,6 +459,7 @@ export const METRIC_CONFIG: Record<DropdownVarId, VariableConfig[]> = {
           metricId: "death_ratio_age_adjusted",
           fullCardTitleName:
             "Age-adjusted risk of COVID-19 death compared to White (NH)",
+          chartTitle: "Risk of COVID-19 death compared to White (NH)",
```
benhammondmusic, line 462: i think the word `in` is missing.... looks like @eriwarr  has been including it in the metric config properties, but maybe it'd be better to move it out of the config and into the title function. Either way, rn the chart title says "Risk of COVID-19 death compared to White (NH) the United States" (missing `in`)

frontend/src/data/config/MetricConfig.ts
```diff
@@ -520,6 +521,7 @@ export const METRIC_CONFIG: Record<DropdownVarId, VariableConfig[]> = {
           fullCardTitleName:
             "Age-adjusted risk of COVID-19 hospitalization compared to White (NH)",
           shortLabel: "Risk of COVID-19 hospitalization", // Table header-row label
+          chartTitle: "Risk of COVID-19 hospitalization compared to White (NH)",
```
benhammondmusic, line 524: same "in" problem

frontend/src/charts/AgeAdjustedTableChart.tsx
joshzarrabi: Cool

frontend/src/data/config/MetricConfig.ts
joshzarrabi: good call. I think it should be in the function too

eriwarr: That's a good approach, I think we'll just have to update the metric config to remove the "in", and then change the chart title to   

```
  const chartTitle = `${
    metricConfig.chartTitle ? `${metricConfig.chartTitle} in`: ""
  } ${breakdown ? `${breakdown} in` : ""}`;
```

benhammondmusic (approved):
sweet

eriwarr (approved):
👍🏿

PR #1897, opened 2022-11-16
Have `<UnknownsMapCard />` handle all `0` values

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1897--health-equity-tracker.netlify.app/exploredata?mls=1.women_in_legislative_office-3.00#unknown-demographic-map" target="_blank">
  <br />
Click to demo updates </a> 

## Description

Before, the BRFSS based topics were suppressing the unknowns map and using a info alert to inform the user that the dataset contained no unknown demographics, since the `pct_share` unknown rows contained `undefined` values. The CAWP dataset contains `0 pct_share` for unknowns rows, and was therefor displaying a full map of zeros and using the "warning" alert.

- this PR lets the unknownmap card handle cases when the amount of unknown share is specifically ZERO
- also adjusts severity in the normal RateInfo box to be "info" if the unknown rate is ZERO since that is a GOOD thing from an equity perspective
- clarifies SELF vs CHILD geo level in the unknowns map card.... a lot of work could be done here with variable ad component naming to make it more clear

## How has this been tested?

manually

Screen Shot 2022-11-16 at 11 37 47 AM
Screen Shot 2022-11-16 at 11 38 10 AM

## Types of changes

### DIFF
--- a/frontend/src/cards/UnknownsMapCard.tsx
+++ b/frontend/src/cards/UnknownsMapCard.tsx
@@ -151,7 +151,7 @@ function UnknownsMapCardWithKey(props: UnknownsMapCardProps) {
             (unknown: Row) => unknown[metricConfig.metricId] === undefined
           );
 
-        // when suppressing states with too low COVID numbers
+        // for data sets where some geos might contain `0` for every unknown pct_share, like CAWP US Congress National
         const unknownsAllZero =
           unknowns.length > 0 &&
           unknowns.every(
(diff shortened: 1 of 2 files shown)


eriwarr (approved):
👍🏿

joshzarrabi:
I think this is great, the difference between 0% unknown and zero unknown values is probably something we can describe better here. 
Screen Shot 2022-11-16 at 2 19 59 PM

joshzarrabi (approved):

PR #1905, opened 2022-11-21
Make `zero_out_pct_rel_inequity` reusable

benhammondmusic:
## Description

- moves fn into `dataset_utils`
- accepts a map of `rate_cols: pct_inequity_cols` to allow for the zeroing of multiple conditions (like covid deaths, hosps, cases) while also allowing for conditions where the "rate" is not a 100k (like CAWP conditions)
- accepts an option `pop_pct_col` which will allow the fn to preserve null `pct_rel_inequity` on rows where we don't have population info (e.g. the Middle Eastern & North African race group for CAWP where we can show a RATE but not a PCT_REL_INEQUITY because we don't have population data)

## Motivation and Context

Allows reusable functions to be shared by COVID over time and #1851 and any other upcoming time series topics

## How has this been tested?

Adds unit testing for this fn that covers cases where:
- there is a state/year where rates are >0 so pct_ineq isn't altered
- there is a state/year where rates are 0 or none, so pct_ineq are (initially) set to 0
- one of those races does not contain population data, so the pct_ineq is reverted back to null
- there is a year where one state has no rates and another state has rates; the "having rates" state doesn't get affected

## Types of changes


joshzarrabi (approved):
Looks great, thank you for this

joshzarrabi:
Confirmed that this does not change the output of the pipelines
Screen Shot 2022-11-22 at 11 32 41 AM

PR #1911, opened 2022-11-29
Removes `fullCardTitleName` from Metric Config

### LINKED ISSUE #1858: Remove/Rename deprecated `fullCardTitleName` in metricConfig
**Describe the bug**

We no longer have metric-level information in our actual card titles, only generic cues as to what the card will be doing. We moved all relevant metric detail into the `chartTitle` metric, but the `fullCardTitleName` is still kicking around because it was used in other places that needed that metric description like the downloadable file names and the missing data alerts. Having this excess, stale text makes it harder to tell which string appear where, and harder to maintain

**Expected behavior**

- remove all `fullCardTitleName` from the metric config
- use the `chartTitle` anywhere that full metric-level description is needed. some amount of string tweaking might be needed for this. If we truly need different strings for different things we should create new, clearly labeled options on the metricConfig object, but i think in general the chartTitle should be fine for everything
- we no longer have a metric-level, location specific title appearing on our tables (the DataTable nor the AgeAdjustedTable); to be consistent and to account for the likelihood of screenshots, we should attempt to add these chartTitle strings as captions/titles for the table. This makes it more accessible and more clearly labeled what is exactly being shown in the table

eriwarr:
## Preview Link

<a href="https://deploy-preview-1911--health-equity-tracker.netlify.app/exploredata" target="_blank">
  <br />
Click to demo updates </a> 

## Description

This PR:
- removes  the fullCardTitleName from the Metric Config, 
- renames some of the remaining ones to display the column titles on the Data Table. 

## Motivation and Context

- Fixes #1858

## How has this been tested?

Tested locally

## Types of changes


benhammondmusic (commented):
awesome Eric; left some small bugs i found (most of which were already there before this PR lol) but if you don't mind fixing them up as part of this and then we'll move on to the next phase of untangling this metric config

benhammondmusic:
Did you forget to push your changes maybe? I still see the "with unknown" bug and uppercased "According" etc

benhammondmusic (approved):
Nice

Issue #1662, opened 2022-07-11
Data Typing - Pandas, BigQuery, and JS


benhammondmusic:
## Tasks:
- [x] Confirm above assumptions / loose understandings
- [x] Confirm outlined plan is the simplest and most accurate
- [x] Created backend util fn that automatically generates the needed BigQuery dtypes

Refactor each data source, run pipelines, confirm frontend working:
- [x] cawp, svi, bjs, vera in #1912 
- [ ] covid
- [ ] covid vaxx all levels
- [ ] poverty / health insurance
- [ ] AHR
- [x] ACS / ACS2010
- [ ] Census Pop. Est. / Census Pop. Est. SC

## Background
Coordinating data types between Python, Pandas, BiqQuery, and JavaScript has been problematic. A lot of time is wasted in pipeline testing that can't be done locally to ensure correct types for all data

## What's Missing / Problems with the current approach
We would like to send 4 different types from our Python/Pandas dataframes to BiqQuery:
- floats (we would like our `_pct_share` metrics to include a single decimal value, e.g. `40.4%`
- ints (we would like our `_per_100k` rates to display as whole numbers, e.g. `345 per 100k`
- strings (all text, group labels, etc)
- boolean (used only for `race_includes_hispanic` but maybe this will be going away after #1474 is done

There are a few issues with our current approach due to BigQuery  and Panda quirks:
- Pandas doesn't like nullable `int` columns (meaning a column contains a mix of int values and `null` values). We can get around this using the newer `dtype='Int64'` in Pandas
- BigQuery seems to accept these `INT64` data just fine, and the BigQuery ui displays them being stored, however when we fetch them on the frontend it seems to cast BigQuery `INTEGER` (an alias for `INT64` I believe) to JS `string`. Confusingly, it seems to be casting BigQuery `FLOAT` as JS `number` which is exactly what we want. 
- The frontend has a section in `DataFetcher` dedicated to testing whether a cell value in a nullable column is `== null` and if so, explicitly setting it to `null`. Otherwise, it is casting to `Number()`. 

## Solution

if we set everything on Pandas writing to BQ to `FLOAT` when we import into the frontend it will handle all of the above issues. I believe we already have frontend code in place to explicitly include `.0` as needed for whole number `_pct_share` values, and it seems JS auto trims dangling `.0` by default, so we don't need to further cast these "`float`" values as whole numbers

## Screenshots / Mockups

[...]

PR #1913, opened 2022-11-30
Updates ACS to make a new combo race `AIAN_API` for CAWP comparisons

benhammondmusic:
## Description

Sums together 3 ACS race groupings (to be used in #1851 as a comparison population)
ACS: `ASIAN` + `AIAN` + `NHPI` = `AIAN_API`

Updates `/data/acs_2010` json with manually calculated population totals for the combined race group based on the ACS_2010 tables

## Motivation and Context

#1851 so we can include some of CAWP's non-standard race groups (combined only) on the comparison bar chart and inequities over time charts: 
- `Asian American & Pacific Islander` + `Native American, Alaska Native, & Native Hawaiian` = `AIAN_API`

#1662 updates data types for ACS and ACS_2010

## How has this been tested?

- ACS tests updated to include new race grouping

- Running the DAGs on my GCP produces the expected rows in my Big Query:
National
Screen Shot 2022-11-30 at 5 20 23 PM

State + PR + DC
Screen Shot 2022-11-30 at 3 49 12 PM

2010 Territories GU AS MP VI
Screen Shot 2022-11-30 at 5 17 06 PM

## Types of changes


joshzarrabi (approved):
Looks good, thanks for this

PR #1919, opened 2022-12-02
Extricate non-CawpTime stuff from CawpTime frontend PR

### LINKED ISSUE #1825: "Sums over 100%" Alert needs proper margin
**Describe the bug**

fix margin in >100% alert  (i think i`<Alert>` needs to be inside `<CardContent>` or similar) 

**To Reproduce**
1. Go to "poverty" by hispanic-inclusive race and view the population vs share card's alert that's beneath the bar chart

**Expected behavior**

The alerts on the cards need a margin and not to extend to the edge of the card
**Screenshots**

Screen Shot 2022-09-27 at 4 29 15 PM

benhammondmusic:
## Preview Link

<a href="https://deploy-preview-1919--health-equity-tracker.netlify.app/exploredata" target="_blank">
  <br />
Click to demo updates </a> 

## Description

To keep the #1894 PR focused this PR updates some of the changes that I've made in process but that aren't specific to the new CAWP time. 

- metricQuery must now send a `VariableId` which is helpful since we often need to fetch / shape data differently based on the current data type selection.
- improves Incarceration fetching using above VariableId change
- use CawpOverlappingRacesAlert with link to legislation pushing for MENA on the US Census
- stronger typing
- optional chaining to prevent possible undefined errors
- some text content updates
- updates datasource metadata to include date ranges on our time series data
- fixes #1825 

## Motivation and Context

Keep PRs smaller and easier to understand / roll back

## How has this been tested?

frontend tests still passing
- [x] visually inspected several different pages

Screen Shot 2022-12-02 at 1 01 40 PM
Screen Shot 2022-12-02 at 12 52 50 PM

## Types of changes


frontend/src/cards/TableCard.tsx
```diff
@@ -66,7 +66,7 @@ export function TableCard(props: TableCardProps) {
     )
   );
 
-  let metricConfigs: Record<string, MetricConfig> = {};
+  let metricConfigs: Partial<Record<MetricId, MetricConfig>> = {};
```
eriwarr, line 69: Nice! I'm just learning about the Partial type property.

frontend/src/data/variables/IncarcerationProvider.tsx
```diff
@@ -94,17 +97,9 @@ class IncarcerationProvider extends VariableProvider {
...
+    let variableId: VariableId | undefined = metricQuery?.variableId;
 
-    if (breakdowns.geography === "county") {
-      if (metricQuery.metricIds.some((id) => JAIL_METRICS.includes(id)))
-        dataType = "jail";
-      else if (metricQuery.metricIds.some((id) => PRISON_METRICS.includes(id)))
-        dataType = "prison";
-    }
-
-    const datasetId = this.getDatasetId(breakdowns, dataType);
```
eriwarr, line 107: So instead of using this calculation to determine which dataset to retrieve, we're sending the variable id with the data set needed? Nice!!

eriwarr (approved):
Awesome refactor and updates! Great job! 🚀

frontend/src/cards/TableCard.tsx
benhammondmusic: Ya it's pretty slick; otherwise TypeScript Record requires you to map every possible item; the Partial lets you just define some of them

frontend/src/data/variables/IncarcerationProvider.tsx
benhammondmusic: Thanks! Ya, all the original sources we were bundling all the datatype info into the same table (like with covid there are columns for hosp, for deaths and for cases). But now with Incarceration at the county level and for the CAWP over time stuff we need to store the datatypes in separate tables for various reasons. So i figured rather than do a ton of work to guess at what the calling VariableID was, we should just ship it along with the request :)

joshzarrabi (approved):

PR #1923, opened 2022-12-05
Removes sticky header property from table header

### LINKED ISSUE #1917: MUI Table headers have higher z-index than sticky madlib
**Describe the bug**

The sticky madlib header should be "on top" of everything other than modals like the small multiple window. However, as you scroll down the page, the header row of our MUI Table components (used in the `<TableCard>` and the `<AgeAdjustedTableCard>`) actually float over the top of the sticky madlib header

**To Reproduce**
0. use a wide screen desktop view
1. Go to https://healthequitytracker.org/exploredata?mls=1.covid-3.00&mlp=disparity&demo=age&dt1=deaths#data-table
2. slowly scroll down and observe the table header going on top of the madlib header
3. keep going and observe on the age-adjusted card as well

**Expected behavior**

The table card header should go "under" the madlib header

This may require finding out what z-index MUI is assigning the tablecard header and then bumping up everything else that should be "on top" accordingly like the madlib header, the small multiple modal, etc. Might be tricky given that our components are set using our defined z values in the sass variables, but some of these MUI and other 3rd party components set their own z-index

**Screenshots**

https://user-images.githubusercontent.com/41567007/205212847-16305247-87c7-49a0-8e48-7d1f00e98d0a.mov

Screen Shot 2022-12-01 at 9 04 55 PM

 - Device: MacOS Chrome

eriwarr:
## Preview Link

<a href="https://deploy-preview-1923--health-equity-tracker.netlify.app/exploredata?mls=1.covid-3.00#data-table" target="_blank">
  <br />
Click to demo updates </a> 

## Description

- The PR removes the `stickyHeader` property from the Age-Adjusted Table Chart & Table Chart. 
- The stickyHeader` property is not needed as the headers to the tables should be at the same z-index as the other properties of the table. 
- Adjusts the Madlib z-index to be 3 instead of 1, which puts it on top of the remaining MUI table headers (auto-assigned to 2) that are present on the long complex altData tables below the time charts

## Motivation and Context

- Fixes #1917

## How has this been tested?

- Tested locally

Screen Shot 2022-12-05 at 4 01 08 PM

## Types of changes


benhammondmusic (approved):
awesome, thanks for the quick turnaround! I added a little extra code to allow the altData under the time charts to have sticky headers which is helpful, but keeps them stacked in the correct vertical order

Issue #1926, opened 2022-12-05
Age Adjust over time


joshzarrabi:
## Is your feature request related to a problem? Please describe.
Currently our COVID death and hospitalization age adjusted ratios are not age adjusted. This is a problem, because the crude rates can lead people to incorrect assumptions about who is dying.

## Describe the solution you'd like
For right now, we should just do the age-adjustment on the backend, and let users download it, we can think about the best way to show the users the data later.

benhammondmusic:
**Backend half is done; only the display question remains.** Re-scoping rather than closing.

The opening claim, "our COVID death and hospitalization age adjusted ratios are not age adjusted," is no longer true. `python/datasources/age_adjust_cdc_restricted.py` computes both:

```python
column_types[std_col.COVID_HOSP_RATIO_AGE_ADJUSTED] = "FLOAT"
column_types[std_col.COVID_DEATH_RATIO_AGE_ADJUSTED] = "FLOAT"
```

There is a parallel `age_adjust_cdc_hiv.py`, and both are registered in `data_sources.py`. So the issue's own stated near-term goal ("for right now, we should just do the age-adjustment on the backend, and let users download it") is met.

What is left is the single unchecked box: **figure out how to show this to the user.** That is a real and non-trivial question, since an age-adjusted ratio is easy to misread. It is also the piece with the most equity impact, because crude rates are exactly what "can lead people to incorrect assumptions about who is dying."

The title "Age Adjust over time" is now misleading. Suggest retitling to something like "Surface age-adjusted ratios in the UI" and treating time-series age adjustment as a separate question if we still want it.

Added to the new "Data Accuracy & Freshness" milestone.

PR #1931, opened 2022-12-06
Age Adjusts COVID deaths and hospitalizations over time 

joshzarrabi:
## Description

Runs the age adjustment algorithm on time series COVID hospitalization and death data at the national and state level.

In order to accomplish this I had to: 

* Break down the `by_race_age_state` data generated by `cdc_restricted_local.py` by month
* Significantly refactor the age adjustment code to make it fast enough to run on time series data in a reasonable amount of time.

Note: These are only the backend changes to the code, we still need to figure out a way to let the users access this data.

## Motivation and Context

Backend portion of #1926

## How has this been tested?

Yes, I wrote tests, and verified that the refactor didn't change any of the existing numbers. Potentially more validation should be done on the time series numbers. 

Screen Shot 2022-12-09 at 11 20 50 AM

## Types of changes


python/datasources/age_adjust_cdc_restricted.py
```diff
@@ -116,146 +146,123 @@ def merge_age_adjusted(df, age_adjusted_df):
...
     age_adjusted_df = age_adjusted_df.reset_index(drop=True)
 
     return pd.merge(df, age_adjusted_df, how='left', on=merge_cols)
 
 
-def get_expected_hosps(race_and_age_df, population_df):
-    """Calculates the age adjusted expected hospitalizations of each racial group.
-       I made this function to break up the age adjustment into smaller, more easily testable pieces.
+def get_expected_col(race_and_age_df, population_df, expected_col, raw_number_col):
+    """Calculates the age adjusted expected deaths of each racial group.
```
benhammondmusic, line 159: this is multi-purpose now right? So should the doc string be updated to say like "deaths or hospitalizations" instead of just deaths?

benhammondmusic (approved):
looks great; excited to figure out how to integrate this data for the users

python/datasources/age_adjust_cdc_restricted.py
joshzarrabi: Yup, good catch, thanks

PR #1934, opened 2022-12-08
Replace `mobileChartTitle` with `chartTitleLines` 

eriwarr:
## Preview Link

<a href="https://deploy-preview-1934--health-equity-tracker.netlify.app/exploredata?mls=1.covid-3.00" target="_blank">
  <br />
Click to demo updates </a> 

## Description

This PR includes a small refactor to the MetricConfig.ts file and the files that use its variables. 
- [x] Remove `mobileChartTitle` from MetricConfig.ts and anywhere else it is used.
- [x] Replace `chartTitle` with `chartTitleLines[]` in MetricConfig.ts.
- [x] Update breakdown and location strings used by the `useCreateChartTitle` function. 
- [x] Refactor the `useCreateChartTitle` function to use new variables.
- [x] Add test file for useCreateChartTitle

## How has this been tested?

- Tested locally

## Types of changes


frontend/src/cards/UnknownsMapCard.tsx
```diff
@@ -58,11 +58,9 @@ function UnknownsMapCardWithKey(props: UnknownsMapCardProps) {
   const preloadHeight = useGuessPreloadHeight([700, 1000]);
   const metricConfig = props.variableConfig.metrics["pct_share"];
   const currentBreakdown = props.currentBreakdown;
-  const breakdownString =
-    BREAKDOWN_VAR_DISPLAY_NAMES_LOWER_CASE[currentBreakdown];
-
+  const breakdownString = `with unknown ${BREAKDOWN_VAR_DISPLAY_NAMES_LOWER_CASE[currentBreakdown]}`;
```
benhammondmusic, line 61: this creates a bug `We do not currently have Share of all voter participation **broken down by** with unknown`

 (i think it's from this line)... 

Screen Shot 2022-12-09 at 11 37 34 AM

benhammondmusic:
@eriwarr 
there's a bug `We do not currently have Share of all voter participation **broken down by** with unknown`

i think you need to suppress the "broken down by" part on the unknowns card or otherwise switch up the string combinations

Screen Shot 2022-12-09 at 11 37 34 AM

frontend/src/cards/AgeAdjustedTableCard.tsx
```diff
@@ -92,7 +92,7 @@ export function AgeAdjustedTableCard(props: AgeAdjustedTableCardProps) {
     config.metricId.includes("ratio")
   );
 
-  const locationName = props.fips.getSentenceDisplayName();
+  const locationName = `in ${props.fips.getSentenceDisplayName()}`;
```
benhammondmusic, line 95: very minor thing, but for naming a variable like this, I like to use something that indicates the preposition "in" is coming along... using `locationName` implies to me it's only the name. I might use something like `locationPhrase` or something like that so it's clear there is more than just the name itself included in the variable string

frontend/src/cards/DisparityBarChartCard.tsx
```diff
@@ -82,7 +82,9 @@ function DisparityBarChartCardWithKey(props: DisparityBarChartCardProps) {
     locationName
   );
 
-  const filename = `${metricConfig.populationComparisonMetric?.chartTitle}${locationName}`;
+  const filename = `${metricConfig.populationComparisonMetric?.chartTitleLines.join(
```
benhammondmusic, line 85: it seems like we shouldn't have to duplicate effort making this joined filename and making the joined titles elsewhere.... can we just make it once when the component loads and then use that "chartTitleOneLine" string as the filename, the missing data text, and wherever else a one-liner version of the title is needed?

frontend/src/cards/DisparityBarChartCard.tsx
```diff
@@ -137,7 +139,7 @@ function DisparityBarChartCardWithKey(props: DisparityBarChartCardProps) {
             ) : (
               <CardContent>
                 <MissingDataAlert
-                  dataName={metricConfig.chartTitle || metricConfig.shortLabel}
+                  dataName={metricConfig.chartTitleLines.join(" ")}
```
benhammondmusic, line 142: like here, couldn't you use "chartTitleOneLine" or whatever you had created above for the filename ?

benhammondmusic (approved):
Nice work thanks!

Issue #1936, opened 2022-12-09
Remove unused races cols


joshzarrabi:
## Is your feature request related to a problem? Please describe.
Currently, we generate four different race columns on the backend, and send them to the frontend. 

These columns are:
* `race_category_id`
* `race_and_ethnicity`
* `race_includes_hispanic`
* `race`

Out of these four, we only ever use two, at any point, which are:
* `race_category_id`
* `race_and_ethnicity`

We use `race_category_id` as our main unique identifier that we use for all of our backend and remaining frontend logic. This column needs to stay. For example, it is what we use to standardize population from the ACS with COVID information.

We use `race_and_ethnicity` as the display name for the races on the frontend. This is fine, but causes some problems:
* It's pointless to ship this around on the backend, as it isn't used anywhere there.
* It makes changing the display name of a racial group relatively cumbersome because, in order to do so, we have to change the backend code, cut a release, and then re-run all pipelines where that race appears. 

The other two fields are not used anywhere, and they are a burden to deal with, as we have to do lots of [special casing in the code for them](https://github.com/SatcherInstitute/health-equity-tracker/blob/1af4818605f0d45cb001feb6d07effda01e9cbbe/python/datasources/cdc_restricted.py#L139-L140), and they inflate the file size we send to the user for no reason. 

UPDATE:
The solution below has proven more difficult than anticipated. Having a shared file between the `python` and `frontend` directories is clunkier than I expected and will probably require some confusing trickery that is probably worse than the problem we are trying to solve. 

Rather, I will simply remove the two unused columns (`race_includes_hispanic` and `race`) from the backend.

----------------------------------------------------
OUTDATED SOLUTION

## Describe the solution you'd like
I propose that we should solve this in two parts:
1. Cut a frontend release that ignores all columns except for `race_category_id`, and uses some sort of mapping function to figure out the display name.

2. Cut a backend release that removes the following columns from everywhere in the python code:
* `race_and_ethnicity`
* `race_includes_hispanic`
* `race`

NOTE: This will cause a change in behavior that we have to think about going forward
Right now, when we want to add a *new* race, we don't have to do anything on the frontend, we just send it via the API and the frontend will know what it is called because of the `race_and_ethnicity` field. This will no longer be true, as we will have to add that race to the frontend map function we make as well. 

[...]

Issue #1683, opened 2022-07-15
Dataset IDs should be strongly typed as a string union


benhammondmusic:
**Is your feature request related to a problem? Please describe.**
Right now they are strings. We have a test to ensure match between the map and the list, but still no way to ensure every dataset id used in a variable provider is present on the list/map

**Describe the solution you'd like**
- Add a type that is a union of the `id`s defined in `frontend/src/data/config/DatasetMetadata.ts`
- Type the datasetid in all providers as this new type, and it'll fail if the list hasn't been updated

**Additional context**

This will prevent these ongoing errors where we mismatch ids or forget to update them entirely

PR #1942, opened 2022-12-19
Refactor `getDatasetId` to respect abstractions

joshzarrabi:
## Preview Link

<a href="https://deploy-preview-1942--health-equity-tracker.netlify.app/exploredata" target="_blank">
  <br />
Click to demo updates </a> 

## Description

The `Breakdowns` class provides abstract functions to tell its users what kind demographic breakdown (race / sex /age) it is, and we should use these in our `getDatasetId` functions rather than just taking the demographic breakdown name and placing it into the api request string.

Also throw a proper error if the dataset does not exist. 

## Motivation and Context

Originally done for #1936. Not really necessary anymore but still a nice to have.

## How has this been tested?

Tested locally.

## Types of changes

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -5,7 +5,23 @@
   "overrides": {
     "react-refresh": "0.11.0"
   },
+  "devDependencies": {
+    "@playwright/test": "^1.24.2",
+    "@storybook/addon-actions": "^6.5.9",
+    "@storybook/addon-essentials": "^6.5.9",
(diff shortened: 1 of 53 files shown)


benhammondmusic (approved):
this is great and i think is a step towards #1683 which would be really great and prevent all sorts issues that arise from mismatched datasetids and metadata

joshzarrabi:
Originally done for #1936.

Issue #1805, opened 2022-09-21
Add sanity checks to the data pipelines


joshzarrabi:
## Background

As engineers developing features, we would like to have confidence that the data looks something like we would expect it to. 
There are basic things we look for in the data after running updates, and we can automate some of those checks.

## What's Missing / Problems with the current approach
Recently. as described in #1786 we ran into an issue where some `pct_share` metrics were both:
* Impossible (added up to well over 100%)
* Not the same as what was generated in the dev pipeline.

These are both things that could have been caught in a basic pipeline check.

## Proposed Solution
We should steps the Airflow pipeline, right before the `exporter` step, that run basic sanity checks on the data. If any of these steps fail the pipeline should fail and the API will not be updated.

## End State 
This epic will be done when we have tasks in very pipeline that ensure the data does not have any glaring errors. 

## How will this move the needle on health equity?
This will help us avoid any obvious errors in the data that will make our users potentially lose confidence in the tracker.

## Tasks:
- [ ] Linked sub-issues #1234
- [ ]

benhammondmusic:
**Epic has no remaining defined work.** Flagging for closure or re-scope.

Both tasks listed in the body are closed:

- #1806 Verify that the output of each `prod` pipeline is exactly the same as the `dev` pipeline. **CLOSED**
- #1807 Verify that `pct_share` metrics for every COVID geo/demo/time breakdown adds up to exactly 100%. **CLOSED**

The stated end state ("we have tasks in every pipeline that ensure the data does not have any glaring errors") was never actually reached, so closing this outright would lose the intent. But as written the epic tracks nothing.

Two options:

1. **Close it** and let the new "Pipeline Trust & Automation" milestone carry the goal forward. That milestone already groups #3322 (test DAG for dependency verification), #1796 (scheduled data-refresh reminder issues), and #4832 (Gen 3 fixture convention).
2. **Re-scope it** into a concrete issue: add a validation step to the DAG workflows immediately before the exporter, failing the run on impossible values. The original #1786 incident (pct_share summing well over 100%) is still the best test case, and the checks would generalize: pct_share sums, rate bounds, null denominators, row-count deltas against the prior run.

I have added this to the "Pipeline Trust & Automation" milestone for now so it does not get lost either way.

PR #1943, opened 2022-12-20
Pct share sanity check

### LINKED ISSUE #1806: Verify that the output of each `prod` pipeline is exactly the same as the `dev` pipeline.

Recently, we had a strange situation where the output of our prod pipeline was not exactly the same as the output of the dev pipeline. This should never happen and if it does we should fail to update the API and investigate. 

We should add a new Airflow job that runs right before the `exporter` step on each `prod` pipeline that ensures that the updated prod big query tables are exactly the same as the dev big query tables.

We can utilize the logic in the https://github.com/SatcherInstitute/health-equity-tracker/blob/main/e2e_tests/scripts/ensure_datasets_equal.py script to do this.

## Additional context

An important thing to note is this is the first time we will have a diversion in the logic between the dev and prod pipelines. In my opinion we should handle this in the step itself. so the actual airflow pipelines will be the same, but the `verify` step, or whatever we choose to call the output of this, will just always pass on the `dev` pipeline.

eriwarr:
## Description

This PR includes a fix to ensure that the sum of percentage share values for a particular demographic, location and time on a given data frame are within the expected range (99-101), as described in issue #1805.

- Installed a new python module to contain the sanity checks. 
- Added the sanity check to the dag directory. 
- Note: Airflow could not import the sanity check module when running the Airflow server, so I included the logic in the Airflow directory. There is likely a way to use the Python Virtual Environment Operator so that the `sanity_check.py` file is not duplicated, but I will revisit it. 

## Motivation and Context

- Fixes #1806

## How has this been tested?

- Ran pytests on `tests/sanity_checks`.
- Ran tests on Airflow server. 

Airflow error example
airflow_error

Succesful run
succesful_run

## Types of changes


python/tests/sanity_checks/test_sanity_check.py
```diff
@@ -0,0 +1,51 @@
...
+test_dtype = {
+    'county_fips': str,
+    'state_fips': str,
+}
+
+
+def testGenerateCountyDatasetAge():
+    df = pd.read_json(CDC_RESTRICTED['age_county'], dtype=test_dtype)
+    result = check_pct_values(df)
+    return result
```
joshzarrabi, line 26: I think these should be `assert`s. You can do it all in one line: like `assert check_pct_values(df)`

python/tests/sanity_checks/test_sanity_check.py
```diff
@@ -0,0 +1,51 @@
...
+TEST_DIR = os.path.join(THIS_DIR, os.pardir, "data", "sanity_checks")
+
+CDC_RESTRICTED = {
+    'age_county': os.path.join(TEST_DIR, 'cdc_restricted_age_county.json'),
+    'sex_county_time': os.path.join(TEST_DIR, 'cdc_restricted_sex_county_time.json'),
+    'sex_county': os.path.join(TEST_DIR, 'cdc_restricted_sex_county.json'),
+    'sex_national': os.path.join(TEST_DIR, 'cdc_restricted_sex_national.json'),
+    'sex_state': os.path.join(TEST_DIR, 'cdc_restricted_sex_state.json')
+}
+
```
joshzarrabi, line 16: I think in this file you should also test an error case. Like a case where you expect it to fail and error. See https://github.com/SatcherInstitute/health-equity-tracker/blob/89f6457cf5f1a4dcef520cfd349ce73a0de61025/python/tests/ingestion/test_dataset_utils.py#L278-L281 as an example of how to do that.

python/tests/sanity_checks/test_sanity_check.py
```diff
@@ -0,0 +1,51 @@
...
+
+def testGenerateCountyDatasetAge():
+    df = pd.read_json(CDC_RESTRICTED['age_county'], dtype=test_dtype)
+    result = check_pct_values(df)
+    return result
+
+
+def testGenerateCountyDatasetSexTime():
+    df = pd.read_json(CDC_RESTRICTED['sex_county_time'], dtype=test_dtype)
```
joshzarrabi, line 30: Nice job testing time

python/sanity_checks/build/lib/sanity_checks/sanity_check.py
```diff
@@ -0,0 +1,53 @@
+import pandas as pd
```
joshzarrabi, line 1: I dont think this file should be checked in at all.

python/sanity_checks/sanity_check.py
```diff
@@ -0,0 +1,53 @@
+import pandas as pd
+import ingestion.standardized_columns as std_col
+
+
+def generate_cols(df: pd.DataFrame):
```
joshzarrabi, line 5: Do we use this file? It looks like you copied the code to the `airflow/dags/sanity_check.py` file and use that in the pipeline. We should use the code that we run our tests on.

joshzarrabi (changes requested):

benhammondmusic (approved):
looks good to me; i've started doing screenshots of bigquery / airflow / backend stuff for my backend PRs as I think it can be helpful when reviewing and referencing exactly what these backend changes are doing. Maybe consider taking a few screenshots of the printout from a successful run and a run where things aren't adding up properly

airflow/dags/util.py
```diff
@@ -112,6 +116,22 @@ def service_request(url: str, data: dict, **kwargs):
         raise Exception('Failed response code: {}'.format(err))
 
 
+def sanity_check_request(dataset_id: str):
+    bq_client = bigquery.Client()
+
+    tables = bq_client.list_tables(dataset_id)
+    for table in tables:
+        table_name = "{}.{}.{}".format(
```
joshzarrabi, line 124: Would you mind switching this to an `fstring`

airflow/dags/test_sanity_check.py
```diff
@@ -0,0 +1,49 @@
+import os
```
joshzarrabi, line 1: This looks good but I think you should put it in a `test` directory

joshzarrabi (approved):
Thank you Eric!

eriwarr:
> looks good to me; i've started doing screenshots of bigquery / airflow / backend stuff for my backend PRs as I think it can be helpful when reviewing and referencing exactly what these backend changes are doing. Maybe consider taking a few screenshots of the printout from a successful run and a run where things aren't adding up properly
> looks good to me; i've started doing screenshots of bigquery / airflow / backend stuff for my backend PRs as I think it can be helpful when reviewing and referencing exactly what these backend changes are doing. Maybe consider taking a few screenshots of the printout from a successful run and a run where things aren't adding up properly

Great point! I'll update the PR to include them.

PR #1944, opened 2022-12-21
Remove unused race cols

### LINKED ISSUE #1936: Remove unused races cols

joshzarrabi:
## Description

Removes the unused race columns
* `race`
* `race_includes_ethnicity`

## Motivation and Context

Closes #1936

## How has this been tested?

Yes, I ran all of the pipelines and tested them locally.

## Types of changes


python/datasources/census_pop_estimates.py
```diff
@@ -56,13 +56,10 @@ def write_to_bq(self, dataset, gcs_bucket, **attrs):
 
         state_df = generate_state_pop_data(df)
 
-        column_types = {c: 'STRING' for c in state_df.columns}
-
-        if std_col.RACE_INCLUDES_HISPANIC_COL in df.columns:
-            column_types[std_col.RACE_INCLUDES_HISPANIC_COL] = 'BOOL'
+        col_types = gcs_to_bq_util.get_bq_column_types(state_df, [])
```
benhammondmusic, line 59: would it be better for us to make the `float_cols` an optional kwarg so we don't have to send in an empty list for datasets that won't produce any floats?

benhammondmusic (approved):
awesome!

python/datasources/census_pop_estimates.py
joshzarrabi: Hmm, I think in most cases we will want to pass in a list. Don't think its worth making it optional in this case.

PR #1950, opened 2023-01-04
Fixes CAWP state leg. backend missing edge case

benhammondmusic:
## Description

- The reduced test files didn't cover the weirdness found in the CAWP state leg. files including `*` in the year column and buggy mix of index/year columns. This PR extends the test files to include these weirdness edge cases and then fixes the processing code so that the tests continue to pass and the GCP dag doesn't break

- Fixes issue where territory state-leg tables in `/data` were being loaded in with the years as `int` instead of `string`, causing `null` counts where there should have been some (for example, USVI in 2022 should have 15 total state leg, but `null` was showing on BigQuery

- adds missing USVI leg. totals to `/data` back from 2007 to 1950s; we still restrict state-level legislator figures to 1983 onwards for now based on CAWP state limitations

## Motivation and Context

Getting #1939 to be able to read from the backend files and not /tmp files

## How has this been tested?

tests still pass

## Screenshots

Screen Shot 2023-01-04 at 9 57 29 PM
Screen Shot 2023-01-04 at 9 58 20 PM

## Types of changes


joshzarrabi (approved):

python/datasources/cawp_time.py
```diff
@@ -555,20 +555,25 @@ def get_state_leg_totals_df():
     state_dfs = []
     for fips, id in FIPS_TO_STATE_TABLE_MAP.items():
         state_df = gcs_to_bq_util.load_csv_as_df_from_web(get_stleg_url(id))
+
+        # remove weird chars from col headers
         state_df.columns = state_df.columns.str.replace(r'\W', '', regex=True)
 
+        # remove non-digits from buggy index/years column
+        state_df["10"] = state_df["10"].astype(str).replace(
```
joshzarrabi, line 563: Why is the col name "10"?

joshzarrabi (approved):

eriwarr (approved):

python/datasources/cawp_time.py
benhammondmusic: That's the buggy index/year column situation. This is what the source table looks like:
Screen Shot 2023-01-04 at 2 57 47 PM

Linked from the state leg. table on this page (and some of the other states) https://cawp.rutgers.edu/facts/state-state-information/arizona

joshzarrabi: Cool, would you mind throwing a comment on there describing this?

Issue #1956, opened 2023-01-09
Remove abandoned `_std` to csv functionality


benhammondmusic:
## Background
It looks like they were planning to have it so that if there was a _std on the id then it would duplicate the bq export as a .csv file, and then presumably you could download the csv directly on the frontend. See #1804 

## What's Missing / Problems with the current approach
that last part was never implemented. So, right now, it works to just have the frontend download the dataset as normal and then convert the dataforge df into a csv file for download. To do “direct” csv downloads, we either have to make the GCP bucket like public access (maybe?), or we’d have to download as a blob and basically rebuild the csv like we are already doing, just instead we rebuild csv into csv rather than currently json into csv.

## Proposed Solution
it’d really simplify the code to remove the entire _std concept, and just always store json on GCP, and then the frontend always downloads the json into dataforge, and then uses the existing conversion to csv for the data downloads page.

## End State 

- the frontend will only fetch json files, none of which have the `_std` suffix
- our BQ and S3 buckets will only contain json files, none of which have the `_std` suffix
- our backend will not use the `_std` on any generated BQ tables
- our exporter will not include a step for copying _std into csv files

## How will this move the needle on health equity?

Simplify codebase allowing devs to move faster and more easily understand the flow of data

## Tasks in order:
- [x] #1959
- [x] #1960 
- [x] #1961 
- [x] manually delete all .csv files and all _std json files from GCP; same for BQ tables

PR #1959, opened 2023-01-11
BACKEND: Removes _std to csv exporter step

benhammondmusic:
## Description

- exporter no longer writes a second time as CSV for BQ tables that end in `_std`
- updates tests

## Motivation and Context

Step 1 of #1956 

## How has this been tested?

updated tests passing

## Types of changes


eriwarr (approved):
👍🏿

PR #1971, opened 2023-01-23
properly merge age_adjusted time series files

joshzarrabi:
## Description

The age adjusted files were being merged incorrectly.
It didn't effect anything on the frontend because we dont use the time series age adjusted files yet but our sanity check pipeline caught it.

## Types of changes


benhammondmusic (approved):
nice

Issue #1487, opened 2022-02-24
Calculate all state and county metrics for ACS Health Insurance on the Backend


joshzarrabi:
## Background

This is in service of: https://github.com/SatcherInstitute/health-equity-tracker/issues/1474

Currently, in order to calculate the `per_100k` and `pct_share` metrics for state the state and county level, we do the following

1. Get the number of people with and without health insurance from the ACS API [using the acs_health_insurance.py](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/python/datasources/acs_health_insurance.py#L241-L247) script.

2. Generate a table of `by_race_age` and `by_sex_age` to big query with the following columns

* `with_health_insurance`
* `without_health_insurance`
* `total_health_insurance`

3. Do the necessary aggregations [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsHealthInsuranceProvider.ts#L68-L80). 

4. To generate the `per_100k` metric, we divide the `with_health_insurance` with the `without_health_insurance` [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsHealthInsuranceProvider.ts#L113-L116).

5. To generate the `pct_share` metric we do it on the [frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsHealthInsuranceProvider.ts#L124-L130)

6.  To generate the `population_pct` column, we use the `total_health_insurance` field [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsHealthInsuranceProvider.ts#L132-L139)

Note we did not use the normal acs population numbers for these. See: https://github.com/SatcherInstitute/health-equity-tracker/issues/669

## Problems

The code is unnecessarily complicated, hard to test, and asking the user to do lots of calculations and download way more data than they need. For example, the disaggregated by_race_age_county files are massive.

## Proposed Implementation

Calculate the following metrics on the backend

* `uninsered_per_100k`
* `uninsured_pct_share`

## End state

[...]

PR #1976, opened 2023-01-25
Health insurance backend

joshzarrabi:
## Description

* Moves all acs_health_insurance calculations to the backend
* Adds tests
* Creates new `_processed` files to put the calculations in.

## Motivation and Context

Backend portion of of #1487

## How has this been tested?

Yes, added tests and tested manually.

## Types of changes


python/ingestion/dataset_utils.py
```diff
@@ -202,7 +201,7 @@ def add_sum_of_rows(df, breakdown_col, value_col, new_row_breakdown_val,
        df: The DataFrame to calculate new rows from.
        breakdown_col: The name of the breakdown column that a new value is being
                       summed over.
-       value_col: The name of the column whose values should be summed.
+       value_col: The name or names of the column/s whose values should be summed.
```
benhammondmusic, line 204: can we specify the expected types here in the doc string since we're overloading this arg? just like, `the string name (or list of string names) of the column/s...` and i dont know enough about python typing but can you indicate str OR str[] in the function signature?

python/datasources/acs_health_insurance.py
```diff
@@ -126,490 +129,214 @@ def write_local_files_debug(self):
     # Returns:
     # FileDiff = If the data has changed by diffing the old run vs the new run.
     # (presumably to skip the write to bq step though not 100% sure as of writing this)
-    def upload_to_gcs(self, bucket):
+    def upload_to_gcs(self, bucket, **attrs):
```
benhammondmusic, line 132: does this method need any test coverage?

joshzarrabi: I don't think its worth it, its simple and would be hard to test

benhammondmusic: cool, ya i think the same for the acs_population one i'll skip it

python/tests/datasources/test_acs_health_insurance.py
```diff
@@ -0,0 +1,174 @@
...
+                  mock_fips: mock.MagicMock,
+                  mock_acs: mock.MagicMock,
+                  mock_json: mock.MagicMock):
+    acsHealthInsurance = AcsHealthInsurance()
+
+    acsHealthInsurance.write_to_bq('dataset', 'gcs_bucket')
+
+    assert mock_json.call_count == 1
+
+    # One call per race per geo, and then one call for sex at each geo
```
benhammondmusic, line 121: thanks for adding these comments in (and the explicit addition) i think it's really makes these assertions more useful because if/when i break it later changing code it's easier for me to understand why lol

benhammondmusic (approved):
this looks great josh. thanks for clearing up the code so much

python/tests/datasources/test_acs_health_insurance.py
joshzarrabi: Thats what tests are for!

Issue #1970, opened 2023-01-20
Add "HIV Diagnoses" Data to the tracker


joshzarrabi:
## Is your feature request related to a problem? Please describe.
As part of the Gilead HIV project, we need to start putting some HIV demographic data on the tracker.

I believe this fits in with the overall goal of the tracker as well, as it is well known that HIV disproportionately effects marginalized communities. 

From: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4937112/ 

> HIV remains a major public health concern due in part to significant disparities[1](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4937112/#B1) in rates of infection among racial/ethnic minority communities, with black/African American (hereinafter referred to as African American) and Hispanic/Latino populations being the most affected subgroups.

While we want to put many HIV related measures on the tracker, we have chosen to start with a single measure, HIV diagnoses, as a proof of concept and so we can get something on the tracker to show the Gilead funders.

## Did someone outside of the dev team ask for this? Who are they?
The Gilead project team (Donrie and Jareese).

## Describe the solution you'd like
A user should be able to select `HIV diagnoses` on the tracker, and see a visualization of the data found on this site: https://gis.cdc.gov/grasp/nchhstpatlas/maps.html at the national, state, and county level for race, sex, and age.

We will start with only the latest data, but should expand to time based data in the future. 

There may be a way to download all of the data easily, but for now you can download the data by making the selector look like: 

<img width="250" alt="Screen Shot 2023-01-20 at 1 14 46 PM" src="https://user-images.githubusercontent.com/3689026/213806622-bffe4b5e-a278-4d8c-b921-6d8db157f44d.png">

And then using the `csv` option from the download menu.

Then you should change the settings to download the other files you need for the county level and age/sex breakdowns.

I believe we should use the 2019 data as 2020, 2021, and 2022 all have some warning associated with them, but we should confirm with the gilead team.

PR #1977, opened 2023-01-25
HIV backend

eriwarr:
## Description

This PR adds the CDC HIV diagnoses dataset to the backend. 

- The data is stored by category in a demographic breakdown and the location (the ensures the population data is included). 
- The national dataset obtained from AtlasPlus contains no population info, so the population numbers in the final national data frame are merged from the state-level data. 
- All data was pulled manually, directly from CDC's AtlasPlus database. Unfortunately, their downloads/files do not have a link address, so any future downloads will have to be done manually until the CDC updates its data retrieval process. 
 

## Motivation and Context

- #1970

## How has this been tested?

- Tested locally and on airflow 

Screenshot 2023-02-07 at 2 51 20 PM

Screenshot 2023-02-07 at 2 53 15 PM

## Types of changes


python/datasources/cdc_hiv.py
```diff
@@ -0,0 +1,228 @@
...
+        missing_data = ['Data suppressed', 'Data not available']
+
+        # county fips needs 5 digits, state digits use 2
+        format_num = 5 if geo_level == COUNTY_LEVEL else 2
+        geo = COUNTY_LEVEL if geo_level == COUNTY_LEVEL else STATE_LEVEL
+
+        for group in GROUP_DICT[breakdown].keys():
+            source_group_df = gcs_to_bq_util.load_csv_as_df_from_data_dir(
+                f'cdc_hiv/{geo}', f'{breakdown}_{demo_dict.get(group, group)}_{geo}_{SOURCE_YEAR}.csv',
+                skiprows=9, thousands=',', dtype={'FIPS': str})
```
joshzarrabi, line 154: Also, would you mind putting comments explaining the `thousands` and `skiprows` params

python/datasources/cdc_hiv.py
```diff
@@ -0,0 +1,228 @@
...
+            # adds leading zeros to fips
+            source_group_df['FIPS'] = source_group_df['FIPS'].str.zfill(
+                format_num)
+            source_dfs.append(source_group_df)
+
+        if geo_level != NATIONAL_LEVEL:
+            alls_df = generate_alls_df(geo_level)
+            source_dfs.append(alls_df)
+
+        merged_df = pd.concat(source_dfs, axis=0)
```
joshzarrabi, line 164: I think `merged_df` is a confusing name. We should use something more semantic like `state_population_df` or something.

python/datasources/cdc_hiv.py
```diff
@@ -0,0 +1,228 @@
...
+        if geo_level == COUNTY_LEVEL:
+            df = merge_utils.merge_county_names(df)
+            df[std_col.STATE_FIPS_COL] = df[std_col.COUNTY_FIPS_COL].str.slice(
+                0, 2)
+
+        if geo_level == NATIONAL_LEVEL:
+            alls_df = generate_alls_df(NATIONAL_LEVEL)
+            nat_breakdown_df = generate_nat_from_state_df(breakdown)
+
+            merged_nat_df = pd.concat([nat_breakdown_df, alls_df])
```
joshzarrabi, line 177: Also don't use `merged` in the name here. It implies it was created with a `pd.merge` IMO

python/ingestion/gcs_to_bq_util.py
```diff
@@ -239,15 +239,16 @@ def load_csv_as_df_from_web(url, dtype=None, params=None, encoding=None):
     return pd.read_csv(url, dtype=dtype, encoding=encoding)
 
 
-def load_csv_as_df_from_data_dir(directory, filename, dtype=None):
+def load_csv_as_df_from_data_dir(directory, filename, dtype=None, skiprows=None, thousands=None):
```
joshzarrabi, line 242: Can you add `skiprows` and `thousands` to the docstring please

python/tests/datasources/test_cdc_hiv.py
```diff
@@ -0,0 +1,130 @@
+from unittest import mock
+from pandas._testing import assert_frame_equal
+from datasources.cdc_hiv import CDCHIVData
+import pandas as pd
+import os
+
+
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+TEST_DIR = os.path.join(THIS_DIR, os.pardir, "data",)
+GOLDEN_DIR = 'cdc_hiv/golden_data'
```
joshzarrabi, line 10: please use `os.path.join` here as well

joshzarrabi (changes requested):
Thanks Eric, this is great. 

I know it might be a little bit of extra work, but I think it would be worth it to shrink the test files a bit. For example, only keep one or two states and and one or two counties. I think it will make things easier to work with in the future.

python/datasources/cdc_hiv.py
```diff
@@ -0,0 +1,228 @@
...
+            missing_data), std_col.HIV_PER_100K] = np.nan
+
+        df[std_col.HIV_CASES] = df[std_col.HIV_CASES].replace(
+            ',', '', regex=True)
+        df.loc[df[std_col.HIV_CASES].isin(
+            missing_data), std_col.HIV_CASES] = np.nan
+        df[std_col.HIV_CASES] = df[std_col.HIV_CASES].astype(float)
+        df[std_col.HIV_PER_100K] = df[std_col.HIV_PER_100K].astype(float)
+
+        df[std_col.HIV_PER_100K] = (df[std_col.HIV_PER_100K]) * 1000
```
joshzarrabi, line 212: Also, We don't need to multiply here, it is already a per 100k

python/datasources/cdc_hiv.py
eriwarr: This line can also refer to county data when on a county level. I'll change it to `df` to avoid confusion.

python/datasources/cdc_hiv.py
```diff
@@ -0,0 +1,228 @@
...
+
+            # merge dataframes
+            df = merged_nat_df.merge(df, on=group_by_cols, how='left')
+            df.loc[df[breakdown] == 'All',
+                   std_col.POPULATION_COL] = df[std_col.POPULATION_COL].sum()
+
+        df.loc[df[std_col.HIV_PER_100K].isin(
+            missing_data), std_col.HIV_PER_100K] = np.nan
+
+        df[std_col.HIV_CASES] = df[std_col.HIV_CASES].replace(
```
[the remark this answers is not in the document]
eriwarr: Unfortunately, The `hiv_cases` column has strings that are not formatted as a floating point number, i.e., 'Data Supressed,' so the thousands parameter skips the entire column. Currently, it only works on the population columns.

python/tests/datasources/test_cdc_hiv.py
```diff
@@ -0,0 +1,130 @@
...
+    df = datasource.generate_breakdown_df(
+        'race_and_ethnicity', 'national')
+    expected_df_race_national = pd.read_csv(
+        GOLDEN_DATA['race_national'], dtype={'state_fips': str})
+
+    assert_frame_equal(df, expected_df_race_national, check_like=True)
+
+
+@mock.patch('ingestion.gcs_to_bq_util.load_csv_as_df_from_data_dir', side_effect=_load_csv_as_df_from_data_dir)
+def testGenerateSexNational(
```
benhammondmusic, line 41: this test function name should match what it's testing to avoid confusion later. it looks like this one should be `testGenerateAgeNational` and the next one down should be `...Sex...`

python/datasources/cdc_hiv.py
```diff
@@ -0,0 +1,228 @@
...
+def generate_alls_df(geo_level: str):
+    """
+    Fetches the combined dataframe of each demographic for county & state levels
+
+    geo_level: string of `state` or `county`
+
+    returns a formatted dataframe with total values for specified geo_level
+    """
+    alls_df = gcs_to_bq_util.load_csv_as_df_from_data_dir(
+        f'cdc_hiv/{geo_level}', f'totals_{geo_level}_2019.csv', dtype={'FIPS': str}, skiprows=9, thousands=',')
```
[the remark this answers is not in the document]
eriwarr: I will have to send in the filename and directory strings separately as the `load_csv_as_df_from_data_dir` function has an `os.path.join` internally and adds an extra `/` to the end of the file path, which causes an error.

python/ingestion/gcs_to_bq_util.py
```diff
@@ -239,15 +239,26 @@ def load_csv_as_df_from_web(url, dtype=None, params=None, encoding=None):
...
+    filename: file to load the csv file from
+    subdirectory: combined directory and filename path
+    skiprows: how many rows to skip when reading csv
+    thousands: reads commas in the csv file as a thousand place indicator
+    """
+    if subdirectory is None:
+        file_path = os.path.join(DATA_DIR, directory, filename)
+
+    else:
+        file_path = os.path.join(DATA_DIR, subdirectory)
```
benhammondmusic, line 259: i think `directory` and `filename` should always be required, and only `subdirectory` should be optional. then this line could be something like `file_path = os.path.join(DATA_DIR, directory, subdirectory, filename)`

python/ingestion/standardized_columns.py
```diff
@@ -122,6 +122,12 @@
 JAIL_PREFIX = "jail"
 INCARCERATED_PREFIX = "incarcerated"
 
+# HIV
+HIV_CASES = 'hiv_cases'
```
benhammondmusic, line 126: let's not call this cases, and instead keep the `diagnoses` terminology. since when we expand HIV to include other data types, it'll be confusing what the generic `cases` is referring to

python/datasources/cdc_hiv.py
```diff
@@ -0,0 +1,248 @@
...
+
+    return alls_df
+
+
+def generate_nat_breakdown_df(breakdown: str):
+    nat_from_state_df = gcs_to_bq_util.load_csv_as_df_from_data_dir(
+        subdirectory=NAT_BREAKDOWN_DICT[breakdown], skiprows=9, thousands=",")
+
+    nat_from_state_df['FIPS'] = '00'
+    nat_from_state_df['Geography'] = 'United States'
```
benhammondmusic, line 91: i agree with josh, these should be `constants.US_FIPS` and `constants.US_NAME`

benhammondmusic (approved):
looks good to me but holding on merging until @joshzarrabi gives final approval on his stuff.

joshzarrabi (approved):
Thanks Eric, looks great

PR #1979, opened 2023-02-09
Checks in time-based hiv source files to `/data`

benhammondmusic:
## Description

By clicking "underlying data" on the CDCAtlasPlus table interface, you're able to click and download the breakdown tables that include all years, all specific groups (alls are still separate), and all FIPS, including the population data in its own column.

Note: the site was timing out when I downloaded the county-level race and age tables, so I ended up splitting each by half of the groups and then manually combining the data in the tables. 

Note: this still isn't as ideal as a true API endpoint or hard-coded URL, but it saves us about a million clicks doing it via the map interface. 

## Motivation and Context

#1970 expanding to include time-series data

https://user-images.githubusercontent.com/41567007/217704992-32c80c97-8868-441c-aa28-388543e29632.mov

## Types of changes


eriwarr (approved):
Thanks so much!!!

PR #1999, opened 2023-02-16
Fixes deprecated `vaccinations` param; adds error handling

benhammondmusic:
## Description

- continue to handle only deprecated datatypes in the existing swap function
- create a new hook that handles deprecated dropdownIds in the params by redirecting to the new version
- the hook also redirects to a default setting if there is a malformed dropdown in the first setting
- adds a generic error boundary around the entire `/exploredata` which should catch any type of malformed param (currently messing with the mlp setting, the demo setting, and more will crash the site) and simply redirect to a default tracker setting
- changes suggested report links in help box to show HIV nationally and COVID age at county level

## Preview Link

<a href="https://deploy-preview-234--health-equity-tracker.netlify.app/exploredata" target="_blank">
  <br />
Click to demo updates </a> 

## Motivation and Context

a link we have been promoting for vaccinations was using the deprecated dropdown id param, and the existing fix doesn't work when there is no topic selected (our current default setting)

## Has this been tested? How?

manually

- deprecated datatypes still work (`deaths` is treated as `covid_deaths`)
- deprecated dropdownId is replaced (`vaccinations` to `covid_vaccinations`)
- other malformed params including bad fips, bad carousel mode, etc causes an error
- errors are handled by redirecting to the same URL stripped of all params/hash. It also temporarily shows the user a small text alert, and the nav bar is still available. If the automated redirect doesn't work the user could still manually navigate

https://user-images.githubusercontent.com/41567007/219285559-fbf7fab8-8fc7-4565-b5e2-15e6fbf1ef9a.mp4

https://user-images.githubusercontent.com/41567007/219285564-6e594924-7539-46f4-9a48-330c4ffc3c6d.mp4

https://user-images.githubusercontent.com/41567007/219285568-88966d97-3d9e-4274-bc85-994a85e030a9.mp4

https://user-images.githubusercontent.com/41567007/219285570-dd6b9113-b7c3-40d0-aa21-ad66d7543206.mp4

## Types of changes


frontend/src/ErrorBoundaryDropParams.jsx
```diff
@@ -0,0 +1,32 @@
...
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(error) {
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true };
+  }
+
+  componentDidCatch(error, info) {
+    // Example "componentStack":
```
eriwarr, line 15: Is this comment needed?

joshzarrabi (approved):
Thanks for fixing this so quickly

frontend/src/ErrorBoundaryDropParams.jsx
benhammondmusic: good catch; this was straight from the react docs.

PR #2012, opened 2023-02-22
Vite

benhammondmusic:
## Description

Upgrades the frontend build step to use Vite which is an extremely fast bundler, both in terms of developer experience waiting for builds and also client performance when the bundles are more streamlined and efficient. It is also a much more modern tool suitable for production, whereas many people are not big fans of using `create-react-app` for production environments

Biggest reasons for this change:
- It's fast (a few quick glances show our node build step going from ~2mins down to ~30secs or 4x faster, the overall Node action going down from ~7mins down to 4mins including tests)
- It's more modern and will (likely) be easier to maintain against evolving dependencies then webpack/react-scripts

- removes `react-scripts` and replaces with `vite` (the commands in package.json have been slightly tweaked but remain mostly the same
- removes `jest` and replaces with `vitest` (the API is almost identical so our tests didn't have to change)
- NOTE: I manually added `NODE_VERSION`: `16` to the Netlify environmental variables, without that specification Netlify was still trying to build HET frontend on Node 12. Adding that env fixed the netlify builds, but we need to watch for this issue on DEV and PROD

## Preview Link

<a href="https://deploy-preview-2012--health-equity-tracker.netlify.app/exploredata" target="_blank">
  <br />
Click to demo updates </a> 

## Motivation and Context

Code health, developer experience

## Has this been tested? How?

All vitest tests are passing, all playwright tests are passing, all manual pages i've checked are loading. Testing in DEV and PROD environments may expose some other errors based on the handling of envs

Screen Shot 2023-02-22 at 9 10 25 PM
Screen Shot 2023-02-22 at 9 39 37 PM

## Types of changes


joshzarrabi (approved):
beautiful

PR #2018, opened 2023-02-23
Improves CI and Testing

benhammondmusic:
## Description

- uses `slim` build of SuperLinter in GitHub action which removes heavy language that we weren't using anyways
- splits the Node CI into 3 jobs that run concurrently: Playwright E2E, Frontend Build, and Jest/CSpell. Right now it is installing node modules for each of these jobs which seems inefficient; we should look into #2022 
- prefers `npm ci` to `npm i` in github actions which is the recommended way to install packages on CI
- splits out many of the internal routes into their own tests to be run concurrently
- moves a11y checks to the end of each route check; previously these were flaky because they were being run against a react page that wasn't fully loaded (and unintentionally passing sometimes)
- bumps timeout for Playwright; before the first round of tests were always failing at 1min timeout, requiring another full run that took ~45 seconds. Now the first-round tests pass
- changes the "reporter" type which should make it easier to see real issues in playwright on CI
- reduced the number of workers to help with test flakiness
- adjusted Skip link placements that were failing a11y checks (and only now were caught by the improved test setups)

## Preview Link

<a href="https://deploy-preview-2018--health-equity-tracker.netlify.app/exploredata" target="_blank">
  <br />
Click to demo updates </a> 

## Motivation and Context

speed and reliability improvements of our testing / CICD process

## Has this been tested? How?

manually

Screen Shot 2023-02-27 at 12 30 48 PM

## Types of changes


eriwarr (approved):
This is awesome! Thanks Ben!

joshzarrabi (approved):
Nice work

PR #2020, opened 2023-02-27
BACKEND: Adds time-series data for Vera

benhammondmusic:
## Description

- [x] overwrites the values for ALL JAIL RAW COUNT with the SUM OF GROUPS RAW COUNTS for sex and race. This doesn't affect any other calculations as Vera already provides `per 100k` values, and it allows us to calculate the `pct_share` values for jail the same way (post-melt) that we do everything else. This gets us closer to the goal of getting the source data into a HET style df as soon as possible and then running our util functions
- [x] `pct_share` cols for `jail_` and `prison_` for every available year
- [x] reads in `100k` cols for `jail_` and `prison_` for every available year
- [x] reads in estimated count cols, including `total_confined_children` for every year
- [x] generates `pct_relative_inequity` for every available year
- [x] makes new `ingestion.types.py` file for reusable types that allow us to easily cast variables as the needed `Literal` types
- [x] moves constant column names into `std_col`

## Motivation and Context

Adds time-series data to Incarceration on the tracker

## Has this been tested? How?

- [x] updates tests and test files to handle time_series data 
- [x] adds stronger typing for enum string literals and casts as needed to pass mypy tests. NOTE: using the variable as the type didn't actually populate the options in vscode, only the type variable name. For this reason I've used the explicit `Literal["age","sex"]` etc when typing the function signatures.

Screen Shot 2023-03-01 at 6 36 41 PM
Screen Shot 2023-03-01 at 6 38 06 PM
Screen Shot 2023-03-01 at 6 49 13 PM
Screen Shot 2023-03-01 at 6 50 21 PM

## Types of changes


python/datasources/vera_incarceration_county.py
```diff
@@ -251,93 +215,135 @@ def write_to_bq(self, dataset, gcs_bucket, **attrs):
...
+                       *PCT_SHARE_COL_MAP.values(),
+                       *PCT_REL_INEQUITY_COL_MAP.values(),
+                       *RAW_COL_MAP.values(),
+                       POP,
+                       CHILDREN
+                       ]
+
+        # by_race gets race and race_id cols
         if demo_type == std_col.RACE_OR_HISPANIC_COL:
             std_col.add_race_columns_from_category_id(breakdown_df)
```
joshzarrabi, line 334: Probably need to add `race_and_ethnicity` col here too?

python/datasources/cawp_time.py
```diff
@@ -1,20 +1,26 @@
...
-import ingestion.standardized_columns as std_col
-from ingestion import gcs_to_bq_util, merge_utils
-from ingestion.dataset_utils import (generate_pct_rel_inequity_col,
-                                     zero_out_pct_rel_inequity)
-from ingestion.standardized_columns import Race
-from ingestion.merge_utils import ACS_EARLIEST_YEAR, ACS_LATEST_YEAR
-import pandas as pd
+from typing import Literal, cast, List
+
+SEX_RACE_AGE_TYPE = Literal["sex", "age", "race"]
```
joshzarrabi, line 18: Fancy

python/datasources/vera_incarceration_county.py
```diff
@@ -474,9 +498,55 @@ def generate_partial_breakdown(df, demo_type, data_type, property_type):
...
+    Returns same df replacing juvenile cols with a summed, rounded `total_confined_children` col
+    """
+    df[CHILDREN] = df[JUVENILE_COLS].sum(
+        axis="columns", numeric_only=True).round(0)
+    df = df.drop(columns=JUVENILE_COLS)
+
+    return df
+
+
+def add_jail_pct_share_col(df, demo_type: SEX_RACE_ETH_AGE_TYPE):
```
joshzarrabi, line 520: Is there a good reason to not massage all of the data to be ready for our util functions and then run those on them? If the main issue is that the sum wont add up to the total, maybe write a wrapper around the util function that accounts for that case (but leave it in this file so people dont accidentally use it)

joshzarrabi (approved):
Thanks for this

python/datasources/vera_incarceration_county.py
benhammondmusic: updated the comment; is that what you mean ?

python/datasources/cawp_time.py
benhammondmusic: I think we should slowly be typing all our functions; do you support that? I feel like now that I'm used to typescript it seems so silly to be able to break a function by sending in something of the wrong type. Python's typing isn't as easy to work with but seems ok for stuff like this where it needs to be specific keywords and it's nice when it populated in the IDE hints

python/datasources/vera_incarceration_county.py
benhammondmusic: Age didn't change because we don't have age groups anyways, but for sex and race I basically overwrite the ALL raw_count with the SUM OF GROUPS raw_count, and then proceed as normal and calc the pct_share after the melt using the normal util function. Saves having to keep track of the two types of "ALL" rows and we don't need the raw_count for per_100k calcs since vera provides those

python/datasources/cawp_time.py
eriwarr: Typescript is so great to use because of typing, but I always struggle to type with Python correctly. So I'll probably implement this in my HIV PR.

python/datasources/vera_incarceration_county.py
```diff
@@ -36,6 +43,11 @@
     POP: "incarceration_population_pct"
 }
 
+PCT_REL_INEQUITY_COL_MAP = {
+    JAIL: "jail_pct_relative_inequity",
```
eriwarr, line 47: Just curious, should these be constants, or is this the better pattern for this data source?

eriwarr (approved):

python/datasources/vera_incarceration_county.py
benhammondmusic: good call, that is the current pattern. Though I'm wondering now why we declare variable-specific constants globally.... seems like the only stuff that should be in that std_col file are things used across the codebase?

python/datasources/vera_incarceration_county.py
joshzarrabi: yeah

joshzarrabi (approved):
Looks great, thanks

PR #2024, opened 2023-03-01
BACKEND: Adds time-series data for HIV deaths and PrEP

eriwarr:
## Description
This PR adds two HIV time-series datasets, HIV deaths and PrEP,  to the backend. 

PrEP (16+):
- Data is available at every geographic level, but data for race and ethnicity are only available at the national level.
- The CDC obtains data from IQVIA and only has race/ethnicity data for "Black," "White," "Hispanic," and "Other." 
- Deaths and diagnoses do not have an "Other" category but instead use the  "Multiracial" category.  The  `MULTI_OR_OTHER_STANDARD_NH`, is likely the best category across the three datasets.
- To allow for comparison with other HIV indicators, the CDC presents for 13+ and 13–24-year-olds when viewing PrEP, but it represents 16+ and 16-24-year-olds (the remaining age groups are the same). We can handle this logic on the front end to prevent using separate tables. 
- `hiv_prep_coverage` is a percentage that measures the number of people prescribed with PrEP, divided by the estimated number of persons with indications for PrEP.

Deaths:
- Data is available only at the state and national level but is available for every demographic.

Note: Added optional parameter to the `pd.read_csv`,  function, na_values,  to turn data unavailable or suppressed into N/A values. This allows the thousands parameter to work as intended and removes the need to replace the commas in the data frame.   

## Has this been tested? How?

- Updated tests to include deaths and PrEP data. 
- Manual calculations to confirm percent share and percent inequities values.
- Airflow

- **National-race comparison**
- CDC:
Screenshot 2023-03-03 at 1 53 36 PM
Screenshot 2023-03-03 at 2 30 49 PM
- Airflow:
Screenshot 2023-03-03 at 2 43 08 PM

- **State-age comparison**
- CDC:
Screenshot 2023-03-03 at 2 37 54 PM
- Airflow:
Screenshot 2023-03-03 at 2 45 50 PM

## Types of changes

- Data Diver: contributes to the body of knowledge surrounding health equity (e.g., analysts, researchers, professors in academic or corporate settings)
- Data Interpreter: connects others by translating complex data into something actionable (e.g., think tank or legislative settings)
- Informer: helps everyday people understand important current events (e.g., journalists working in storytelling settings)
- Persuader: the passage of legislation by presenting concise and compelling arguments (e.g., staff members, policy advisors in legislative or nonprofit settings)


frontend/src/data/config/DatasetMetadata.ts
```diff
@@ -50,55 +50,55 @@ export const datasetMetadataList: DatasetMetadata[] = [
   {
     id: "cdc_hiv_data-race_and_ethnicity_county_time_series",
     name: "HIV Diagnoses by race/ethnicity and county",
```
benhammondmusic, line 52: These should all indicate diagnoses, deaths and PREP. Also, let's lowercase `Diagnoses` to keep things sentence cased

python/datasources/cdc_hiv.py
```diff
@@ -149,14 +189,50 @@ def load_df_from_data_dir(geo_level: str, breakdown: str):
     """
     load_df_from_data_dir fetches the csv file from the data dir
```
benhammondmusic, line 190: maybe indicate here (and in the fn name?) that this fn actually loads several source tables and combines them into a single df for the requested breakdown

python/ingestion/standardized_columns.py
```diff
@@ -127,6 +127,15 @@
...
 HIV_DIAGNOSES_PCT_SHARE = 'hiv_diagnoses_pct_share'
+HIV_DEATHS = 'hiv_deaths'
+HIV_DEATHS_PER_100K = 'hiv_deaths_per_100k'
+HIV_DEATHS_PCT_INEQUITY = "hiv_deaths_pct_relative_inequity"
+HIV_DEATHS_PCT_SHARE = 'hiv_deaths_pct_share'
+HIV_PREP = 'hiv_prep'
+HIV_PREP_COVERAGE = 'hiv_prep_coverage'
+HIV_PREP_PCT_INEQUITY = 'hiv_prep_pct_relative_inequity'
+HIV_PREP_PCT_SHARE = 'hiv_prep_pct_share'
+HIV_PREP_POPULATION = 'hiv_prep_population'
```
benhammondmusic, line 138: maybe worth adding a comment here explaining why there are unique population columns for PREP vs other HIV indicators?

python/datasources/cdc_hiv.py
benhammondmusic: or, would it make more sense to keep this function only loading a single CDC table, and then do the merging outside of the function? It looks like there is a lot of duplication between the multiple `read_csv` calls so could make sense to have the fn do a single read and then iterate of the needed types of file reads outside the fn ?

python/datasources/cdc_hiv.py
```diff
@@ -10,30 +10,45 @@
...
-    std_col.HIV_DIAGNOSES_PER_100K: {'Data suppressed': np.nan, 'Data not available': np.nan},
     std_col.RACE_CATEGORY_ID_COL: {
-        'All races/ethnicities': std_col.Race.ALL,
+        'All races/ethnicities': std_col.Race.ALL.value,
         'American Indian/Alaska Native': std_col.Race.AIAN_NH.value,
         'Asian': std_col.Race.ASIAN_NH.value,
         'Black/African American': std_col.Race.BLACK_NH.value,
         'Hispanic/Latino': std_col.Race.HISP.value,
-        'Multiracial': std_col.Race.MULTI_NH.value,
+        'Multiracial': std_col.Race.MULTI_OR_OTHER_STANDARD_NH,
```
joshzarrabi, line 30: I noticed in your comment that this is likely the best categorization for all three topics. It's OK for the topics to have different categorizations if that is correct, lets just make sure this is right.

joshzarrabi, line 30: You are missing a `.value`

python/ingestion/standardized_columns.py
```diff
@@ -127,6 +127,15 @@
 HIV_DIAGNOSES_PER_100K = 'hiv_diagnoses_per_100k'
 HIV_DIAGNOSES_PCT_INEQUITY = "hiv_diagnoses_pct_relative_inequity"
 HIV_DIAGNOSES_PCT_SHARE = 'hiv_diagnoses_pct_share'
+HIV_DEATHS = 'hiv_deaths'
```
joshzarrabi, line 130: For similar column names I have tried to use a `prefix` and `suffix` standard, and then to tie them together with the `generate_column_name` function. You can see examples in the `uhc.py` and `cdc_restricted.py` files

python/datasources/cdc_hiv.py
```diff
@@ -138,9 +168,19 @@ def generate_breakdown_df(self, breakdown: str, geo_level: str, alls_df: pd.Data
                                            std_col.HIV_DIAGNOSES_PCT_SHARE,
                                            std_col.HIV_POPULATION_PCT,
                                            std_col.HIV_DIAGNOSES_PCT_INEQUITY)
+        df = generate_pct_rel_inequity_col(df,
```
joshzarrabi, line 171: If you use the prefix/suffix pattern for col names you could put these into loops and simplify the code a little.

python/datasources/cdc_hiv.py
```diff
@@ -149,14 +189,50 @@ def load_df_from_data_dir(geo_level: str, breakdown: str):
...
+    # data for HIV diagnoses is available at every geo level and breakdown
+    diagnoses_df = gcs_to_bq_util.load_csv_as_df_from_data_dir(HIV_DIR,
+                                                               f'{DIAGNOSES}-{geo_level}-{breakdown}.csv',
+                                                               subdirectory=DIAGNOSES,
+                                                               skiprows=8,
+                                                               na_values=NA_VALUES,
+                                                               thousands=',',
+                                                               dtype=DTYPE)
+    df = diagnoses_df.rename(columns=DIAGNOSES_TO_STANDARD)
+    df = df.drop(columns='Indicator')
```
joshzarrabi, line 206: You should use constants for repeated strings like `Indicator` and `Multiracial`, as it helps reduce the probabilities of errors.

python/datasources/cdc_hiv.py
joshzarrabi: Yes I agree, try to make this a general function that takes in a generic Atlas plus source.

joshzarrabi (changes requested):
Thanks Eric, I think now that there are three sources and we want to add more, I think it would be worth it to think about how you can generalize the functions here to take in any source from Atlas plus and process them with less duplicated code. It'll be a little more of a time investment now but should make things easier going forward.

python/ingestion/standardized_columns.py
```diff
@@ -196,9 +190,16 @@ class Race(Enum):
     # so we do not know the relationship between Hispanic/Latino and race.
     # ETHNICITY_UNKNOWN refers to data that does not know whether the person is
     # Hispanic or Latino. (Some datasets use "ethnicity" to refer to whether
-    # someone is Hispanic or Latino)
+    # someone is Hispanic or Latino).
+    # PrEP data only collects data for the following race groups: Black, White,
+    # Hispanic, and Other. The dataset does nto distinguish what constitutes as other,
```
benhammondmusic, line 195: typo: `nto`

python/ingestion/standardized_columns.py
```diff
@@ -196,9 +190,16 @@ class Race(Enum):
...
+    # the other demographic does not include individuals who are hispanic.
+    # Using the a unknown race and ethnicity instead of other which is not could be anyone
+    # oustide the standard race group is preferred.
```
benhammondmusic, line 198: typo `oustide` and in general i think we can make this comment more clear (especially once we decide what is happening with the race group and what we want to call it)

joshzarrabi (approved):
Great work Eric, thanks a ton

PR #2031, opened 2023-03-03
Have node action reuse cached node_modules across jobs

benhammondmusic:
## Description

- starts by installing or loading cached `node_modules` based on the lock file changes
- has subsequent jobs wait for this to complete before running; each should retrieve the cached modules and eliminate need to independently re-install for each job

## Motivation and Context

reduce wasted computing / transfers

## Types of changes


benhammondmusic:
Closed as didnt improve speed and vastly complicated the code

Issue #1491, opened 2022-02-28
Calculate all national poverty numbers on the backend


joshzarrabi:
## Background

This is in service of #1474 

Currently our process to generate state and county level property statistics is the following:

1. Get the state and county level information from the acs api [in python](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/python/datasources/acs_poverty.py#L288-L320)

2. Write output to big query tables with the following columns

* `below_poverty_line`
* `above_poverty_line`

for each demographic group and location level (state and county)

3. Aggregate the numbers to national [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsPovertyProvider.ts#L53-L60) 

4. Generate the total and calculate per_100k [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsPovertyProvider.ts#L87-L95)

5. Generate the `pct_share` and `pop_pct_share` [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsPovertyProvider.ts#L101-L115)

Note we did not use the normal acs population numbers for these. See: https://github.com/SatcherInstitute/health-equity-tracker/issues/669

## Problems

The code is unnecessarily complicated, hard to test, and asking the user to do lots of calculations and download way more data than they need.

## Proposed Implementation

Calculate the following metrics on the backend

* `poverty_per_100k`
* `poverty_pct_share`

## End state

I can log onto big query and see the following tables, with the above columns in them.

- [ ] `acs_poverty.by_sex_national`
- [ ] `acs_poverty.by_race_national`
- [ ] `acs_poverty.by_age_national`

-----------------------------

- [ ] All of [this code](https://github.com/SatcherInstitute/health-equity-tracker/blob/d9dca7f563f0991f240332d4f823931271bd598f/frontend/src/data/variables/AcsPovertyProvider.ts#L52-L146) is deleted on the frontend.

PR #2034, opened 2023-03-07
Create generic ACS condition provider with poverty and health insurance

joshzarrabi:
## Description

* Creates a new data set provider called `acs_condition` which can take in arbitrary metrics from ACS and generate age/sex/race tables for them.
* Removes the poverty and health insurance providers.
* Also fixes a miscalculation with the poverty sex and age numbers. 

- The acs_condition code is based off of the acs_health_insurance provider code, as that was more up to date and already had all calculations done on the backend. 

## Motivation and Context

Backend portion of #1491 and #1490

## Has this been tested? How?

Yes, added tests and tested on my airflow.

Screenshot 2023-03-08 at 6 49 33 PM

## Types of changes

- Data Diver: contributes to the body of knowledge surrounding health equity (e.g. analysts, researchers, professors in academic or corporate settings)
- Informer: helps everyday people understand important current events (e.g. journalists working in storytelling settings)
- Persuader: passage of legislation by presenting concise and compelling arguments (e.g. staff members, policy advisors in legislative or nonprofit settings)


airflow/dags/acs_condition.py
```diff
@@ -0,0 +1,66 @@
+# Ignore the Airflow module, it is installed in both dev and prod
+from airflow import DAG  # type: ignore
+from airflow.utils.dates import days_ago  # type: ignore
+
+import util
+
+_ACS_BASE_URL = "https://api.census.gov/data/2019/acs/acs5"
```
benhammondmusic, line 7: is 2019 the most recent year we can use ?

config/data_sources/acs_condition.tf
```diff
@@ -0,0 +1,5 @@
+# Create a BigQuery dataset for acs health insurance information
```
benhammondmusic, line 1: this comment should indicate this is pulling more than just health insurance from ACS

python/tests/datasources/test_acs_condition.py
```diff
@@ -37,7 +37,7 @@ def _get_by_race_as_df(*args):
 @mock.patch('ingestion.gcs_to_bq_util.load_values_as_df',
             side_effect=_get_by_race_as_df)
 def testSexNational(mock_acs: mock.MagicMock, mock_fips: mock.MagicMock):
-    acsHealthInsurance = AcsHealthInsurance()
+    acsHealthInsurance = AcsCondition()
```
benhammondmusic, line 40: should probably rename the instantiated class variable to something more generic since we aren't just testing health insurance right? something like `datasource = AcsCondition()` or whatever

python/datasources/acs_condition.py
```diff
@@ -0,0 +1,509 @@
...
+
+                if demo == RACE:
+                    add_race_columns_from_category_id(df)
+
+                dfs[table_name] = df
+
+        for table_name, df in dfs.items():
+            float_cols = [std_col.UNINSURED_PER_100K_COL,
+                          std_col.UNINSURED_POPULATION_PCT,
+                          std_col.UNINSURED_PCT_SHARE_COL]
```
benhammondmusic, line 274: should this list include POVERTY metric col names as well?

python/datasources/acs_condition.py
```diff
@@ -0,0 +1,509 @@
...
+
+        return df
+
+    def post_process(self, df, demo, geo):
+        """Merge population data, state, and county names.
+           Do all needed calculations to generate per100k and pct share
+           columns.
+           Returns a dataframe ready for the frontend.
+
+           df: Dataframe with raw health insurance numbers.
```
benhammondmusic, line 452: generalize this doc string to reflect multiple ACS raw counts not just health insurance

benhammondmusic (approved):
nice. excited to extend this for more topics soon

config/data_sources/acs_condition.tf
joshzarrabi: good catch, thanks

python/tests/datasources/test_acs_condition.py
joshzarrabi: Yes

python/datasources/acs_condition.py
joshzarrabi: Good catch, kind of weird this didnt cause an issue

benhammondmusic: I think it infers the type of it's not provided. We may not have to explicitly define the float cols actually, because nulls are allowed in bq floats. I wonder if it would auto-cast to ints if it's all whole numbers... hm

benhammondmusic: Actually the fn should have made them strings on bq

Issue #2049, opened 2023-03-13
EPIC: Get 2020 territory pops


benhammondmusic:
## Background

HET currently relies on population data from 2010 for the smallest territories

## What's Missing / Problems with the current approach

Obviously this data is out of date, and not as accurate as we would like. It also lacks data for the territory county equivalents

## Proposed Solution
- [x] add new pipeline `decia_2020` that makes data available on our backend #2036 
- [x] run pipeline to get it on PROD
- [ ] #2051 
- [ ] Frontend: update frontend dataset_ids to reflect correct source used, including data shown directly on frontend and also datasets used in calculations on the backend
- [ ] Frontend: update alert to explain where / when all data is coming from. Indicate that we have used non-NH races as _NH races for AS, GU, and MP where _NH races are not available. The number of people identifying as hispanic in these areas is extremely low, and using the numbers this way allows us to include these islands' populations in more of our calculations and reports

## End State 

All Census population data on the tracker (other than time-series data from 2010-2020) will be sourced from 2020 instead of 2010

## How will this move the needle on health equity?
Striving for accuracy and transparency in data collection and calculation is fundamental. Using more current numbers will improve accuracy and legitimacy for the important and often overlooked part of the country.

PR #2051, opened 2023-03-13
Backend: Use `decia_2020` for territory populations

benhammondmusic:
## Description

- [x] update `geo_context` data source ingest code to get population using the standard merge util #2054 
- [x] update `merge_utils` to get territory population and population_share data from the new source; this will load it into all other data sources as needed
- [x] for older time_series merges: use `acs_2010` , for newer use `decia_2020`  Continue not merging historic population for territories data pre-2010
- [x] rename existing `acs_2010` to `decia_2010` to prevent future confusion
- [x] for islands that don't have _NH race numbers (GU, AS, MP), merge the non-NH population AS _NH populations, which will allow more of our data sources to use these population numbers. TODO: Address this in the methodology, explaining _NH races aren't available, likely due to the extremely low number of people identifying as hispanic in the three island areas.
- [x] split covid pipeline by demographic to avoid 10 minute timeout; improve testing clarity

## Motivation and Context

backend for #2049 

## Has this been tested? How?

- [x] tests updated for `geo_context`
- [x] tests update for merge_util to include merging county-equivalents
- [x] tests updated for data sources using mocked population data
- [x] all changed pipelines run on my airflow

- County equivalents now get population numbers and calculated rates
- time series sets show different population data from up to 2015 vs 2016 and beyond
Screen Shot 2023-03-17 at 8 04 03 AM
Screen Shot 2023-03-17 at 8 24 43 AM
Screen Shot 2023-03-17 at 8 38 13 AM

## Types of changes


joshzarrabi (approved):

airflow/dags/cdc_restricted.py
```diff
@@ -91,10 +141,15 @@
 
 # CDC Restricted Data Ingestion DAG
 (
-    dummy_operator >> [
-        cdc_restricted_bq_op_race,
-        cdc_restricted_bq_op_sex,
-        cdc_restricted_bq_op_age
+    cdc_restricted_bq_op_sex_county >>
```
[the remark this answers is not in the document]
benhammondmusic: Oh ya. I was thinking it'd time out but you're right

python/ingestion/merge_utils.py
```diff
@@ -126,25 +128,37 @@ def merge_state_ids(df, keep_postal=False):
     return df
 
 
-def merge_pop_numbers(df, demo: str, loc: str):
+def merge_pop_numbers(df, demo: Literal['age', 'race', 'sex'], loc: Literal['county', 'state', 'national']):
```
eriwarr, line 131: Could we use the `SEX_RACE_AGE_TYPE` and `GEO_TYPE` types for this function and others also using it?

python/ingestion/merge_utils.py
```diff
@@ -231,34 +246,45 @@ def _merge_pop(df, demo, loc, on_time_period: bool = None):
...
-
     if std_col.STATE_FIPS_COL in df.columns:
         needed_cols.append(std_col.STATE_FIPS_COL)
 
     if loc == 'county':
         needed_cols.append(std_col.COUNTY_FIPS_COL)
 
-    pop_df = pop_df[needed_cols]
+    keep_cols = ([*needed_cols, std_col.TIME_PERIOD_COL]
+                 if on_time_period
```
eriwarr, line 256: Nice!

airflow/dags/cdc_restricted.py
```diff
@@ -91,10 +141,15 @@
...
+    cdc_restricted_bq_op_age_county >> [
+        cdc_restricted_bq_op_race_state,
+        cdc_restricted_bq_op_sex_state,
+        cdc_restricted_bq_op_age_state
+    ] >> cdc_restricted_bq_op_race_county >> [
```
eriwarr, line 149: Should this county task be grouped with the others? (not a suggestion, just curious)

airflow/dags/decia_2010_territory_population.py
```diff
@@ -0,0 +1,63 @@
+from airflow import DAG
+from airflow.models import Variable  # type: ignore
+from airflow.utils.dates import days_ago  # type: ignore
+
+import util
+
+_DECIA_2010_POPULATION_GCS_FILENAMES = (
+    'decia_2010_territory_population-by_race_and_ethnicity_territory.json,'
+    'decia_2010_territory_population-by_sex_territory.json,'
```
eriwarr, line 9: The commas may be in the wrong place here. 

```
_DECIA_2010_POPULATION_GCS_FILENAMES = (
    'decia_2010_territory_population-by_race_and_ethnicity_territory.json',
    'decia_2010_territory_population-by_sex_territory.json',
    'decia_2010_territory_population-by_age_territory.json'
)
```

eriwarr (changes requested):

python/datasources/decia_2020_territory_population.py
```diff
@@ -202,10 +216,17 @@ def generate_breakdown_df(self,
             pct_share_group_cols_map = get_melt_map(
                 SEX_CODES_TO_STD, TMP_PCT_SHARE_SUFFIX)
         if breakdown == std_col.RACE_OR_HISPANIC_COL:
+
+            race_map = {**RACE_CODES_TO_STD[postal], **NON_NH_TO_NH_RACE_MAP}
+
+            print("race_map")
+            print(race_map)
```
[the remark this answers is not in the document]
benhammondmusic: nope :)

python/ingestion/merge_utils.py
benhammondmusic: So what I found was, when I used a variable as a type, the vscode intellisense didn't render the underlying types (like it does for typescript). So on hover you only saw the SEX_AGE_RACE_TYPE instead of the actual strings needed.  I definitely think it's helpful to be able to hover a function call and see the specific strings required to make that fn work, otherwise we'd have to click into the function and then hover that inner type variable. There may be a way to fix this in vs code; because definietly I'd rather use reusable types

Defining function signature explicitly:
Screen Shot 2023-03-20 at 2 09 26 PM

Using variable in function signature (for demographic breakdown): 
Screen Shot 2023-03-20 at 2 12 42 PM

airflow/dags/cdc_restricted.py
benhammondmusic: I think we can just run them all in parallel so I'll try that. I had moved it in the middle because you cant have two subsequent lists of tasks; so I was trying to break them up with the single longer task. Could also use a dummy operator there instead

airflow/dags/decia_2010_territory_population.py
benhammondmusic: oh weird, ya before they were inside the strings.... why? @joshzarrabi ?

code before:
```
_ACS_2010_POPULATION_GCS_FILENAMES = (
    'acs_2010_population-by_race_and_ethnicity_territory.json,'
    'acs_2010_population-by_sex_territory.json,'
    'acs_2010_population-by_age_territory.json'
)
```

eriwarr (approved):

airflow/dags/decia_2010_territory_population.py
benhammondmusic: good catch. for future reference, this is just making one long, comma separated string, in the strange way python decided to implement multi-line strings

PR #2054, opened 2023-03-14
Backend: Refactor `geo_context` to get `population` col from merge util

benhammondmusic:
## Description

In order to streamline the code (and prepare for more complete territory data as part of #2051 ), it makes sense to have `geo_context` retrieve population count data using the normal `merge_pop` util function, rather than calling the BQ table directly and replicating the territory / county-equivalent merge logic. To do so involved adding a new function that could be used elsewhere in existing code and for future data:

- adds `dataset_utils/scaffold_fips_df()` which returns a single column DF with every FIPS code in an appropriate FIPS column for the given geo level
- uses this to start the geo_context df, and then merges pop data onto it; this allows the merge util to merge territory data onto state level requests, and also will allow a single change to the merge util to access the county-equivalent pop data from `decia_2020`, rather than making multiple changes to access this new data 

## Motivation and Context

prepare for #2051 

## Has this been tested? How?

automated tests updated and still passing

## Types of changes


eriwarr (approved):
💪🏿

PR #2066, opened 2023-03-17
Frontend: Updates HET language to indicate decia_2020

### LINKED ISSUE #2049: EPIC: Get 2020 territory pops

benhammondmusic:
## Description

- updates Vitest; installs helper package that suppresses warning re: canvas
- [x] swaps existing references to `acs_2010` with `decia_2010` in dataset metadata 
- [x] adjusts logic in all providers manually pushing decia pop as a "consumed dataset id" to use 2010 only for time_series metricQueries; 2020 should push to time_series and "current" requests 
- [x] adjust logic to push correct table based on state or territory level
- [x] ensure Frontend calcs are being done properly for vaccination providers; prioritize backend calc refactor to eliminate this 

## Motivation and Context

closes #2049 

## Has this been tested? How?

- [x] Frontend unit tests updated to reflect correct dataset loading

## Types of changes


eriwarr (approved):
Great job!!

Issue #1481, opened 2022-02-22
Move all COVID Vaccination National Calculations to the backend


joshzarrabi:
## Background

This is in service of: https://github.com/SatcherInstitute/health-equity-tracker/issues/1474

Right now we get most of the correct data from the CDC API directly, however because we don't have a national population dataset on the backend yet, we have to merge it on [the frontend here](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/VaccineProvider.ts#L64-L91)

This is bad because frontend merges are very slow.

## Proposed Implementation

After completing https://github.com/SatcherInstitute/health-equity-tracker/issues/1475, merge those national `population_pct` population numbers on the backend. 

## End State

- [ ] When I log onto big query, I can see the `cdc_vaccine_national.*` datasets with the correct `population_pct` column. 
- [ ] [This code](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/VaccineProvider.ts#L64-L91) is deleted from the frontend.

PR #2074, opened 2023-03-22
Downloads processed files for state and national vaccine

### LINKED ISSUE #1480: Move all state level vaccine calculations to the backend
##  Background 

Currently, our state level vaccine processing is a mish mash of complicated code handling a bunch of different cases. The reason for this is because the Kaiser Family Foundation presents the numbers in a bunch of different ways and we have to have cases for each of them. 

The current structure is as follows:

* Get the raw numbers for state total vaccinations from the kff [in python](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/python/datasources/kff_vaccination.py#L145-L159)

* Get the percentages of each race of people vaccinated, and the population pct defined by the KFF from the KFF [in python](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/python/datasources/kff_vaccination.py#L112-L142)

* Merge the state total population numbers using [big query sql](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/config/data_sources/AGG_kff_vaccination.sql)

* Multiply the `pct_share`, `population_pct_share` and `vaccinated_pct` by the correct constants to get the correct corresponding numbers [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/frontend/src/data/variables/VaccineProvider.ts#L93-L124)

* Calculate the per_100k metric for TOTAL population [on the frontend](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/VaccineProvider.ts#L140-L147).

## Why is this bad

The frontend code is very complicated and slower than it needs to be. Its weird that our `TOTAL` row in the big query table has a raw count of vaccination while the other races have percentages.

## Proposed Implementation
Every row on the big query table should have the same, correctly filled out columns 

* `vaccinated_pct_share`
* `vaccinated_per_100k`
* `population_pct`
* `acs_vaccine_population_pct`

## End State

- [ ] I can log onto big query and see the above columns filled out in the `kff_vaccination.race_and_ethnicity` table
- [ ] [This code](https://github.com/SatcherInstitute/health-equity-tracker/blob/281052a957b096a26aac28c79766a2be62d0c7dc/frontend/src/data/variables/VaccineProvider.ts#L93-L162) is deleted

joshzarrabi:
## Description

* Downloads the processed files for state and national vaccine data
* Removes all frontend processing code for both of them

## Motivation and Context

Frontend portion of #1480 and #1481
Closes #1480 and #1481

## Has this been tested? How?

Yes, tested locally 

Screenshot 2023-03-27 at 12 49 47 PM

## Types of changes

- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD

## Preview link
Preview link will be below in an auto-generated comment from Netlify 😎


frontend/src/data/variables/VaccineProvider.ts
```diff
@@ -67,104 +63,14 @@ class VaccineProvider extends VariableProvider {
...
-      df = totalDf.concat(nonTotalDf).resetIndex();
-
-      df = df
-        .generateSeries({
-          vaccinated_share_of_known: (row) => row["vaccinated_pct_share"],
-        })
-        .resetIndex();
+      consumedDatasetIds = consumedDatasetIds.concat(
+        "acs_population-by_race_state",
+        "acs_2010_population-by_race_and_ethnicity_territory"
```
benhammondmusic, line 72: this should be `decia_2020_territory_population-by_race_and_ethnicity_territory_state_level` now right? And there is some more conditional logic, like the territory pops are only used on national reports with undefined `filterFips`, or on state level reports where `isIslandArea(filterFips) =true`

benhammondmusic (approved):
nice! just update that territory pop to whatever table is actually being merged on the backend

PR #2098, opened 2023-03-31
Keep state maps when county data unavailable

benhammondmusic:
## Description

- detects reports where there is state-level data but not county-level data
- instead of suppressing an empty county-level state map in those cases, the map will continue to show a state level map, populated with only a single color representing the state level data. the user is not able to hover counties or click down to the county level
- behavior extends to the small multiples modal as well
- adds small warning alert that county-level data isn't available

## Motivation and Context

Having "missing data" as the first card of state level reports is not a great user experience; it makes it seem like there isn't any STATE data at all further down the page, which isn't true, there is STATE data but just not the COUNTY data that populates our first card on the STATE level report

## Has this been tested? How?

Manually; playwright tests

BEFORE:

Screen Shot 2023-03-30 at 9 41 20 PM

AFTER:

Screen Shot 2023-03-30 at 9 37 30 PM
Screen Shot 2023-03-30 at 9 37 59 PM

## Types of changes
- New content or feature

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
👍🏿

PR #2101, opened 2023-03-31
Fix ESLINT 

benhammondmusic:
## Description

- removes broken reference to create-react-app config
- init new eslint config
- customize rules to get passing now; can tighten rules progressively if we want

## Motivation and Context

we were unable to commit changes because `lint-staged` was breaking on the eslint step

## Has this been tested? How?

This commit pushed!

## Types of changes
- Bug fix
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
👍🏿

PR #2124, opened 2023-04-13
Updates the HIV exclusions list to include the new topics

eriwarr:
## Description

- bulleted high-level changes

## Types of changes
- Bug fix
- New content or feature
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD

### DIFF
--- a/frontend/src/data/utils/datasetutils.ts
+++ b/frontend/src/data/utils/datasetutils.ts
@@ -201,6 +201,13 @@ const NON_STANDARD_AND_MULTI: DemographicGroup[] = [
   MULTI_OR_OTHER_STANDARD_NH,
 ]
 
+const VARIABLES_NEEDING_13PLUS: VariableId[] = [
+  'hiv_deaths',
+  'hiv_diagnoses',
+  'hiv_care',
+  'hiv_prevalence',
+]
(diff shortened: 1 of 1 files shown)


frontend/src/data/utils/datasetutils.ts
```diff
@@ -201,6 +201,13 @@ const NON_STANDARD_AND_MULTI: DemographicGroup[] = [
   MULTI_OR_OTHER_STANDARD_NH,
 ]
 
+const VARIABLES_NEEDING_13PLUS: VariableId[] = [
```
benhammondmusic, line 204: Can we avoid duplication between this array and the one in the HIVSuffix hook? https://github.com/SatcherInstitute/health-equity-tracker/blob/b4cee7f337b2d2dbe8e915ffe77dd8db6629ebee/frontend/src/utils/hooks/useHIVLabelSuffix.tsx#L8

I feel like we should define global constants that are topic-specific in their Frontend provider file; and then export/import them as needed throughout the app

benhammondmusic (approved):
If you can combine those duplicated lists and share it from the provider file otherwise this looks good to go

PR #2174, opened 2023-05-04
Extract Chart Titles to HTML from Vega

### LINKED ISSUE #2118: Chart titles overlap legend on tiny mobile
Three line chart titles overlap legend on smallest mobile devices (we need to code for down to 360w ideally)

Image

### LINKED ISSUE #2131: Long titles cause hover/shrink with Vega
**Describe the bug**
Hovering over Vega charts where the label exceeds the container size causes the strange "shrinking" behavior when hovering

**To Reproduce**
1. Go to https://dev.healthequitytracker.org/exploredata?mls=1.covid_deaths-3.13121&demo=race_and_ethnicity&mlp=disparity#population-vs-distribution
2. make the screen the right width to cut off the chart title
3. hover the comparison bar chart

**Expected behavior**
chart shouldn't shrink

**Screenshots**

https://user-images.githubusercontent.com/41567007/232377852-8cf3665d-7e0c-4955-9532-a0d92bb894fe.mov

### LINKED ISSUE #2143: Move titles out of Vega and use regular React/HTML
## Background
Right now some titles are done as normal HTML/React and styled with MUI/css, while others are embedded within the Vega component (maps, bar charts). Vega is much more difficult to deal with for this type of text rendering and styling and requires an annoying amounf of work to handle simple things like word wrapping

## Proposed Solution

- As I've done for MapCard with #2138 , we should extract the title from the UnknownsMap, SimpleBarChart and DisparityBarChart, and instead use the same Grid setup to have the title appear directly above the chart, centered, as normal HTML
- should rely on sass variables for constants and consider using a @ mixin to get the same exact card title styling in every card. Could also consider using a particular MUI component styled as we need like Typography or similar

## End State 
The vega viz will only render the charts and their legends/labels/axes. The title will be outside as a normal component

Clean, responsive design is essential to a good user experience, and the codebase will simplify be removing the extensive logic required to handle title wrapping in Vega (which is still causing lots of issues like #2131 and #2118

benhammondmusic:
## Description

- moves the remaining chart titles from being rendered in Vega (Unknown Map, Simple Bar, Disparity Bar) and now renders as normal HTML
- makes component to render styled `<ChartTitle />` in every card consistently
- remove / refactor Vega title utils
- update `MetricConfig` chart titles to be single strings, update language
- disable Vega actions everywhere rather than via props
- fixes confusing / inaccurate wording in methodology re: `relative inequity`
- updates `// TODO:` to use consistent format for easier searching

## Motivation and Context

closes #2143 
fixes #2131 
fixes #2118 

basically should look like it did, other than new wording and some spacing adjustments

## Types of changes
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
🙌🏿

PR #2178, opened 2023-05-09
Fix cspell for non-frontend files vscode

benhammondmusic:
## Description

- moves cspell config file to the root rather than `/frontend`
- updates ignored directories and words
- adjusts npm command to run against root directory (detecting new config file)
- adjust lint-staged to use  npm command for the same functionality regardless of method: `npm run cspell`, `npx cspell .`, or as a pre-commit hook on changed files
- adds summary to cspell output to report # of checked files

## Motivation and Context

only frontend files detected custom dictionary; anything backend was causing squiggly lines for python files etc even though those files had been safe-listed

## Has this been tested? How?

manually
- running the frontend `npm run cspell` passes
- running `npx cspell .` passes
- passes on commits
- fails on local commits and PR pushes that intentionally include a typo in `/python` or `/frontend`

Intentional typo fails:
Screen Shot 2023-05-10 at 12 33 03 PM

## Types of changes
- Refactor / chore


eriwarr (approved):
👍🏿

PR #2182, opened 2023-05-11
Moves data table tiles to MetricConfig

eriwarr:
## Description

- bulleted high-level changes

## Types of changes
- Bug fix
- New content or feature
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD

### DIFF
--- a/frontend/src/cards/TableCard.tsx
+++ b/frontend/src/cards/TableCard.tsx
@@ -139,8 +139,8 @@ export function TableCard(props: TableCardProps) {
                     (colName) => !NEVER_SHOW_PROPERTIES.includes(colName)
                   )}
                   variableId={props.variableConfig.variableId}
-                  variableName={props.variableConfig.variableFullDisplayName}
                   fips={props.fips}
+                  dataTableTitle={props.variableConfig.dataTableTitle ?? ""}
                 />
               </div>
             )}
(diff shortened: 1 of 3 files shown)


frontend/src/charts/TableChart.tsx
```diff
@@ -190,7 +175,7 @@ export function TableChart(props: TableChartProps) {
         <figure>
           <figcaption>
             <ChartTitle
-              title={`Breakdown summary for ${sentenceCaseName} in ${props.fips.getSentenceDisplayName()}`}
+              title={`${props.dataTableTitle} in ${props.fips.getSentenceDisplayName()}`}
```
benhammondmusic, line 178: it's possible this could be sent in as an empty string right, so the title would read ` in Texas` or something; maybe instead of empty string as a fallback in the parent we could use a generic term like `Data`. I'll add a note up there

frontend/src/cards/TableCard.tsx
```diff
@@ -139,8 +139,8 @@ export function TableCard(props: TableCardProps) {
                     (colName) => !NEVER_SHOW_PROPERTIES.includes(colName)
                   )}
                   variableId={props.variableConfig.variableId}
-                  variableName={props.variableConfig.variableFullDisplayName}
                   fips={props.fips}
+                  dataTableTitle={props.variableConfig.dataTableTitle ?? ""}
```
benhammondmusic, line 143: Maybe this should default to `Data` or something generic so that the title never gets weird and unreadable if we forget to fill in the optional metric config value

benhammondmusic (approved):
awesome! so much cleaner and more explicit. thanks Eric!

PR #2184, opened 2023-05-12
Adds GH action to allow pipeline testing on INFRA TEST

benhammondmusic:
## Description

- adds new GH action triggered on a merge to `infra-test` branch, which runs all pipeline related actions but not the data-server and frontend actions. this **should** ~~will~~ allow us to test out backend changes on InfraTest without blocking the `main` branch of our repo, which then blocks all releases

## Motivation and Context

Reducing cloud spend by removing personal gcloud setups and only keeping INFRA TEST and INFRA PROD

## Has this been tested? How?

nope. YOLO! (we can't really test until this is merged to `main`)

## Types of changes
- Refactor / chore

PR #2211, opened 2023-05-16
Updates Readme with new pipeline staging and cleanup

benhammondmusic:
## Description

- Adds specific info on running staging tests on backend code changes
- Hides complex, unused README info behind expandable dropdowns
- General cleanup

## Motivation and Context

Documentation for #2184 

## Types of changes

- Refactor / chore


README.md
kccrtv, line 1: is it necessary to keep the sections labeled UNUSED? if so, should we consider using the term DEPRECATED instead?

benhammondmusic: I think you're right; I was reluctant to remove them simply because a lot of work was done by Google to make this a "make your own Health Equity Tracker" concept. I think what I'll do is move all of those UNUSED items to the bottom in an expandable section with something like "Set up your own HET". So the bare-minimum items required for updating the existing tracker are up front for onboarding / maintainers, and then people wanting to clone and utilize the code structures in their own project will still have access to the complicated local docker / terraform / gcloud in the expandable bottom section

kccrtv (approved):
👍🏼

PR #2223, opened 2023-05-22
FRONTEND: Adds card options menu

eriwarr:
## Description

- bulleted high-level changes

## Types of changes
- Bug fix
- New content or feature
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD

### DIFF
--- a/frontend/src/cards/Card.module.scss
+++ b/frontend/src/cards/Card.module.scss
@@ -28,6 +28,7 @@
 
 .ChartCard {
   margin: $card-gutter;
+  position: relative;
 }
 
 .SmallMarginContent {
--- a/frontend/src/cards/ui/Sources.tsx
+++ b/frontend/src/cards/ui/Sources.tsx
@@ -9,10 +9,7 @@ import { DATA_CATALOG_PAGE_LINK } from '../../utils/internalRoutes'
 import { DataSourceMetadataMap } from '../../data/config/MetadataMap'
 import { type MetricQueryResponse } from '../../data/query/MetricQuery'
 import { DatasetMetadataMap } from '../../data/config/DatasetMetadata'
-import CopyLinkButton from './CopyLinkButton'
-import { type ScrollableHashId } from '../../utils/hooks/useStepObserver'
 import { Grid } from '@mui/material'
-import { DownloadCardImageButton } from './DownloadCardImageButton'
 
 function insertPunctuation(idx: number, numSources: number) {
   let punctuation = ''
@@ -84,8 +81,6 @@ interface SourcesProps {
   metadata: MapOfDatasetMetadata
   isAgeAdjustedTable?: boolean
   hideNH?: boolean
-  scrollToHash?: ScrollableHashId
-  downloadTargetScreenshot: () => Promise<boolean>
 }
 
 export function Sources(props: SourcesProps) {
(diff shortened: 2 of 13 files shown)


frontend/src/cards/ui/OptionMenuIcons.tsx
benhammondmusic, line 1: Can we rename this something like `CardShareIcons` or similar? I think it needs a more specific name to indicate it's card-level and that it contains the share to social buttons

frontend/src/cards/ui/OptionMenuIcons.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+interface OptionMenuIconsProps {
+    popover: PopoverElements
+    reportTitle: string;
+}
+
+export default function OptionMenuIcons(props: OptionMenuIconsProps) {
+    let title = `Health Equity Tracker - ${props.reportTitle}`
+    const emailShareBody = `${title}${'\n'}${'\n'}` // Add line breaks here if needed
+    const sharedUrl = window.location.href
```
benhammondmusic, line 31: The `sharedUrl` should also include the card `#hash` so that the shared link on social navigates directly to the card that was shared. That prop is already available a level up in the copy link function, so you can just drill it down into this component as well and tack it onto the end of the shared URL

frontend/src/cards/ui/Sources.tsx
```diff
@@ -149,25 +147,32 @@ export function Sources(props: SourcesProps) {
           <p className={styles.FootnoteTextNH}>Note. NH: Non-Hispanic. </p>
         )}
       </Grid>
-      {/* Sources inline with card action buttons */}
-      <Grid item xs={8} sm={9} md={10} container alignItems={'center'}>
-        {sourcesInfo}
-      </Grid>
+      {props.isMulti ?
```
benhammondmusic, line 150: i think we should store the state of the multimap view in the URL bar, and then the user can link directly to an opened multimap just like any other view. im ok merging this first and then adding that as a follow up feature, or holding on this and getting that working as well, up to you

benhammondmusic: but then the multimap modal could have a normal 3-dot menu just like every other card gets

frontend/src/cards/Card.module.scss
```diff
@@ -28,6 +28,7 @@
 
 .ChartCard {
   margin: $card-gutter;
+  position: relative;
```
benhammondmusic, line 31: What's this doing ?

frontend/src/cards/ui/OptionMenuIcons.tsx
eriwarr: The #hash is included in the URL when accessing `window.location.href`. The link takes you directly to the card. I console-logged the URL for each card to be sure. 

> https://deploy-preview-2223--health-equity-tracker.netlify.app/exploredata?mls=1.hiv_care-3.00&group1=All&group2=All**#rate-map**

benhammondmusic: I don't think this is working then; when I click "share to email" the link that is generated in the email doesn't contain a #hash 


https://github.com/SatcherInstitute/health-equity-tracker/assets/41567007/a70f8177-ddc1-42a2-87bc-e0fe9513aeab

benhammondmusic (approved):
Awesome! Thanks so much for hitting all the details; it looks and works great!!

PR #2245, opened 2023-06-07
Fixes the warning in console for the color scale

eriwarr:
## Description

Fixes the error in the console where the component attempts to access the color scale in the Vega spec before the Vega spec is defined. 

- Updated the initial state of the spec to be null. 
- Updated the render not to load the Vega component until the Vega spec is defined. 

## Motivation and Context

Screenshot 2023-06-07 at 11 49 33 AM

## Has this been tested? How?

Tested locally. Error not showing in the console after the fix. 

## Types of changes
- Bug fix

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


benhammondmusic (approved):
Awesome!! Thanks so much eric

PR #2258, opened 2023-06-20
Updates frontend for new COVID CDC data to June (data still from April)

benhammondmusic:
- Updates metadata dates to correspond with manually updated CDC_RESTRICTED data on the backend
- adds optional inline topic name with fallback for use mid-sentence (avoiding reliance on `toLowercase()` which doesn't work for things like `COVID-19` that should always be capitalized
- fixes typos / adds clarifying language per @kccrtv 
- rewords to avoid having hard-coded dates in the data source description, this means only `DatasetMetadata.ts` needs to have its dates updated ongoing

FIXES:

Screen Shot 2023-06-20 at 3 34 27 PM
Screen Shot 2023-06-20 at 3 34 52 PM
Screen Shot 2023-06-20 at 3 35 12 PM


kccrtv (approved):
👍🏼

PR #2281, opened 2023-07-14
Frontend: Adds Age-Adjusted HIV Death Ratios

### LINKED ISSUE #2181: Add Age-Adjusted Risk Ratios for HIV Deaths
## Background
Age is a confounding factor when comparing populations; for this reason many places choose to present age-adjusted rates to prevent population age distributions from masking a condition's true effect on a group. For the HET, we have chosen to only present *risk ratios* rather than age-adjusted rates, and present those in a distinct table. the rest of the report maintains the crude rates. 

KFF only calculates and presents age-adjusted rates for HIV Deaths, but not HIV infections or other metrics (and they age adjust across states rather than across age groups). I think to start we should do the same and only pursue age-adjusting for deaths, using the logic as we did for COVID, that older people are more likely to die from HIV than younger people, much like COVID. I'm not sure the science shows that an older person is more likely to contract HIV than a younger person, simply due to their age. 

We only age-adjust COVID deaths and hospitalizations as of now. We need fine-grained data to age adjust a condition; which thankfully ARE available from CDC Atlas for HIV deaths

## Proposed Solution

Following the exact algorithm outlined on our [age-adjustment methodology page](https://healthequitytracker.org/ageadjustment), we should generated age-adjusted risk ratios at the national and state level for the HIV deaths 

## End State 

The age-adjusted table will appear for HIV deaths in addition to the current COVID conditions

Highlighting inequities in health is the entire aim of the tracker; this will bring into focus the impact conditions have on marginalized race groups even beyond the disparity found in the crude rates we already show

## Tasks:
- [x] HIV deaths, by-age, by-race, by-state, by-national are available from CDC
- [x] check in raw CDC table to `/data`
- [x] generate cross-sectional, single-year HET tables on our backend for `by race / by age`
- [x] WONT DO consider generating cross-sectional, multi year HET tables for `by race / by sex` and populating only with the data columns currently used for `Black women`
- [ ] copy the age-adjustment process used by our backend for COVID over to the HIV flow
- [ ] update the frontend age-adjustment card to be more generic and not COVID specific (if needed)
- [ ] consider ways to age-adjust intersectional groups: could we for instance calculate a risk ratio for `Black women compared to White (all sexes)`??
- [ ] consider ways to visualize age-adjustment over time (first idea using our existing time-chart but with a logarithmic y-axis, using `1` as the center point and going up to `x2 x3 x4` and going down to `x1/2 x1/3 x1/4`. COVID age-adjusted ratios over time exist, and we could also generate for HIV deaths

benhammondmusic:
## Description

- national and state HIV deaths age-adjust cards fetch `-with_age_adjust` tables

## Motivation and Context

closes #2181 

## Has this been tested? How?

manually

Age-adjusted risk of HIV deaths compared to White (NH)  in  the United States from Health Equity Tracker Jul 2023
Age-adjusted risk of HIV deaths compared to White (NH)  in Georgia from Health Equity Tracker Jul 2023
Age-adjusted risk of HIV deaths compared to White (NH)  in Louisiana from Health Equity Tracker Jul 2023

## Types of changes

- New content or feature

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
Nice work!

kccrtv (approved):
👍🏼

PR #2308, opened 2023-08-07
RF: migrate some shared state into URL using `jotai-location` and `useParamState` hook

benhammondmusic:
### State Management
- load `locationAtom` from `jotai-location`, which is a util that keeps a piece of state synced with the urlparams, and can be accessed from any file
- makes custom `useParamState` hook that loads that shared locationAtom, and returns a getter/setter just like `useState`. It accepts the type, the string value to use as the URL param key, and an option default param value.
- optimized `useGetParamState` for when we only need to GET that param state, but not set it

### Controls These States With New Hooks
- `demographicType`
- `multimapOpen`
- `topicInfoOpen` 
- `dataTypeId` 1 and 2 

### Renaming
- rename all variables so that the `DemographicType` refers to the type of data disaggregation (like by `age`, `sex`, or `race_and_ethnicity`), and reserve the word `Breakdown` to refer to any dataset that has been "broken down" by demographic, geographic, timeview, etc. 
- `activeBreakdownFilter` -> `focusedGroup`

## Motivation and Context

Refactor code for readability and usability

## Has this been tested? How?

Test still pass

## Types of changes
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


benhammondmusic:
rebuilding piece by piece #2317 #2316

PR #2319, opened 2023-08-11
Refactor: manage `DemographicType` with `useParamState()`

benhammondmusic:
## Description

- [x] replace the multiple useStates() in Report and CompareReport that handled demographic type state with the shared useParamState 
- [x] checked for excessive prop-drilling; probably not helpful to direct access vs props since it's only passed one layer and used in both layers
- [x] fix some of the linter errors caused by tech debt from non-strict types 

## Motivation and Context

refactor to simplify code

## Has this been tested? How?

tests passing

## Types of changes
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
🤘

PR #2324, opened 2023-08-18
Adds numerator / denominator calculation to rate cells of data table

benhammondmusic:
## Description

- fetches the numerator/denominator count metrics if defined in the metricConfig
- adds to the metricQuery from the tableCard
- renders the `( n units / m units)` if both numbers are defined
- removes the plural `s` from the unit if the count is only 1 😎
- swaps rendering logic so only 100k get the "less than 1" override, but not pct_rate
- only apply `#` in component, don't include in the metricConfig labels (more consistent with how unit labels are generated)

## Motivation and Context

better data visibility for users

## Has this been tested? How?

tests passing

Screen Shot 2023-08-18 at 1 24 41 PM

## Types of changes

- New content or feature

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
Awesome work!

PR #2325, opened 2023-08-22
Fix for HIV Stigma charts not displaying

eriwarr:
## Description
- Updates the source year for HIV stigma. 

## Motivation and Context

We currently only have 2021 data for prevalence, diagnoses, PrEP, and linkage to HIV care. Stigma is sourced from 2019. 

## Has this been tested? How?
Tested locally. 

## Screenshots (if approp
_Bug_:
Screenshot 2023-08-22 at 2 37 31 PM
riate):

_Fix_:
Screenshot 2023-08-22 at 2 38 02 PM

## Types of changes
- Bug fix

## Post-merge TODO
I have inspected frontend changes and/or run affected data pipelines:
- [ ] on DEV
- [ ] on PROD

## Preview link below in Netlify comment 😎

PR #2332, opened 2023-08-23
Detect most recent year for provider

eriwarr:
## Description

- Uses `getMostRecentYear` to detect the most recent year for a data set (HIV & CAWP). 
- Adds playwright test to `DatasetTimeUtils.ts `

Note: I will add this to the AHR when API is completed and to BJS once time data is available. 

## Motivation and Context

A bug occurred where HIV stigma charts would not display because the incorrect source year was provided to the `filterTimeView` function. This bug was fixed in #2325  by statically providing the source year.

## Has this been tested? How?
Tested locally. 

## Types of changes
- New content or feature

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


frontend/src/data/utils/DatasetTimeUtils.test.ts
```diff
@@ -206,3 +208,15 @@ describe('Tests getPrettyDate() function', () => {
...
+describe('Tests getMostRecentYearAsString()', () => {
+test('correct year string is returned', async() => {
+  const mockDataFrame = new DataFrame([
+    { time_period: '2019', pct_share_of_women_state_leg: 10 },
+    { time_period: '2017', pct_share_of_women_us_congress: 20 },
+])
+
+const mostRecentYear = getMostRecentYearAsString(mockDataFrame, ['pct_share_of_women_us_congress'])
+expect(mostRecentYear).toEqual('2017')
+
```
benhammondmusic, line 221: this is great. can you expand the test df a little and add a few more years for each metric_id? the test you added shows that it detects the most recent year for the given metricId which is great, i'd love to also see:
- that it still gives '2017' for pct_share_of_women_us_congress when there is a '2016' year
- that it gives '2019' for 'pct_share_of_women_state_leg' 
- that we handle the case where the requested metric id isn't in the df at all? what happens there? 
-

benhammondmusic (changes requested):

frontend/src/data/utils/DatasetTimeUtils.test.ts
eriwarr: Absolutely!

benhammondmusic (approved):
awesome!

PR #2336, opened 2023-08-25
Tweaks `getMostRecentYear()` api; uses for remaining time data sets

benhammondmusic:
## Description

- adjusts function api to accept a specific `metricId` rather than an array and extracting the first inside the function
- adjust function to return `undefined` when the df doesn't have a `time_period` column
- uses function in IncarcerationProvider (time only at county level), and AcsConditionProvider

## Motivation and Context

- handle edge cases in function
- use function across all time based data sets that need to detect most recent year

## Has this been tested? How?

tests adjusted, added, and passing

## Types of changes
- New content or feature
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
🤘🏿

PR #2337, opened 2023-08-25
Adds stronger typing `DatasetId`, `DataSourceId`

benhammondmusic:
### Stronger Typing
- adds new custom types
  -  `DatasetId`: corresponds with the name of the json to look for in the bucket like `'cdc_hiv_data-race_and_ethnicity_national_time_series-with_age_adjust'`. Whenever we add/update our json tables, we will update these type members and then Typescript will yell at us until all the files in the app are updated properly
  - `DatasetIdWithStateFIPSCode` which uses [Typescript's template literal typing](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to ensure that it will be any of the above declared `DatasetId`s followed by a `-` and a two-digit state FIPs code
  - `DataSourceId` the parent source enum like `cdc_atlas`
- uses these new stronger types throughout the app to keep ids in sync across Providers, DatasetMetaData, and MetaDataMap, and DataFetcher/Loader/Caching

### Metadata
- refactors the `DatasetMetaData` file; previously it declared a giant array of objects `datasetMetadataList`, and then iterated over it to embed the related data *source* ids into an exported `DatasetMetadataMap`. Now, instead, we just declare those datasources up front, and instead of a giant array of objects where one of the properties is the dataset id, the `DatasetMetadataMap` is a typescript `Record<DatasetId, DatasetMetadata>`. This helps a lot with ensuring there are no "forgotten" dataset ids because `Record` ensures that every possible member of the first type is present and only associated with a single value

### Fixes/Refactors
- `getDatasetId()` now always returns the pure `DatasetId`, and any addition of a FIPS code to the end (to fetch the smaller, state-specific county-level file as needed) is done after the `getDatasetId()` and before the actual fetch of the data. Tests have been updated to expect the non-fips version, and to append the fips if needed inside the test function. 
- some tests were flaking and I couldn't figure it out so i adjusted them to only test the `getDatasetId()` instead of the entire mocked query/response
- updates new vaccination tables names that should have been done in #2333 and now were caught because of the type checking
- updates geo context table names
- adds hiv black women test datasetid
- improves naming in private DataFetcher methods
- improves provider `if` tree to be consistent and to always return a `DatasetId` and to always use the same structure of nested `if` statements
- replaces all instances of a datasetid previously typed as `string` with either `DatasetId` (or `DatasetId | DatasetIdWithStateFIPSCode` if it could possibly be the more specific file)

## Motivation and Context

Having a strongly typed `DatasetId` will prevent a lot of the bugs that occur when adding/adjusting data and the ids used for the Provider files become out of sync with the Metadata (used for the sources in the footer and also the data sources downloads page)

finally closes #1684 

## Has this been tested? How?

tests added and passing

## Types of changes
- Bug fix
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
Nice work!

PR #2340, opened 2023-08-29
Improves the multi map card layout

eriwarr:
## Description

- bulleted high-level changes

## Types of changes
- Bug fix
- New content or feature
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD

### DIFF
--- a/frontend/src/cards/ui/MultiMapDialog.tsx
+++ b/frontend/src/cards/ui/MultiMapDialog.tsx
@@ -17,7 +17,6 @@ import {
   type FieldRange,
 } from '../../data/utils/DatasetTypes'
 import { type MetricConfig } from '../../data/config/MetricConfig'
-import { Sources } from './Sources'
 import styles from './MultiMapDialog.module.scss'
 import {
   type MetricQuery,
@@ -39,6 +38,8 @@ import TerritoryCircles from './TerritoryCircles'
 import MapBreadcrumbs from './MapBreadcrumbs'
 import { type CountColsMap } from '../MapCard'
 import { RATE_MAP_SCALE } from '../../charts/mapGlobals'
+import CardOptionsMenu from './CardOptionsMenu'
+import { ScrollableHashId } from '../../utils/hooks/useStepObserver'
 
 export interface MultiMapDialogProps {
   // Metric the small maps will evaluate
@@ -75,6 +76,8 @@ export interface MultiMapDialogProps {
   totalPopulationPhrase: string
   handleMapGroupClick: (_: any, newGroup: DemographicGroup) => void
   pageIsSmall: boolean
+  reportTitle: string
+  scrollToHash: ScrollableHashId
 }
 
 /*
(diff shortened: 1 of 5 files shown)


frontend/src/cards/ui/MultiMapDialog.tsx
```diff
@@ -134,25 +142,9 @@ export function MultiMapDialog(props: MultiMapDialogProps) {
...
-                onClick={props.handleClose}
-                color="primary"
-              >
-                <CloseIcon />
-              </Button>
-            </Grid>
             {/* Modal Title */}
             <Grid item xs={12} sm={9} md={10}>
-              <Typography id="modalTitle" variant="h6" component="h2">
+              <Typography id="modalTitle" variant="h6" component="h2" lineHeight={"25px"}>
```
benhammondmusic, line 147: can we use the sass variables to stay more consistent across components? general rule of thumb we shouldn't really have any hard coded px values per item if we can avoid it. you can import variables.module.scss and then use `lineHeight={sass.lhLoose}` which should be pretty close to 25px, or if you need to just make a new variable setting like `lhModalHeading: 25px` and use that

frontend/src/cards/ui/MultiMapDialog.tsx
```diff
@@ -126,6 +129,11 @@ export function MultiMapDialog(props: MultiMapDialogProps) {
       ref={screenshotTargetRef}
     >
       <DialogContent dividers={true}>
+      <CardOptionsMenu
```
benhammondmusic, line 132: probably dont need to fix for this PR, but interestingly the screenshot doesn't do the entire div, only the section that is visible 

image

frontend/src/cards/ui/MultiMapDialog.tsx
```diff
@@ -348,13 +339,15 @@ export function MultiMapDialog(props: MultiMapDialogProps) {
 
       {/* MODAL FOOTER */}
       <footer>
+        {/* mobile-only close button */}
```
benhammondmusic, line 342: I think we should keep the "close" button in the footer for both mobile and desktop, and eliminate this top right "X" button entirely. do you agree? @kccrtv ?

Screen Shot 2023-08-29 at 3 32 36 PM

benhammondmusic: Also, should we try to keep the sources and maybe even the notes if it multimap on desktop? only hide those on mobile?

eriwarr: I forgot to add the mobile prop for the sources, fixing that now.

frontend/src/cards/ui/MultiMapDialog.tsx
eriwarr: Yeah, I noticed this too. I'll create an issue for it.

benhammondmusic (approved):
thanks @eriwarr looks awesome

Issue #2033, opened 2023-03-06
Vectorize `pct_share` util function if possible


benhammondmusic:
## Is your feature request related to a problem? Please describe.

From reading it seems that `df.apply()` should be avoiding whenever possible. Rather than creating a lambda or helper fn and applying that across all rows of a df, it's more performant to use panda's built-in functions. An example of the "vectorized" way would be `df["sum"] = df["a"] + df["b"]`. 

## Describe the solution you'd like

We currently use several applies to calculate our pct_shares. Some of these are easily replaced, some other may be trickier like replicating the "avoid_rounding_to_zero" functionality whereby we round to a single decimal place unless that would be `0` and then try rounding to 2 decimal places (to better represent small-population groups). 

I think this can be done like this:

`df[a] = np.where(df[a] < 0.05, df[a].round(2), df[a].round(1))` but we need to investigate further. In quick testing it seemed to leave trailing zeroes on the 1-digit rounded values; this might be ok though if the frontend drops those zeros?

PR #2342, opened 2023-08-30
Backend: Speed up `cdc_restricted_local` step

benhammondmusic:
### Speed Optimizations 

Significantly speeds up the local step that processes the CDC's restricted, raw, unzipped .csv files into the non-restricted, aggregated .csv files needed for the `cdc_restricted` ingestion step.

To benchmark, I ran the code against just a single raw csv file (instead of all ~20). 

- The original code took 313s to process the one file

Here are the incremental changes I made, and the improvement from each step:
- Using vectorized `combine_race_eth()` to use pandas built in optimized functions rather than mapping/applying a lambda function against each row, and only reading in the needed columns with the arg `usecols` in the `read_csv`: 191s
- Above + Removing unneeded 2nd quote cleaning: 188s
- Above + Removing unneeded 1st quote cleaning: 180s
- Above + chunk_size = 1 million (using chunk size splits the df into chunks so your machine doesn't have to hold the entire thing in memory; however my computer is fast enough to hold the entire thing so it's faster to have a giant chunk size that reduces the need to iterate over chunks (but also would kick in if cdc ever sends a much bigger file along): 75s
- Above + chunk_size = 2 million: 64s
- Above + chunk_size = 5 million: 57s

When running against the full set of all raw files, the run time went from 2h40m down to 34m, and produced identical output files.

### Updates
- specifies `numeric_only` to prevent the futurewarning 

### Abandoned Optimizations

I tried several more vectorization optimizations, replacing the original code's `map` and `applymap` and `apply` with the Pandas vectorized version, but saw no improvement on those. For example replacing:

```
df[COUNTY_FIPS_COL] = df[COUNTY_FIPS_COL].map(lambda x: x.zfill(5) if len(x) > 0 else x)
```

with

```
df[COUNTY_FIPS_COL] = df[COUNTY_FIPS_COL].str.zfill(5)
```

Although Pandas vectorized methods are usually much faster than using `apply`, in these cases they were not and it's probably because the `.str.` vectorized methods are not super optimzed, and sometimes are SLOWER. So going forward we should prioritize refactoring our python code with these vectorized versions (like #2033 ) only when it's doing numerical work and not necessarily with string based work.

## Motivation and Context

It was taking hours to run the local code, extremely slow and frustrating and locked up my machine for the duration

## Has this been tested? How?

First I ran the original code and generated the HET tables with a `old_` prefix, and then switched to this branch and ran the updated code to generate the tables without that prefix. I than used a quick bash script to iterate over the tables and compare the old and new versions of the same table and ensure that the matching files were identical between `main` branch and this branch. I ensured this checker was working by manually diffing age_county against age_state and seeing that it did in fact show a difference

The original unit tests continue to pass with all updates

## Types of changes
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


python/datasources/cdc_restricted_local.py
```diff
@@ -126,18 +140,26 @@ def combine_race_eth(df):
...
 
-        else:
-            return RACE_NAMES_MAPPING[row[RACE_COL]]
+    # Create a mask for other cases
+    other_mask = ~race_missing_mask & ~eth_missing_mask
+
+    # Create a new combined race/eth column Initialize with UNKNOWN
+    df[RACE_ETH_COL] = std_col.Race.UNKNOWN.value
+    # Overwrite specific race if given
+    df.loc[other_mask, RACE_ETH_COL] = df.loc[other_mask, RACE_COL].map(RACE_NAMES_MAPPING)
```
eriwarr, line 156: Nice!

eriwarr (approved):
This is awesome! Great job.

Issue #1680, opened 2022-07-15
Identify state that we could pass via useContext() instead of prop drilling


benhammondmusic:
**Is your feature request related to a problem? Please describe.**

As the app complexity grows, we have several layers deep of nested components. Since we aren't using any sort of state management library for the UI settings, we rely on prop drilling and/or URL params as a way of passing things around. React's useContext() could be used to simplify this process; for example passing the currently selected condition and data type(s) and geographies(s) around to the many components who rely on that information. 

**Describe the solution you'd like**
- Identify what state would benefit from implementation of useContext()
- Replace prop drilling with this context

PR #2359, opened 2023-09-05
Fix screenshot capture of visible div section

### LINKED ISSUE #2343: Screenshot only captures visible section vs div
probably dont need to fix for this PR, but interestingly the screenshot doesn't do the entire div, only the section that is visible 

image

_Originally posted by @benhammondmusic in https://github.com/SatcherInstitute/health-equity-tracker/pull/2340#discussion_r1309373971_

eriwarr:
## Description

This PR fixes the issue where only the viewable portion of the div is saved to the screenshot when viewing the multi-map.

- [x] Types the elementsToHide prop. 
- [x] Uses the typed param key variables instead of the ID strings (will add the alt table views to params in a separate PR). 
- [x] Adds a ref to the footer when viewing the multi-map (the ref for the content cannot be placed on the modal itself, so two refs are needed to capture the content and footer info separately). 

## Motivation and Context

Fixes #2343.

## Has this been tested? How?
Tested locally.

_Desktop View_
HIV prevalence in the United States across all race and ethnicity groups from Health Equity Tracker Sep 2023

_Mobile View_
HIV prevalence in the United States across all race and ethnicity groups from Health Equity Tracker Sep 2023 (16)

## Types of changes
- Bug fix
- Refactor/chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD

### DIFF
--- a/frontend/src/utils/hooks/useDownloadCardImage.tsx
+++ b/frontend/src/utils/hooks/useDownloadCardImage.tsx
@@ -31,12 +31,14 @@ export type HiddenElements =
   | '#card-options-menu'
   | '#download-card-image-button'
   | '#map-group-dropdown'
+  | '#multi-map-close-button'
 
 export function useDownloadCardImage(
   cardTitle: string,
   hiddenElements: HiddenElements[] = [],
-  scrollToHash: ScrollableHashId = 'rate-map',
-  dropdownOpen?: boolean
+  scrollToHash: ScrollableHashId,
+  dropdownOpen?: boolean,
+  footerContentRef?: React.RefObject<HTMLDivElement>
 ) {
   const screenshotTargetRef = createRef<HTMLDivElement>()
   const [dropdownElement, setDropdownElement] = useState<HTMLElement>()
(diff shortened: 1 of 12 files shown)


frontend/src/utils/urlutils.tsx
```diff
@@ -38,6 +38,8 @@ export const MULTIPLE_MAPS_1_PARAM_KEY = 'multiple-maps'
 export const MULTIPLE_MAPS_2_PARAM_KEY = 'multiple-maps2'
 export const HIGHEST_LOWEST_GEOS_1_PARAM_KEY = 'highest-lowest-geos'
 export const HIGHEST_LOWEST_GEOS_2_PARAM_KEY = 'highest-lowest-geos2'
+export const ALT_TABLE_VIEW_1_PARAM_KEY = 'alt-table-view1'
```
benhammondmusic, line 41: technically these shouldn't really be `_PARAM_KEY` yet because they aren't used in the url search params to record the open/closed state of the alt table... but im fine leaving them like this because we should move that state into the url bar anyways as part of #1680

frontend/src/utils/hooks/useDownloadCardImage.tsx
```diff
@@ -20,11 +27,18 @@ const BOTTOM_PADDING = 120
 const URL_FONT_SIZE = 14
 const URL_FONT_STYLE = '"Inter",sans-serif'
 
+export type HiddenElements =
```
benhammondmusic, line 30: this is great! What do you think about using something like `ElementHashIdsHiddenOnScreenshot` ? I know it's super long but i'd like to use `HashId` in the name so it's clear it's a hashed id (and not like a ref to an html el or something) and I'd like it to be clear to future devs the elements _can_ be hidden, but aren't normally hidden

benhammondmusic (commented):
awesome! thanks so much man!

frontend/src/utils/hooks/useDownloadCardImage.tsx
eriwarr: Sounds good to me ☺️

benhammondmusic (approved):
Looks good!

PR #2360, opened 2023-09-05
Make optional metric configs: `pct_share` and `pct_share_unknown`

benhammondmusic:
## Description

- Makes the `pct_share` metric optional since several of our topics can't present this (like voter participation) or it is unwanted (like the PHRMA adherence measures where the pct_share metric is too confusing and not really helpful)
- Make a new optional metric `pct_share_unknown` for cases like PHRMA adherence where it is helpful to see the share of cases with unknown demographic info, but we dont want to see the full pct_share breakdown on the comparison bar chart or in the table
- the comparison bar chart card requires the `pct_share` metric to render
- the unknown map card requires either the `pct_share` or preferably the `pct_share_unknown` to render
- the table card share column requires the `pct_share`
- updates util functions (and tests) to handle undefined metrics

## Motivation and Context

Rather than maintaining manual lists of "should show" "should hide" per topic or source, we should rely on the metric config (and its presense/absense of metrics) to determine what shape the report should take

## Has this been tested? How?

tests pass; optional cases handled

Screen Shot 2023-09-05 at 1 03 23 PM

## Types of changes
- New content or feature
- Refactor / chore


frontend/src/data/utils/DatasetTimeUtils.test.ts
```diff
@@ -141,6 +141,13 @@ describe('Tests for nesting functions', () => {
       getNestedUnknowns(twoYearsOfUnknownsFromNormalData, 'jail_pct_share')
     ).toEqual(expectedNestedUnknowns)
   })
+
+  test('getNestedUnknowns() on missing metricId returns []', () => {
+    expect(getNestedUnknowns(twoYearsOfNormalData, undefined)).toEqual([])
+  })
+  test('getNestedUnknowns() on missing metricId returns []', () => {
+    expect(getNestedUnknowns(twoYearsOfNormalData, undefined)).toEqual([])
+  })
```
eriwarr, line 150: The test is written twice.

eriwarr (approved):
Awesome!

Issue #2361, opened 2023-09-06
Split sources with time data into `_historical` and `_current`


benhammondmusic:
## Background

Initially, the tables only stored a single time_period of data (no column was even needed), and only shipped at most a rate, pct_share, and pop_share column. 

## What's Missing / Problems with the current approach

As our interface expands, the .json tables supplying that data continue to grow. We now have many more conditions, by demography, by geography, sometimes by time, and we offer multiple metrics including per 100k, pct_share, pct_rel_inequity. Now, we are starting to incorporate the "N" counts as well, which means every condition will get a numerator and denominator column. This swells the .json files quite a bit, as an example the Texas-only county level HIV file by race is 23Mb on disk and zips for transfer only down to 641kb which is big and introduces noticeably slow loading.

We have also discussed calculating the "highest/lowest" groups per FIPS on the backend to speed up processing; this is additional data that is only needed (at this point) on the `_current` cards like the Map Card, but not needed over time. 

## Proposed Solution
If we split our tables into two types (similar to how we do for COVID already), we can restrict what each type shows so it's only shipping relevant data. 

- _`_historical`:_ have many more rows since the information is aggregated per year. However, we can ship far fewer columns since we don't show the N counts, the pct_share / pop_pct figures. For `_historical` we really only need: time_period, geography, demography, rate (per100k), pct_rel_inequity, and pct_share (only for showing the unknown bubbles over time so leave off for data sets without any unknown demographics)
- _`_current`:_ have far fewer rows (only presenting the most recent year of data typically), and only need geography, demography, rate (per100k), pct_share, pop_share, n counts (we don't need pct_rel_inequity, at least not yet)

## End State 

- ***MAP, SIMPLE BAR, UNKNOWNS MAP, DISPARITY BAR, DATA TABLE*** will read the `_current` tables, and have access to only the relevant single year of data, with all the "bonus" data like N's, and potentially highest/lowest groups and other extra columns that are only used on single-year cards
- ***RATE TRENDS, SHARE TRENDS*** will read the `_historical` tables, and will have access to every available year of data, but will only have the needed data columns (rate, pct_rel_inequity)

## How will this move the needle on health equity?

We need to continuously work on keeping the speed of the tracker as fast as possible, particularly on low-powered devices, to ensure that accessibility doesn't suffer as we add new features that swell the loading times

## Sources with time data to split and add "N"s (rate numerator and denominator) when possible:
- [x] HIV
- [x] CAWP
[...]

PR #2362, opened 2023-09-06
BACKEND: Split hiv into `_historical` and `_current`

benhammondmusic:
### `data/`

- updates some stigma data that didnt get updated in #2373 

### `cdc_hiv`

- adds new function in HIV for building the big query column types based on the particular breakdown and which columns to keep for each type of table and breakdown. this is huge and ugly, but i think easier to understand than having multiple nested list and dict comprehensions
- uses new util function in #2367 to get "current" version of df for that write
- new HIV data all has 2021 as the most recent year, allowing our new function to work as expected
- reduces columns to match the output of the new function AFTER the breakdown function, so instead of building the "cols_to_keep" inside the breakdown function and dropping before it returns, it just builds the big version of the table, possible with some straggler columns, and then outside of the breakdown function it calculates the big query DTYPES, and then uses those keys as the needed columns (dropping everything else). 
- uses the new term `_historical` to name tables that include time data; this is because our unsplit time-based sources are already using the `time_series` suffix, which makes it much harder to smoothly shift over our production frontend/backend connections. Using `_historical`, we can allow our frontend to continue looking for the older saved `_time_period` tables, then we can generate the new `_historical` tables on prod, and finally we can push the changes to the frontend so it looks for the new prefix instead of the old

### `cdc_hiv_age_adjust`

- the age-adjustment for hiv deaths will accept the new `_current` table as input, and output `_current-age-adjust` with the age adjust ratio column appended
- table names updated
- test files updated
- adjusted dag order to run age-adjust LAST as it needs access to the new _current files that don't exist yet

### Refactoring

- adds new constants for `CURRENT` and `HISTORICAL`
- any touched python files will include auto-formatting from black
- disabled the markdown linting for now; our README is a mess syntax wise and failed the linter multiple ways. we should fix this in a distinct PR

## Motivation and Context

#2361 for HIV

## Has this been tested? How?

tests updated and added (you can see additional columns dropped from the golden_data files which is what we want; smaller files overall)

Screen Shot 2023-09-13 at 4 43 06 PM
Screen Shot 2023-09-13 at 11 17 23 AM
Screen Shot 2023-09-13 at 11 20 10 AM
Screen Shot 2023-09-13 at 11 23 00 AM

## Types of changes
- New content or feature
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


python/datasources/cdc_hiv.py
```diff
@@ -528,12 +578,157 @@ def generate_atlas_cols_to_exclude(breakdown: str):
...
+    elif breakdown not in ["all", std_col.BLACK_WOMEN, "black_women_all"]:
         atlas_cols.extend(
-            filter(lambda x: x != DEM_COLS_STANDARD[breakdown], CDC_DEM_COLS))
+            filter(lambda x: x != DEM_COLS_STANDARD[breakdown], CDC_DEM_COLS)
+        )
 
     return atlas_cols
+
+
+def get_bq_col_types(demo, geo, table_type):
```
eriwarr, line 593: Nice!

python/datasources/age_adjust_cdc_hiv.py
```diff
@@ -101,63 +107,67 @@ def generate_age_adjustment(self, geo):
...
     age_adjusted_df = age_adjusted_df.reset_index(drop=True)
 
+    # print("df")
+    # print(df.columns)
+    # print(df)
+
+    # print("age_adjusted_df")
+    # print(age_adjusted_df.columns)
+    # print(age_adjusted_df)
+
```
eriwarr, line 141: Intended to leave this?

eriwarr (approved):
Looks great! Thanks Ben!

PR #2363, opened 2023-09-07
BACKEND: Refactor HIV tests

benhammondmusic:
## Description

- now that we are passing the demo and geo as args from airflow into cdc_hiv write_to_bq, we can test write_to_bq for a single breakdown, and dont need to test the intermediate step in the generate_breakdown method
- as part of #2362 there will need to be df manipulation done to reduce the filesize of each table, and that will be done AFTER the breakdown method but before the actual write to bq; as such it's much easier to test the final output, rather than feeding a mocked generated breakdown as input to test of write_to_bq

## Types of changes
- Bug fix
- New content or feature
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
This is awesome! Thanks, Ben!

PR #2372, opened 2023-09-11
Fixing rendering issue in `useResponsiveWidth` hook

eriwarr:
## Description
This PR addresses an issue where specific visualizations fail to render correctly when users refresh the page. This behavior is unpredictable as it only occurs sometimes when the user refreshes the page or navigates to a new visualization and only affects some visualizations (viz that are not lazy-loaded). 

The issue occurs when the hook attempts to access the ref before the component it uses has finished rendering (returning 0 as the width) or if the Vega viz attempts to load or render before the ref's width has been updated from its default value. 

- Added the `useResponsiveWidth()` hook to the D3 viz. 
- Added the `forceUpdate()` function to ensure that the width state stays updated in response to changes in `defaultWidth` or window resizing.

## Motivation and Context

A refreshed page should provide users with the same information and visualizations they had before the refresh.

## Has this been tested? How?
Tested locally.

_Bug_
Screenshot 2023-09-12 at 10 17 52 AM

## Types of changes
- Bug fix

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD

### DIFF
--- a/frontend/src/charts/trendsChart/Index.tsx
+++ b/frontend/src/charts/trendsChart/Index.tsx
@@ -41,6 +41,7 @@ import useEscape from '../../utils/hooks/useEscape'
 import { getMinMaxGroups } from '../../data/utils/DatasetTimeUtils'
 import { type DemographicGroup } from '../../data/utils/Constants'
 import ChartTitle from '../../cards/ChartTitle'
+import { useResponsiveWidth } from '../../utils/hooks/useResponsiveWidth'
 
 /* Define type interface */
 export interface TrendsChartProps {
@@ -74,8 +75,6 @@ export function TrendsChart({
   const { groupLabel } = axisConfig ?? {}
 
   /* Refs */
-  // parent container ref - used for setting svg width
-  const containerRef = useRef(null)
   // tooltip wrapper ref
   const toolTipRef = useRef(null)
 
@@ -92,10 +91,8 @@ export function TrendsChart({
     useState<DemographicGroup[]>(defaultGroups)
 
   // manages dynamic svg width
-  const [[width, isMobile], setWidth] = useState<[number, boolean]>([
-    STARTING_WIDTH,
-    false,
-  ])
+  const [containerRef, width] = useResponsiveWidth(STARTING_WIDTH)
+  const [isMobile, setIsMobile] = useState<boolean>(false)
 
   const isCompareMode = window.location.href.includes('compare')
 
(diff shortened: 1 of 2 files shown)


frontend/src/charts/trendsChart/Index.tsx
```diff
@@ -274,48 +267,29 @@ export function TrendsChart({
...
-                isSkinny={isSkinny}
-                chartWidth={width}
-                demographicType={demographicType}
-                legendId={`legend-filter-label-${axisConfig.type}-${
-                  isCompareCard ? '2' : '1'
-                }`}
-              />
-            )}
-          </>
+        {!isMobile && (
```
benhammondmusic, line 270: nice refactor; much easier to read

frontend/src/utils/hooks/useResponsiveWidth.tsx
```diff
@@ -7,23 +7,29 @@ export function useResponsiveWidth(
   // Initial spec state is set in useEffect when default geo is set
   const ref = useRef<HTMLDivElement>(document.createElement('div'))
 
+  const forceUpdate = () => {
+    setWidth((prevWidth) => prevWidth + 1)
+  }
+
```
benhammondmusic, line 13: I don't think this is really doing anything is it? You're defining this function but then it doesn't call it anywhere. It's listed as a side effect dependency in `useEffect` array, but that just means "run this useEffect is the function `forceUpdate()` is redefined" which is impossible anyways since it's defined with a `const`. Did I miss anything here ?

frontend/src/utils/hooks/useResponsiveWidth.tsx
```diff
@@ -7,23 +7,29 @@ export function useResponsiveWidth(
...
-    if (ref?.current) {
-      setWidth(ref.current.offsetWidth)
-    }
+    const element = ref.current
 
     const handleResize = () => {
-      if (ref?.current) {
-        setWidth(ref.current.offsetWidth)
+      if (element) {
+        const newWidth = element.offsetWidth || defaultWidth
```
benhammondmusic, line 19: should we use `??` instead of `||` for this short circuit to prevent newWidth from setting to `0` if that's the offset width? or do want `0` to be possible?

frontend/src/utils/hooks/useResponsiveWidth.tsx
eriwarr: I'll make an issue for this for now.

benhammondmusic (approved):

PR #2375, opened 2023-09-14
Upgrade GitHub actions and Node.js to v19

eriwarr:
## Description

- Pins the GitHub actions to node 19. 
- Sets the node engine to 19 in `package.json.`

## Motivation and Context

Node 16 was moved to the end of life on September 11th, 2023.

## Types of changes
- Refactor/chore


.github/workflows/node.js.yml
benhammondmusic, line 1: there are a few other spots in this action where it specifies `16` that need to be bumped

benhammondmusic (approved):

PR #2396, opened 2023-09-22
Optimize map topojson file

benhammondmusic:
### Simplify `geographies.json`
- download new topojson file from [topojson github repo](https://github.com/topojson/us-atlas#counties-10m.json)
- use [mapshaper](https://mapshaper.org/) tool to adjust the topojson fields and associated "arcs"
- remove `nation` property; keep only `states` and `counties` since we don't show any "summary" style maps nationally (and the code isn't configured to do so right now)
- simplify the map to reduce edge complexity. this results in less fidelity for coastal / islandy areas but still looks basically the same, and is especially fine since we don't allow users to zoom in on our maps

### Measured Improvements
- Reduced size of compressed data transferred from 344kb → 229kb = **33% reduction for transfer size of topojson file**
- Average speed for the fetch of `geographies.json` from ~300-400ms → ~100-200ms = **2x to 3x faster for that individual fetch request**

## Motivation and Context

loading the `geographies.json` file is a giant file and slow

## Has this been tested? How?

manually checked island area county equivalent maps; and an assortment of state and county reports. everything loads as expected, and loads slightly faster

Screen Shot 2023-09-22 at 10 20 48 AM

## Types of changes
- Refactor / chore

## Post-merge TODO
- [ ] on DEV
- [ ] on PROD


eriwarr (approved):
💪🏿

PR #2421, opened 2023-10-03
Memoize `getHighestLowestGroupsByFips()`; remove unneeded alt text items from Maps

benhammondmusic:
### Memoize `getHighestLowestGroupsByFips()`
- Run `getHighestLowestGroupsByFips()` only once in `<MapCard>` with `useMemo()`, to prevent it from recalculating unnecessarily 
- Pass the result via props to `<MultiMap>` as needed, rather than recalculating inside multiple components

### Clean up
- adds key missing in `<CitationInfo>`
- adds `aria-hidden={true}` to `<TerritoryCircles>` as the screen reader already gets territory info inside the normal Vega map, so we want the SR to skip the circles entirely
- prevents the territory maps from generating multiple sets of invisible alt text items for every child FIPS which vastly reduces the number of DOM elements

## Motivation and Context

Cleanup, Speedup

## Has this been tested? How?

Comparing netlify deploys, this PR:
- reduces the number of DOM elements from >6k to >3k
- seems to reduce the javascript thread and execution times

Screen Shot 2023-10-02 at 11 53 47 PM

## Types of changes
- Refactor / chore


eriwarr (approved):
🤘🏿

PR #2423, opened 2023-10-03
Refactor: Improved Madlib FIPS handling

benhammondmusic:
## Description

- extracted logic from `failInvalidFips()` for use in `isFipsString()` within `<TopicOrLocationSelector>`
- optimizes the placement of dropdownIds into their categories; before I was inefficiently using nested loops iterating over each category and then over ALL of the dropdown options and only rendering if the option "belonged" to the current category. Now, instead it just iterates over each category, with a nested loop over only the topics that belong in that category

## Motivation and Context

cleanup; slightly optimize the category/topic rendering

## Has this been tested? How?

working locally and on netlify

## Types of changes
- Refactor / chore


frontend/src/pages/ExploreData/TopicOrLocationSelector.tsx
```diff
@@ -198,30 +209,24 @@ function TopicOrLocationSelector(props: {
...
+                                  props.onOptionUpdate(optionId)
+                                }}
+                              >
+                                <ListItemText
+                                  className={styles.ListItemText}
+                                  primary={DROPDOWN_TOPIC_MAP[optionId]}
+                                />
+                              </ListItemButton>
+                            )
+                          })}
```
eriwarr, line 229: So much cleaner!

eriwarr (approved):

PR #2451, opened 2023-10-25
Fix frontend timeview

benhammondmusic:
## Description

- removes conditionals to only send `current` for CAWP; we should always send
- explicitly sets `timeView` for all cards' MetricQueries 
- fixes error in logic for when to create frontend current subset df

## Motivation and Context

bug! 

## Has this been tested? How?

working locally across all checked topics

BUG in PROD
Screen Shot 2023-10-25 at 11 43 10 AM

## Types of changes
- Bug fix


kccrtv (approved):
👍🏼

eriwarr (approved):

PR #2592, opened 2023-11-15
Readme quickstart

benhammondmusic:
# Description and Motivation

- updates the README to include a frontend quick-start with the bare minimum to get someone started
- fixes linting on the rest of the readme

## Preview at this link

https://github.com/SatcherInstitute/health-equity-tracker/blob/8a64009fd71db560653744a3ab694ab3d6279db7/README.md

## Types of changes

(leave all that apply)

- New content or feature


JDemlow (approved):
This looks really great! I really like that you can copy the terminal commands as well. If I were going to add anything, it would be a fifth step for something like:

5. **Type `git remote -v` to confirm your remote and origin are set up as expected.**, and then maybe include an example of what that output would look like.

PR #2596, opened 2023-11-20
Update frontend for CDC Covid nov 2023 data

benhammondmusic:
# Description and Motivation

- updates data source date info for recent data already added manually to the pipeline
- adds production e2e test that will expect the new date

## Has this been tested? How?

test fails as expected (prod is still showing old month)

- New content or feature


kccrtv (approved):
👍🏼

PR #2499, opened 2023-11-09
Splits downloads and methodology

benhammondmusic:
## Description

- split nav links to "old" methodology page (includes old methodology tab and old age adjustment tab) and data downloads page
- tailwind refactor all associated pages
- make component for horizontal rules with heading title in the middle

### Tweaks
- adjusts NewsPage blog thumbnails and loading indicators to do 3 columns on widest screens, two normal screens, and 1 on mobile. match number of cols between actual articles and loading indicators

## Motivation and Context

breaking out a portion of #2485 to make it more manageable

## Has this been tested? How?

working locally

## Screenshots (if appropriate):

Screen Shot 2023-11-08 at 11 16 58 PM
Screen Shot 2023-11-08 at 11 17 03 PM

## Types of changes
- New content or feature
- Refactor / chore

## Preview link below in Netlify comment 😎

PR #2485, opened 2023-11-01
Overhauls Methodology

benhammondmusic:
### Adds new methodology designs on hidden route `/new-methodology`

- renames existing methodology content as "old" methodology and still keeps it user-facing for now
- names @kccrtv 's content as "new" methodology which is only accessible by manually visiting `/new-methodology` for now
- this will allow us to maintain the new code on `main` without having to keep a whole 2nd feature branch up-to-date and dealing with merge conflicts. 
- adds new tailwind width/font as needed
- content is not ready for release, but will be hidden until we are ready to switch the routing over from "old" to "new"
- adjusts e2e tests
- all Kim's styling has been initially migrated to Tailwind; there has definitely been styling lost or messed up in the process so we need to confirm everything before we can switch over to this. we also need to proof-read all new content 

## Motivation and Context

closes #2389 
closes #2355 
closes #2354 
closes #2347 
closes #2287 
closes #1706

## Has this been tested? How?

tests passing
new content is NOT ready for public yet

## Screenshots (if appropriate):

`/methodology` still goes to the old page
`/new-methodology` goes to the new content but it not ready nor shown to the user unless you manually go to that endpoint

Screen Shot 2023-11-30 at 3 20 15 PM
Screen Shot 2023-11-30 at 3 20 36 PM

## Types of changes
- New content or feature

## Preview link below in Netlify comment 😎

PR #2610, opened 2023-11-30
Fix DataCatalog tests

benhammondmusic:
# Description and Motivation

Split out from #2485  

In #2499 the unit tests were removed that ensured the DataCatalog functionality was working based on the params and incoming props. This PR:

- downgrades the testing jest package to fix missing types (per stack overflow solution)
- fixes up the logic of reading filtered data sources; now it will only use the params (not offering options props anymore)
- reimplements the tests that were formerly called data_explorer

## Has this been tested? How?

tests are passing now, types are cleared up, functionality works to click through from a exploredata/ card footer and see only that particular dataset

## Types of changes

(leave all that apply)

- Refactor / chore


kccrtv (approved):
👍🏼

PR #2632, opened 2023-12-04
BACKEND: CDC WISQARS

eriwarr:
# Description and Motivation

- Adds national and state-level fatal and non-fatal data on gun violence. 
- The race categories contain races with Hispanic, non-Hispanic, and unknown ethnicities. 
- The dataset contains data where values are suppressed, indicated by a `--` or `{some_value}**` . The latter requires special filtering that can not be done through the `pd.read_csv` function. 
- Function used to detect unknowns as multiple breakdowns and geo_level with unknown values. 

## Has this been tested? How?
Tested locally

## Types of changes
- New content or feature


benhammondmusic (approved):
lets merge this in and then adjust as we decide what adjustments we want to make to the calculations (like adding youth-specific by race, or combining multiple subtopics, etc)

PR #2643, opened 2023-12-06
Refactors all MUI `<CardContent>` components

benhammondmusic:
# Description and Motivation

- removes all MUI `<CardContent>` elements
- uses tailwind where needed for spacing
- uses react fragments `<></>` where needed for grouping
- makes all alerts consistent sizes and spacings by default, but allows overriding with custom values as well
- improves styles of multimap topic info notice

## Has this been tested? How?

visually inspected at various breakpoints

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
Not sure why the test is failing; everything looks right

PR #2662, opened 2023-12-20
SC/Refactor: tw sidebar

### LINKED ISSUE #2527: RF Sidebar.module.scss to Tailwind
Refactor Tailwind for the styles in Sidebar.module.scss

colespen:
# Description and Motivation
- refactor scss modules to Tailwind utility classes
- remove scss modules
- convert MUI components (`<Card/>`) to html (`<div/>`)
- closes #2527 

## Has this been tested? How?
- static, unit, `e2e-local` tests all pass

exact UI match:
Screenshot 2023-12-20 at 4 40 58 PM

## Types of changes

***refactor/chore***


frontend/src/pages/ui/Sidebar.tsx
```diff
@@ -34,12 +32,14 @@ export default function Sidebar(props: SidebarProps) {
 
   return (
     <>
-      <div className={styles.StickySidebarBox} style={{ top: tocOffset }}>
-        <Card className={styles.SidebarTopicInfoButtonBox} raised={true}>
+      {/* @BEN: is it acceptable to combine tw utility classes and inline style={{}} prop (for dynamic vals)? */}
+      <div className='sticky' style={{ top: tocOffset }}>
```
benhammondmusic, line 36: nice. yes, this looks great and is pretty clear what it's doing. I guess we could do some weird tailwind thing instead but i like this. can you remove this comment and keep the code as is? nice work!

frontend/src/pages/ui/Sidebar.tsx
```diff
@@ -34,12 +32,14 @@ export default function Sidebar(props: SidebarProps) {
...
+        <div className='rounded-sm bg-white shadow-raised md:m-cardGutter md:flex md:w-[90%] md:flex-col md:justify-center md:p-2'>
           <TopicInfoModalButton />
-        </Card>
+        </div>
+
         <div className='mode-selector-box'>
-          <Card raised={true} className={styles.SidebarModeSelectorBox}>
+          <div className='rounded-sm bg-white shadow-raised md:m-cardGutter md:flex md:w-[90%] md:flex-col md:justify-center md:px-2 md:py-4'>
```
benhammondmusic, line 42: we actually have a 90% width defined in the tailwind.config as it's used elsewhere, can you swap it out for `md:w-[90%]` `md:w-90p` ? The idea is that once we have gone through the whole app, we can see in the config which percentage widths and other design tokens we are using, and eventually standardize them for more visual consistency.

frontend/src/pages/ui/Sidebar.tsx
```diff
@@ -53,15 +53,15 @@ export default function Sidebar(props: SidebarProps) {
               selected={props.trackerMode}
               setSelected={props.setTrackerMode}
             />
-          </Card>
+          </div>
         </div>
 
-        <Card raised={true} className={styles.TableOfContentsBox}>
+        <div className='m-cardGutter flex w-[90%] justify-center rounded-sm bg-white py-4 shadow-raised'>
```
benhammondmusic, line 59: same with `w-90p`

frontend/src/pages/ui/Sidebar.tsx
benhammondmusic: what are your thoughts on semantic html for sidebar navigations? I've seen some side (like the Astro docs) that use an `<aside>` for this type of "on this page" floating internal link box, but then others like Tailwind's docs just use a `<div>`. I'm tempted to go with `<aside>` some a little bit of virtue signaling that we push for accessibility when possible but i'll leave it up to you

frontend/src/pages/ui/TopicInfoModalButton.tsx
```diff
@@ -33,11 +32,12 @@ export default function TopicInfoModalButton() {
   if (!configArray) return <></>
 
   return (
+    // @BEN: keep this as fully MUI component?
```
benhammondmusic, line 35: fine for this PR, yep. just remove this comment and merge as is. If we find this exact type of component elsewhere and it makes sense we can abstract into a `<HetMoreInfoButton>` or something but i dont want to prematurely optimize

frontend/tailwind.config.ts
```diff
@@ -18,22 +18,22 @@ export default {
...
         'rgba(0, 0, 0, 0.2) 0px 5px 5px -3px, rgba(0, 0, 0, 0.14) 0px 8px 10px 1px, rgba(0, 0, 0, 0.12) 0px 3px 14px 2px',
       'raised-tighter':
         'rgba(0, 0, 0, 0.1) 0px 3px 3px -2px, rgba(0, 0, 0, 0.08) 0px 6px 7px 0px, rgba(0, 0, 0, 0.06) 0px 2px 9px 1px',
-
     },
     colors: {
       'alert-color': 'rgb(var(--alert-color) / <alpha-value>)',
-      'alt-black': 'rgb(var(--alt-black) / <alpha-value>)',
+      // 'alt-black': 'rgb(var(--alt-black) / <alpha-value>)',
+      'alt-black': '#383838', // @BEN: ex.: how about this pattern to use tailwind theme config instead?
```
benhammondmusic, line 36: i'd like to keep as is for now so that the design tokens (the actual color values) are defined outside of this config file. also, to allow tailwind opacity and other modifiers to work you have to use that strange rgb / alpha syntax

frontend/tailwind.config.ts
```diff
@@ -167,26 +168,29 @@ export default {
...
       padding: {
         '1p': '1%',
         '15p': '15%',
-
-      }
+      },
+      // for use w/spacing utilities: 'm', 'p', 'gap', etc.
+      spacing: {
+        cardGutter: '8px',
+      },
```
benhammondmusic, line 189: nice

benhammondmusic (changes requested):
fantastic! thank you so much @colespen , really nice work. A few small suggestions to fix up, and then the only thing I really don't want to change yet is the color definition in the tailwind config file... we need to figure something out there but i think it'll be a layer above and not changing the way the Tailwind config is set up. Once you revert those config changes and remove the comments tagging me feel free to "squash and merge" :)

frontend/src/pages/ui/Sidebar.tsx
colespen: > what are your thoughts on semantic html for sidebar navigations? I've seen some side (like the Astro docs) that use an `<aside>` for this type of "on this page" floating internal link box, but then others like Tailwind's docs just use a `<div>`. I'm tempted to go with `<aside>` some a little bit of virtue signaling that we push for accessibility when possible but i'll leave it up to you

If you'd prefer, absolutely. Since the `<TableOfContents/>` is linked to parts of the current page how about encapsulating it in `<nav>`? Then the entire Sidebar instead is contained within `<aside/>`.

frontend/src/pages/ui/Sidebar.tsx
colespen: ah! corrected

frontend/src/pages/ui/Sidebar.tsx
colespen: done

frontend/src/pages/ui/TopicInfoModalButton.tsx
colespen: ok, understood

frontend/tailwind.config.ts
colespen: ok, got it.

frontend/src/pages/ui/Sidebar.tsx
benhammondmusic: we can leave as a div for now and research if there is a more appropriate element for a filter box. i know if we do use `<nav>` we may need to use an aria label to distinguish when having multiple nav landmarks on a page (this has been flagged by the a11y checker before i believe)

colespen: ok sounds good - and aria label added.

benhammondmusic (approved):
🎉

PR #2715, opened 2024-01-05
Fixes previously missed test string description

JDemlow:
# Description and Motivation
Fixes previously missed test string description

## Types of changes

(leave all that apply)

- Refactor / chore


benhammondmusic (approved):
Nice; thanks man!

PR #2839, opened 2024-01-30
Refactors PDOHLink for New Methodology

### LINKED ISSUE #2827: Refactor New Methodogy Category  PdohLink.tsx
To reduce duplication of content, and to improve code maintainability and simplicity, we should rely on the existing MetricConfig to store user-facing labels, definitions, descriptions, etc for each DataSourceId. For the following, refer to PR #2818 that completed these steps for the Behavioral Health category

- [x] move the declaration of sources used for this category page from `{Category}Definitions.tsx` file to the top of `{Category}Link.tsx` file
- [x] copy this category's topics' clinical important text fields from `methodologyTableDefinitions` to this category's MetricConfig. 
- [x] in `{Category}Link.tsx` , import `{CATEGORY}_CATEGORY_DROPDOWNIDS` 
- [x] use that id array to generate correct configs [like this](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/frontend/src/pages/Methodology/methodologySections/BehavioralHealthLink.tsx#L25-L29)
- [x] switch from using `<KeyTerms>` to using `<KeyTermsAccordion>` (sending in the configs list generated above)
- [x] use id array to generate and export a category specific string containing comma-separated topic titles [like this](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/frontend/src/pages/Methodology/methodologySections/BehavioralHealthLink.tsx#L25-L29)
- [x] embed any sub Data Types within the string after their parent dropdown inside parentheses. So the resulting string should look like `Voter Participation, Incarceration (Prison, Jail), `
- [x] refactor manual list of category topic names to the topic titles string generated above [like this](https://github.com/SatcherInstitute/health-equity-tracker/blob/main/frontend/src/pages/Methodology/methodologySections/BehavioralHealthLink.tsx#L57)
- [x] update `TopicsLink.tsx` table to render the dynamic list of all topics string from above
- [x] delete the intermediary `{Category}HealthDefinitions.tsx` file
- [x] remove "Variables" column

kccrtv:
# Description and Motivation

closes #2827 

## Types of changes

(leave all that apply)

- Refactor / chore


frontend/src/pages/Methodology/methodologyComponents/GlossaryTerm.tsx
```diff
@@ -20,7 +20,7 @@ export const parseDescription = (description: string) => {
   const elements: any[] = []
   let remainingText = description
 
-  while (remainingText.length > 0) {
+  while (remainingText?.length > 0) {
```
benhammondmusic, line 23: i added this in; I think we'll eventually factor out this section but for now an empty `remainingText` was throwing an error because it wont have a length. adding the optional chaining `?.` allows that error to instead just be `undefined`

benhammondmusic (approved):
Looks great! Thanks @kccrtv

PR #2855, opened 2024-02-02
CDC WISQARS RF: Intent Handling and Test Files

eriwarr:
# Description and Motivation

1. Remove injury intents from non-fatal data.
2. Exclude unintentional, undetermined, and non-injury intents from fatal data.
3. Update associated test files to align with these changes.

In alignment with the discussions held during our meeting with AECF on 1/26, we are streamlining the data by removing injury intents for non-fatal data and certain injury intents for fatal data. This enhances the accuracy and relevance of the data for our analysis.

https://docs.google.com/document/d/1h_TKz_8VvMyD89fwsTgbqOoEiIS9HYMCI1wBChLeKbI/edit

## Has this been tested? How?
Tested locally and on GCP

## Types of changes
- Refactor/chore


python/datasources/cdc_wisqars.py
```diff
@@ -122,6 +115,9 @@ def write_to_bq(self, dataset, gcs_bucket, **attrs):
             table_name = f"{demographic}_{geo_level}_{table_type}"
 
             if table_type == CURRENT:
+                df_for_bq[std_col.TIME_PERIOD_COL] = (
+                    df_for_bq[std_col.TIME_PERIOD_COL] + '-01'
+                )
```
benhammondmusic, line 120: what is this doing?

benhammondmusic (approved):
awesome

python/datasources/cdc_wisqars.py
eriwarr: This adds a month to the code. It seems that the `preserve_only_current_time_period_rows` function requires a month to parse it correctly: `(constant) DT_FORMAT_YYYY_MM: Literal['%Y-%m']`

Issue #2654, opened 2023-12-15
Should update to node 20


benhammondmusic:
node 20 is stable and we should update everywhere including the docker file, the github actions, and the frontend config files

PR #2857, opened 2024-02-03
Upgrade to Node.js 20.11.0 LTS

Juneezee:
# Description and Motivation

Closes https://github.com/SatcherInstitute/health-equity-tracker/issues/2654.

## Has this been tested? How?

- `npm run build`
- `npm run test`

## Types of changes

- Refactor / chore


.github/workflows/frontend.yml
```diff
@@ -24,9 +24,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
-          node-version: 19
+          node-version-file: frontend/package.json
```
Juneezee, line 29: Automatically detect the `node` version that we set in our `frontend/package.json`, see https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

In the future, we only need to update the `node` version in `package.json`.

frontend/package.json
```diff
@@ -127,8 +127,8 @@
     "url": "npx playwright install chromium && npx playwright test --project=URL"
   },
   "engines": {
-    "node": ">=19",
-    "npm": ">=6.14.8"
+    "node": ">=20.11.0",
+    "npm": ">=10.2.4"
```
Juneezee, line 131: image

https://nodejs.org/en/about/previous-releases#looking-for-latest-release-of-a-version-branch

.github/workflows/frontend.yml
benhammondmusic: oh nice!

benhammondmusic (approved):
fantastic

PR #2908, opened 2024-02-14
BACKEND: Refactor `cdc_wisqars.py` to use utils from `cdc_wisqars_utils.py`

eriwarr:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore


python/datasources/cdc_wisqars.py
```diff
@@ -249,108 +225,77 @@ def load_wisqars_df_from_data_dir(breakdown: str, geo_level: str):
             dtype={"Year": str},
         )
 
+        df.columns = df.columns.str.lower()
```
benhammondmusic, line 228: I'm fine with this but why did you decide to lowercase the incoming columns and match against those?

python/datasources/cdc_wisqars.py
benhammondmusic, line 1: You can do it now or in a future PR, but since we are splitting the tables into `_current` and `_historical`, we should be dropping the unneeded columns in each as well. 

The normal pattern for which metrics go in which tables is

Current:
- 100k (or pct_rate)
- pct_share
- pop_share
- estimated_total(s)

Historical:
- 100k (or pct_rate)
- pct relative inequity
- pct_share (only for the unknown bubbles over time)

I've added a checkbox in #2361; we should really make a single util function that accepts the full df along with the lists of column names, and then outputs the _current and _historical (or _time_series we should align all these) with the needed columns and it could generate the BQ_TYPES as well

benhammondmusic (approved):

python/datasources/cdc_wisqars.py
eriwarr: It's more of a preference. The shared `combine_race_and_ethnicity` function looks for a `ethnicity` column so I just opted to use lowercase column names for consistency.

benhammondmusic: Ah gotcha. Seems like it might be worth changing the function API to require kwargs for the "race" and "ethnicity" column names? I feel like that's the normal
Pattern for our dataset util functions.

PR #2924, opened 2024-02-20
Splits ACS CONDITION into `_historical` and `_current` tables

### LINKED ISSUE #2361: Split sources with time data into `_historical` and `_current`

benhammondmusic:
# Description and Motivation

- completes split table setup of #2361 for acs condition
- N count numerators and denominators are added to `_current` tables for later use on the frontend

## Has this been tested? How?

- [x] tests updated and passing
- [x] runs on infra test

Screenshot 2024-02-20 at 7 46 50 AM
Screenshot 2024-02-20 at 7 48 15 AM

## Types of changes

(leave all that apply)

- Refactor / chore

### DIFF
--- a/python/datasources/acs_condition.py
+++ b/python/datasources/acs_condition.py
@@ -2,6 +2,7 @@ import pandas as pd
 from datasources.data_source import DataSource
 from ingestion import url_file_to_gcs, gcs_to_bq_util, census
 import ingestion.standardized_columns as std_col
+from ingestion.constants import HISTORICAL, CURRENT
 
 from ingestion.census import (
     parse_acs_metadata,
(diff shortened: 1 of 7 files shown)


eriwarr (approved):
👍🏿

PR #2931, opened 2024-02-21
BACKEND: Refactor WISQARS data sources to use `generate_time_df_with_cols_and_types`

eriwarr:
# Description and Motivation

This PR adds a util function, `generate_time_df_with_cols_and_types`,   that accepts a DataFrame and specified lists of column names and then filters and organizes the data according to its type (current or historical). It also automates the generation of BigQuery data types. 

This PR also updates cdc_wisqars.py to use the population_pct columns for the `generate_pct_rel_inequity_col`. 

This addresses the need outlined in issue #2361.

## Has this been tested? How?
Tested locally

Current DF
- Screenshot 2024-02-21 at 3 40 00 PM
- 
Historical DF
- Screenshot 2024-02-21 at 3 39 02 PM

## Types of changes
- New content or feature
- Refactor/chore


python/ingestion/dataset_utils.py
```diff
@@ -596,3 +553,44 @@ def combine_race_ethnicity(
...
+    Parameters:
+    - df: The source DataFrame to be processed.
+    - time_cols: A list of column names related to `table_type` that will be
+      included in the DataFrame and converted to floats.
+    - table_type: `current` or `historical`.
+    - dem_col: The name of the demographic column to be included in the DataFrame.
+
+    Returns:
+    - A tuple containing the processed DataFrame and a dict mapping column names
+      to their BQ data types ('STRING' or 'FLOAT').
```
benhammondmusic, line 577: `FLOAT64`

python/ingestion/dataset_utils.py
```diff
@@ -596,3 +553,44 @@ def combine_race_ethnicity(
...
+    table_type: Literal['current', 'historical'],
+    dem_col: Literal['age', 'race', 'race_and_ethnicity', 'sex'],
+):
+    """
+    Accepts a DataFrame along with list of column names for either current or
+    historical data and generates the appropiate BQ types for each column.
+
+    Parameters:
+    - df: The source DataFrame to be processed.
+    - time_cols: A list of column names related to `table_type` that will be
```
benhammondmusic, line 570: instead of `time_cols` can we use something more descriptive like `numerical_cols_to_keep`

python/ingestion/dataset_utils.py
```diff
@@ -596,3 +553,44 @@ def combine_race_ethnicity(
     df = df.drop(columns=[std_col.RACE_COL, std_col.ETH_COL])
 
     return df
+
+
+def generate_time_df_with_cols_and_types(
```
benhammondmusic, line 558: can you add some tests for this in `test_dataset_utils.py` ?

python/ingestion/dataset_utils.py
benhammondmusic: also somewhere, maybe this doc string, it'd be helpful to specify the types of columns that the user should send in for each time type, e.g. 100k or pct_rate are in both, pct_share are in current and only in historical if there are unknown demographic rows; pct_rel_inequity is only in the historical tables, raw / estimated total are only in the current. maybe mention which frontend tables require which metrics

benhammondmusic (approved):
awesome! lot's of little comments on how we can improve documentation for future users but nothing major, it looks really great. thanks!

Issue #2911, opened 2024-02-15
We should create a self-contained mirror site in case of GCP issues


benhammondmusic:
Right now we rely heavily on GCP to host the microservices including the production servers. However, given our low to medium traffic, I think we could honestly host the frontend entirely on Netlify, perhaps even the free tier. 

This means we wouldn't have access to the GCP Airflow / Cloud Run microservices to ingest and output the BQ tables, however, it would definitely be possible to create local flows that ran the data source .py files and ultimately generated the same .json files that are exported from BQ (and loaded into the frontend)

There is already support in place in the front end to load .json tables from `/public/tmp`; we could adjust this flow and create a new environment that specifically loaded all json files in the tmp folder. 

The new flow would be entirely local; simply run the .py files -> output .json files directly into /frontend/public/tmp and then this mirror environment wouldn't try to load the json from a server anywhere, just from the public folder

Reasons to consider this, and even switching to it entirely:
- far less complexity
- no need for GCP accounts per person (or any at all)
- no need to beg continuously from Google.org for free credits, and no longer at their whim for if/how much will be provided

benhammondmusic:
closing as it appears we have cloud spending WAY under budget, so we'd be fine going forward even if Google continues to reduce their in-kind support with google cloud credits. If they remove the credits entirely, then we can reconsider this migration to a free tier of a PAAS like Netlify, Render etc. There may be some with significant discounts for non-profit, open-source, diversity-focused projects

PR #2936, opened 2024-02-25
Local Pipeline: National COVID Vaccinations

benhammondmusic:
# Description and Motivation

Related to #2911 

By extending the existing `write_to_bq` method, we can keep our existing tests in place, and allow the BQ flow to continue uninterupted, but now optionally redirect the written table to local frontend instead of BQ. NOTE: Eventually we'll likely want to rename this method across the entire codebase including the `data_server` to something like `run_pipeline()`

## Pipeline

- removes deprecated table name methods from the parent DataSource class
- adds a `local_pipeline_utils.py` file to hold new utility functions that will help us run our pipelines locally, including reading real data directly from the `/data` dir, and writing json files to frontend/public/tmp folder (to allow local loading)
- adds a new python file that instantiates the class (and can easily extend to iterate over all data sources, and the calls the write to bq with the "write to local frontend" flag enabled. NOTE: this is still making real calls out to BQ for population data and FIPS data, etc. We should consider re-structuring this, so that here and also for our tests, instead of having to mock, our data sources could just read from the data directory like `data/acs_population/` or `data/state_ids` or similar

## `CDCVaccinationNational`

- adjusts `CDCVaccinationNational.write_to_bq()` to accept an optional boolean keyword arg
- adjusts `CDCVaccinationNational.write_to_bq()` so if that flag is passed, it uses the util from above to write to frontend instead of writing to bigquery

## Has this been tested? How?

- working locally, it writes out json files as expected to the `/frontend/public/tmp` folder
- by adding the file names into the .env file and restarting the localhost server it automatically serves those tables instead of reading from the backend

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
🤘🏿

PR #2940, opened 2024-02-27
Adds console warning if .env file hasn't set the apiBaseUrl

benhammondmusic:
# Description and Motivation

if a user sets up a fresh repo locally, they must copy the .env.example over into an .env.development so that the app can read in the api base url. at some point we may need to have users keeping sensitive info in the local env that we don't want to check in to GitHub publically, so that's why this set up is in place. 

This PR
- removes some deprecated logic from when Storybook was installed
- adds a console warning when the app detects that the env was not retreived and the user is running in dev mode

## Has this been tested? How?

intentionally broke the .env file and observed the warning in the console

Screenshot 2024-02-27 at 9 05 09 AM

## Types of changes

(leave all that apply)

- Refactor / chore

### DIFF
--- a/frontend/src/utils/Environment.ts
+++ b/frontend/src/utils/Environment.ts
@@ -62,7 +62,7 @@ export class HetEnvironment implements Environment {
   }
 
   private getEnvVariable(nonPrefixedName: string): string | undefined {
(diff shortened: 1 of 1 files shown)


alinix1 (approved):
Awesome! I'll look at the README from now on too. Will make it a habit 😊

PR #2944, opened 2024-02-28
Makes all FIPS merges local

benhammondmusic:
# Description and Motivation

In service of #2911, this PR switches all merging of FIPS info to happen locally instead of needing to reach out to public BQ tables

- adds csvs for state and county fips codes in a new folder within ingestion
- updates `ingestion/setup.py` to have those csv files install as `package-data`
- updates merge_utils methods for merging state and county FIPS/names to read from those local package data csvs instead of public bq
- updates all test files by removing the previously required mocks for reads of the public bq, since it's reading locally now
- removes the shared util mock function for fips codes
- removes the files used by the now deleted mock functions
- updates GitHub action for pytest to execute the test calls from the root, rather than from within each of the services `python/` `data_server/`, and `exporter/` (this mimics how we normally run the tests locally, and alleviates some of the file directory issues we encountered before)

## Has this been tested? How?

- all tests passing
- made sure some DAGs ran on infra-test

> this phrma dag was run after the recent updates to use local FIPS data from .csv files not python files

Screenshot 2024-02-28 at 2 02 32 PM

Screenshot 2024-02-27 at 9 52 50 PM

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
Awesome!

eriwarr (approved):
🤘🏿

PR #2952, opened 2024-03-02
Make ACS population merges local

benhammondmusic:
# Description and Motivation

Makes all ACS (and DECIA) population merges happen locally rather than reaching out to the HET BQ tables

- checks in ACS/DECIA population data from HET BQ (current and time_series)
- puts FIPS codes from #2944 into a sub-dir
- installs all FIPS and population csv files in sub-dirs of `ingestions/` as package-data in the setup.py
- updates merge_utils to find those csv based on the calling script (this allows it to run locally from the `/python/ingestion` folder and from the GCP container as `/app/python/ingestion`
- updates merge_utils to source ACS/DECIA population from local tables instead of fetching from HET BQ
- updates golden data for some sources when test results / rates have changed due to now using real 2022 population instead of fake data
- removes the `het_bq_tables_for_mocks` that held mock acs and mock FIPS data previously
- updates test files to no longer mock out the calls to retrieve HET BQ tables
- removes assertions based on testing those population mocks
- removes shared util for mocking HET BQ fetches
- lots of `black` formatting

NOTE: there are still some cases where datasources are fetching from GCP, including the ACS caching step, and also the COVID age-adjustment step.

## Has this been tested? How?

- [x] tests updated and passing
- [x] some sources run on infra-test

Screenshot 2024-03-04 at 12 18 02 PM
Screenshot 2024-03-04 at 12 27 28 PM

## Types of changes

(leave all that apply)

- Refactor / chore


python/ingestion/setup.py
```diff
@@ -5,5 +5,5 @@
     package_dir={'ingestion': ''},
     packages=['ingestion'],
     include_package_data=True,
-    package_data={'ingestion': ['merge_data/state_level_fips.csv', 'merge_data/county_level_fips.csv']},
+    package_data={'ingestion': ['merge_data/**/*.csv']},
 )
```
eriwarr, line 9: Just curious. What does this do?

eriwarr (approved):
🤘🏿

python/ingestion/setup.py
benhammondmusic: the `package_data` declaration is to let python known which non-python files need to be installed as part of the program, so like in this case the csv files (otherwise the .venv locally won't have access to those files)

but it's not working yet, I tried to use the **/* pattern to copy ALL csvs regardless of subdirectory but it's erroring so im trying some other ways

PR #2982, opened 2024-03-08
Merges new non-adult ACS pop. on to AHR

### LINKED ISSUE #2973: Need to merge children pop_pct onto AHR
In order to address #2965 we need to:

- [ ] adjust the AHR python datasource to merge rows for `0-14` and `0-17` onto the age tables, even though there will not be any metric data for those (since AHR measures only adults). 
- [ ] rerun pipeline
- [ ] adjust frontend if needed to ensure that rows for children are shown on the data table (since we have the pop pct share figures) and that the rate and pct_share cells appear as null triangles

benhammondmusic:
# Description and Motivation

closes #2973

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore


eriwarr (approved):
🤘🏿

Issue #2932, opened 2024-02-22
QA for Gun Violence Frontend


benhammondmusic:
Follow up issues from #2915

![image](https://github.com/SatcherInstitute/health-equity-tracker/assets/41567007/00b3679b-8c83-4d70-acfb-da9d2c8a35bf)

- [x] what is "m"
- [x] are we sure that none of these gray states are reporting 0 per 100k?
- [x] the estimated total population in the breadcrumbs after the general population can't be right, and if we are showing the gun violence denominator pop. there which would be great; we need a clear label

/exploredata?mls=1.gun_violence_non_fatal_population-3.00&group1=All

- [x] right now the only thing showing on this by race report is national, by ALL, over time. We should disable the race view and default to a different breakdown demo 
- [x] ![image](https://github.com/SatcherInstitute/health-equity-tracker/assets/41567007/dd7dacbd-21e2-429b-9aa3-a7e5a38a0448) "Females" isn't picking up one of the correct colors. These need to be "Female" and "Male" on the backend 
- [x] ![image](https://github.com/SatcherInstitute/health-equity-tracker/assets/41567007/36a974de-9a90-40f2-aea5-89aee92b46a8) the rates over time chart is missing a title in the metric config so is falling back to "Data" 
- [x] is the data supposed to stop in 2015

PR #3045, opened 2024-03-22
Fixes incorrect "Males" "Females" plural sex groups

benhammondmusic:
# Description and Motivation

- part of #2932 
- the resulting groups for sex need to be `Male` `Female` `Other` and `All`. The source data from WISQARS are `Females` and `Males` which causes the frontend to not properly color the lines and potentially other presentation issues

## Has this been tested? How?

golden data updated, tests passing

## Types of changes

(leave all that apply)

- Bug fix


eriwarr (approved):
Nice catch!

PR #3052, opened 2024-03-26
Fixes labels and spellcheck

benhammondmusic:
# Description and Motivation

- fixes/improvements to cspell config (in trying to ignore the root level `config/` folder we were actually ignoring the MetricConfig files that were inside `frontend/src/config` and so were also ignored
- fixes newly discovered spelling errors
- changes "amongst" to "among" where needed
- removes "among adults" on gun violence topics since it's actually all ages
- fixes incorrect label for HIV linkage map population sub-group

## Has this been tested? How?

tests passing

## Types of changes

(leave all that apply)

- Bug fix


eriwarr (approved):
🤘🏿

PR #3161, opened 2024-04-18
Adds national data handling to maternal mortality

benhammondmusic:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore


alinix1 (commented):
Looks like we will be using both national-level and state-level data then? For the time period of 1999 - 2019. National-level data is broken down into regions (Midwest, Northeast, South, West) as well as overall US, by racial group. 

Before we only had the state-level table included?

benhammondmusic:
Yes, and actually I feel kind of dumb. Just realized that "National" is included in our state level source data. So we don't have to do the whole deal of using that alternate table. I'll submit some new code splitting them out. But this is great because out state and national will have the same fidelity. We need national to fit the HET display patterns as they currently exist. We can't handle regional data really yet, but have considered including support for that

PR #3165, opened 2024-04-19
Fixes national data loading for MMR

benhammondmusic:
# Description and Motivation

- since the national aggregations are already part of the "state level" data table, we can simply load the table in once, and then split it into the needed "NATIONAL" and "STATE" rows
- we then still iterate over the two geo levels, doing the remaining level specific processing and then writing to BQ
- prefers using standardized column strings where possible to avoid potential typos / mismatched labels

NOTE: I did not delete the manual national file i made, as we can still use the live birth and maternal death `estimate_total` values for our pct_share and similar calculations later on

## Has this been tested? How?

prints as expected

## Types of changes

(leave all that apply)

- Refactor / chore


alinix1 (approved):
From my understanding, we call the preprocess_source_data() function (responsible for loading in CSV file and preprocessing the data, specifies columns to use and renames certain columns) within the write_to_bq() method. This method is responsible for writing the preprocessed data to the BigQuery table. Further, it iterates over the geo_levels: STATE_LEVEL and NATIONAL_LEVEL using a `for in` loop and processes the data separately for each level of geography. Specifies which columns to keep, for string and numbers. Other duties include merging population numbers with merge_pop_numbers() and rounds to whole numbers for MM_PER_100K. All of this ensures the data is in the correct format before shipping off to the BigQuery table.

PR #3186, opened 2024-04-23
Fix: Use Population Percentage Instead of Raw Population Counts

### LINKED ISSUE #3185: Bug: should be `pop_pct` not raw population counts?
https://github.com/SatcherInstitute/health-equity-tracker/blob/ddf9acd3e2e4d89366bafb4f5cbd12e84c6c8e44/python/datasources/cdc_wisqars_youth.py#L134-L136

Looking at the util fn signature for calculating pct_rel_inequity and the way it's used elsewhere, I believe the function needs the name of the population percent share column, not the raw population counts columns

eriwarr:
# Description and Motivation

- [ ] Fixes #3185 (uses population percent to calculate the percent relative inequity instead of the raw population count.
- [ ] Updates and moves docstring to the top of the file. 

## Has this been tested? How?
Manually tested.  

## Types of changes
- Bug fix

### DIFF
--- a/python/datasources/cdc_wisqars_youth.py
+++ b/python/datasources/cdc_wisqars_youth.py
@@ -1,3 +1,30 @@
+"""
+This documentation outlines the procedure for acquiring gun violence data for the general
+population from the CDC WISQARS database. The data, once downloaded, is stored locally in
+the `data/cdc_wisqars` directory for further processing and analysis.
+
+Instructions for Downloading Data:
+1. Access the WISQARS website at https://wisqars.cdc.gov/reports/.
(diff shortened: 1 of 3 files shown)


python/datasources/cdc_wisqars_youth.py
```diff
@@ -1,3 +1,30 @@
+"""
+This documentation outlines the procedure for acquiring gun violence data for the general
```
benhammondmusic, line 2: this is for youth / young adult not general pop.

benhammondmusic (approved):

PR #3200, opened 2024-04-25
initial commit

eriwarr:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore


benhammondmusic (commented):
Screenshot 2024-04-25 at 5 25 49 PM

- missing the subtitle on the map card (should say like "Ages 18-25" etc)
- can we check the sub-population metric? it doesn't make sense to me that the population of people ages 18-25 in the entire nation is only 9000 ish

benhammondmusic (commented):
Screenshot 2024-04-25 at 5 30 18 PM

- need to enforce that this report is only available as a "race" breakdown, and not as "sex" or "age"

benhammondmusic (approved):
awesome

PR #3276, opened 2024-05-09
adds test coverage for wisqars

### LINKED ISSUE #3195: Need to add tests for wisqars_util functions
There are some utility functions defined and shared between all the gun violence / wisqars data sources. we should ensure some basic test coverage of these functions, particularly as they are used across multiple files

https://github.com/SatcherInstitute/health-equity-tracker/blob/ed6d9a0f21fb9becdf9da0d85fdc8dca5ede321e/python/ingestion/cdc_wisqars_utils.py#L61-L120

alinix1:
# Description and Motivation

Updated doc strings in wisqars_utils for generate_cols_map function. 
Wrote test coverage for wisqars_utils functions including sad paths.

closes #3195

## Has this been tested? How?

Tests pass locally

## Types of changes

(leave all that apply)

- Refactor / chore


benhammondmusic:
the PR is failing because yet again we misunderstood the function.
it actually accepts a list of prefixs, and in the case those prefixs end with _estimated_total, it chops that off. then it outputs a map of the incoming list of prefixs to their new column name combining the prefix and the suffix.
so you can remove the assertion in the util function, and remove the test for the assertion error. you can also have the sample input data be a list that includes estimate total like ["cat", "dog", "some_condition_estimated_total"]  and then the output if the suffix is "per_100k" should be
{
"cat": "cat_per_100k",
"dog": "dog_per_100k",
"some_condition": "some_condition_per_100k"
}
make sure you update the doc string to indicate that the incoming list if prefixes like Eric initially had it, but indicate that those prefixes can optionally end in _estimated_total and that if they do the real prefix will be extracted

python/ingestion/cdc_wisqars_utils.py
```diff
@@ -104,17 +104,24 @@ def convert_columns_to_numeric(df: pd.DataFrame, columns_to_convert: List[str]):
         df[column] = pd.to_numeric(df[column], errors='coerce')
 
 
-def generate_cols_map(prefixes, suffix):
+def generate_cols_map(estimated_total_cols: List[str], suffix: str):
```
benhammondmusic, line 107: let's keep as `prefixes`

python/ingestion/cdc_wisqars_utils.py
```diff
@@ -104,17 +104,24 @@ def convert_columns_to_numeric(df: pd.DataFrame, columns_to_convert: List[str]):
...
     """
-    Generates a mapping of columns with given prefixes to their corresponding columns
-    with specified suffix.
+    Generates a mapping of column names from a list of estimated total column names to a column name
+    with a specified suffix.
```
benhammondmusic, line 110: Generates a mapping of column prefixes to new column names, based on incoming list of prefixes and incoming suffix. Note: if any of the incoming prefixes are in the count column format (ending in `_estimated_total`), the resulting dict will use the original count col as the key, but will swap `_estimated_total` for the new suffix in the column name value.

python/ingestion/cdc_wisqars_utils.py
```diff
@@ -104,17 +104,24 @@ def convert_columns_to_numeric(df: pd.DataFrame, columns_to_convert: List[str]):
...
 
     Parameters:
-    prefixes (list): A list of prefixes to generate the mapping for.
-    suffix (str): The suffix to add to the columns.
+        estimated_total_cols (List[str]): A list of column names, each ending with a standard raw suffix.
```
benhammondmusic, line 113: estimated_total_cols (List[str]): A list of column prefix strings

python/ingestion/cdc_wisqars_utils.py
```diff
@@ -104,17 +104,24 @@ def convert_columns_to_numeric(df: pd.DataFrame, columns_to_convert: List[str]):
...
+        suffix (str): The suffix that will replace each RAW_SUFFIX in the new column names.
```
benhammondmusic, line 114: suffix (str): The new suffix that will applied to the incoming list of prefixes for the generated column names

python/ingestion/cdc_wisqars_utils.py
```diff
@@ -104,17 +104,24 @@ def convert_columns_to_numeric(df: pd.DataFrame, columns_to_convert: List[str]):
...
 
     Returns:
-    dict: A dictionary mapping the original prefixes to the modified prefixes with the
-    specified suffix.
+        dict: A dictionary mapping original estimated total column names to new column names with the specified suffix.
```
benhammondmusic, line 117: dict: A dictionary mapping original prefixes to new column names with the specified suffix.

python/ingestion/cdc_wisqars_utils.py
```diff
@@ -104,17 +104,24 @@ def convert_columns_to_numeric(df: pd.DataFrame, columns_to_convert: List[str]):
...
+
+    Raises:
+        AssertionError: If any column name in `estimated_total_cols` does not end with the standard raw suffix.
```
benhammondmusic, line 120: remove this

python/ingestion/cdc_wisqars_utils.py
```diff
@@ -104,17 +104,24 @@ def convert_columns_to_numeric(df: pd.DataFrame, columns_to_convert: List[str]):
...
     """
-    return {prefix: prefix.replace(f"_{std_col.RAW_SUFFIX}", "") + f"_{suffix}" for prefix in prefixes}
+    for col in estimated_total_cols:
+        assert col.endswith(std_col.RAW_SUFFIX)
```
benhammondmusic, line 123: remove this

python/tests/ingestion/test_cdc_wisqars_utils.py
```diff
@@ -1,20 +1,97 @@
...
+    expected_map = {}
+    assert generate_cols_map(count_cols, suffix) == expected_map
+
+
+def test_generate_cols_map_bad_count_cols():
+
+    count_cols = ['cat_estimated_total', 'dog_estimated_total', 'bird']
+    suffix = 'per_100k'
+    with pytest.raises(AssertionError):
+        generate_cols_map(count_cols, suffix)
```
benhammondmusic, line 97: instead of expected an assertion here, just assert that it'll equal

```python

{
'cat_estimated_total': 'cat_per_100k',
'dog_estimated_total': 'dog_per_100k',
'bird': 'bird_per_100k'
}

```

benhammondmusic (changes requested):

python/ingestion/cdc_wisqars_utils.py
alinix1: does this format look correct? 

def generate_cols_map(prefixes: List[str], suffix: str):
    """
    Generates a mapping of column prefixes to new column names, based on incoming list of prefixes and incoming 
    suffix. Note: if any of the incoming prefixes are in the count column format (ending in `_estimated_total`), the 
    resulting dict will use the original count col as the key, but will swap `_estimated_total` for the new suffix in the 
    column name value. 

    Parameters:
        estimated_total_cols (List[str]): A list of column prefix strings.
        suffix (str): The new suffix that will be applied to the incoming list of prefixes for the generated column names.

    Returns:
        dict: A dictionary mapping original prefixes to new column names with the specified suffix. 
    """
   
    return {
        estimated_total_col: estimated_total_col.replace(f"_{std_col.RAW_SUFFIX}", "") + f"_{suffix}"
        for estimated_total_col in prefixes
    }

benhammondmusic: yep!

benhammondmusic (approved):
awesome! thanks so much Ali!

PR #3348, opened 2024-05-30
Fixes missing multimap bug

### LINKED ISSUE #3071: Some group maps are invisible on multimap view
**Describe the bug**
Some race maps don't appear, but can be hovered and show a value

**To Reproduce**
1. http://localhost:3000/exploredata?mls=1.hiv-3.24510&mlp=disparity&dt1=hiv_prevalence&group1=Hisp%7ELat&demo=race_and_ethnicity&multiple-maps=true
2. observe "Asian" isn't visible, but can be hovered and shows a value

Related to #2711 possibly

benhammondmusic:
# Description and Motivation

- Fixes #3071 
- the `fieldRange` prop was getting passed through, and that function was only filtering out `NaN` values but not `null` values. This was causing the weirdness (i think)
- adds unit test coverage for this edge case
- memoizes the fieldRange call rather than invoking it on every render

## Has this been tested? How?

- adds unit test
- tests pass
- manually confirmed on various reports, including one that includes a real zero value which gets properly colored

Screenshot 2024-05-30 at 11 39 02 AM
Screenshot 2024-05-30 at 11 39 15 AM

## Types of changes

(leave all that apply)

- Bug fix
- Refactor / chore


eriwarr (approved):
Awesome! Thanks Ben!

PR #3357, opened 2024-06-05
Adds new utils: estimating counts from rates, pct shares of summed group counts

### LINKED ISSUE #2994: Make a better `generate_estimated_total_col()` function
We deleted `estimate_total()` in #2995 as it is prone to misuse, inferring estimated totals from the "ALL" populations despite the topic potentially being a sub group like "adults" etc. 

We should remake this function to allow us to get pct_share, pct_relative_inequity, and other unique HET outcomes, while ensuring we avoid future data errors. 

- [ ] should make this a df based function, rather than a row-based that needs `apply()` as most numerical operations are MUCH faster using python's built in vectorized methods. 
- [ ] should add a warning that ensures the user is providing an accurate standard population on which the calculation can be made. 
- [ ] should allow the correct estimation, by providing the specific population to use (e.g. age=18+, sex=all, race=each race group). this might involve requiring the input df to have all three demographic columns and then the merge can be done properly against the population table with all 3? or possibly specifying the groups for each breakdown as args and then adjusting the df inside the function to allow the correct merge
- [ ] we will need to extend the age buckets in our acs by race by age by sex table to include 18+ and 15+ (for suicides) if we want to pursue this option.

benhammondmusic:
# Description and Motivation

- needed for AHR, CHR, and other sources where they only provide the rates but we would like to present our other metrics like N-counts, pct_share, and potentially pct_relative_inequity over time 
- closes #2994 
- adds `generate_estimated_total_col()` that back calculates raw counts from given rates against a standardized reference population. allows for specific sub-pops like "Black Men by age", etc. NOTE: due to ACS limitations, when using the sex/age/race table, every race except WHITE is potentially inclusive of Hispanic; white is specifically NH. We need to note this in the updated methodology explaining this back calculation
- adds `sum_age_groups()` to calculate population counts for age groups not provided in out intersectional ACS table. Currently only does `18+` but should be easily extensible to any combo group
- adds `generate_pct_share_col_of_summed_alls()` to get pct_share column but calculated as each group's pct_share of the sum of all groups (rather than the share of the back-calculated "All" that might result in the shares not totalling 100%)

## Has this been tested? How?

- unit tests added by sex covering all three new utils

## Types of changes

(leave all that apply)

- New content or feature


eriwarr (approved):
This is great! Thanks, Ben!

PR #3359, opened 2024-06-07
Split `merge_intersectional_population()` from `generate_estimated_total_col()`

benhammondmusic:
# Description and Motivation

In starting the task of adding estimated totals, pct_shares, etc to graphql_ahr, i confused myself with the number of things this original fn was doing and found myself trying to pass in too many flags / options to get the fn to work as needed. Better instead to breakthe large function apart so each just does "one thing", and then add other conditional logic OUTSIDE the util functions

- `generate_estimated_total_col()` was too large and did too many things; to improve the API i have extracted the population merge aspect of the function into a new function within merge_utils; the remaining functionality that does the back calculation of the estimated total remains in dataset_utils
- updates tests

## Has this been tested? How?

tests updated and passing

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
Nice!

PR #3366, opened 2024-06-10
Merges pop/pop18+ columns onto graphql_ahr

benhammondmusic:
# Description and Motivation

- ships the regular general population columns to BQ
- updates `merge_intersectional_pop()` to handle national level (by just summing the state-level table)
- uses that fn to generate the 18+ population totals. NOTE: we should decide if, when using a subset population like this, we should prefer using pct_share of subset population as our comparison population shares
- updates graphql_ahr golden data to expect these various pop cols

## Has this been tested? How?

- tests updated and passing

## Types of changes

(leave all that apply)

- New content or feature


python/ingestion/merge_utils.py
```diff
@@ -352,18 +352,28 @@ def merge_intersectional_pop(
...
     pop_dtype: dict[str, Union[Type[float], Type[str]]] = {
         std_col.POPULATION_COL: float,
     }
 
+    geo_file = ''
+
     if geo_level == COUNTY_LEVEL:
         pop_dtype[std_col.COUNTY_FIPS_COL] = str
-    if geo_level == STATE_LEVEL:
+        geo_file = 'county'
```
eriwarr, line 366: Could we use the predefined constants for this line?

`geo_file = COUNTY_LEVEL`

python/ingestion/merge_utils.py
```diff
@@ -352,18 +352,28 @@ def merge_intersectional_pop(
...
+    if geo_level in [STATE_LEVEL, NATIONAL_LEVEL]:
         pop_dtype[std_col.STATE_FIPS_COL] = str
+        geo_file = 'state'
```
eriwarr, line 369: Could we use the predefined constants for this line?

`geo_file = STATE_LEVEL`

eriwarr (approved):
Awesome!

PR #3379, opened 2024-06-17
Adds methodology for gun deaths measures

eriwarr:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore

### DIFF
--- a/frontend/src/data/config/DatasetMetadata.ts
+++ b/frontend/src/data/config/DatasetMetadata.ts
@@ -869,6 +869,7 @@ export const DatasetMetadataMap: Record<DatasetId, DatasetMetadata> = {
   'cdc_wisqars_data-race_and_ethnicity_state_historical': {
     name: 'Gun-related Deaths, by race/ethnicity and state',
     original_data_sourced: '2018-2021',
+    contains_nh: true,
     source_id: 'cdc_wisqars_data',
   },
   'cdc_wisqars_data-sex_state_historical': {
@@ -885,6 +886,7 @@ export const DatasetMetadataMap: Record<DatasetId, DatasetMetadata> = {
   'cdc_wisqars_data-race_and_ethnicity_state_current': {
     name: 'Gun-related Deaths, by race/ethnicity and state',
     original_data_sourced: '2021',
+    contains_nh: true,
     source_id: 'cdc_wisqars_data',
   },
   'cdc_wisqars_data-sex_state_current': {
@@ -901,6 +903,7 @@ export const DatasetMetadataMap: Record<DatasetId, DatasetMetadata> = {
   'cdc_wisqars_data-race_and_ethnicity_national_historical': {
     name: 'Gun-related Deaths, by race/ethnicity, nationally',
     original_data_sourced: '2018-2021',
+    contains_nh: true,
     source_id: 'cdc_wisqars_data',
   },
   'cdc_wisqars_data-sex_national_historical': {
@@ -917,6 +920,7 @@ export const DatasetMetadataMap: Record<DatasetId, DatasetMetadata> = {
   'cdc_wisqars_data-race_and_ethnicity_national_current': {
     name: 'Gun-related Deaths, by race/ethnicity, nationally',
     original_data_sourced: '2021',
+    contains_nh: true,
     source_id: 'cdc_wisqars_data',
   },
   'cdc_wisqars_data-sex_national_current': {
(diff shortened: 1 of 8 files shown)


frontend/src/pages/Methodology/methodologyContent/missingDataBlurbs.tsx
```diff
@@ -244,3 +244,27 @@ export function MissingAHRData() {
...
+      <h4 className='text-text font-light'>
+        Missing WISQARS Data
+      </h4>
+      <ul className='m-0 ml-1 self-start text-small text-altBlack'>
+        <li>
+          <b>Legal intervention data:</b> Data on deaths caused by legal intervention
+          is limited. Therefore, we choose to show raw counts of deaths caused by legal
+          intervention in states where the data is available. This approach ensures
+          that the information presented is accurate and reflective of the available data
+          without making potentially misleading calculations based on incomplete information.
```
benhammondmusic, line 260: We should note here that the number of deaths from legal intervention available from the CDC is widely considered to be underreported

frontend/src/pages/Methodology/methodologySections/CommunitySafetyLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+                <p>
+                    The CDC’s Web-based Injury Statistics Query and Reporting System (WISQARS)
+                    collects and provides data on gun deaths in the United States. This information
+                    is gathered from state and local injury surveillance programs and is used to
+                    understand the impact of firearm-related injuries across the country. WISQARS™
+                    allows users to explore data by intent of injury, mechanism, geographic location,
+                    and demographics, supporting public health and policy decisions.
+                </p>
+                <HetNotice className='my-12' title='A note about CDC WISQARS'>
+                    The CDC’s WISQARS limit data released at state and county levels to protect
```
benhammondmusic, line 60: should be `limits` since WISQARS is singular?

frontend/src/pages/Methodology/methodologySections/CommunitySafetyLink.tsx
```diff
@@ -0,0 +1,174 @@
...
+                                <HetTerm>Gun-Related deaths by Suicides</HetTerm>
+                            </li>
+                            <li>
+                                <HetTerm>Gun-Related deaths for Children</HetTerm>
+                            </li>
+                            <li>
+                                <HetTerm>Gun-Related deaths for Young Adults</HetTerm>
+                            </li>
+                            <li>
+                                <HetTerm>Gun-Related deaths for Black Men</HetTerm>
```
benhammondmusic, line 88: usually we like to use sentence case, but i think title case is ok here. if that's what we are using we should capitalize `Deaths` as well

frontend/src/reports/WhatDataAreMissing.tsx
```diff
@@ -43,6 +44,7 @@ export default function WhatDataAreMissing(props: WhatDataAreMissingProps) {
   const isPhrma = currentDropDownIds.some((condition) =>
     PHRMA_CONDITIONS.includes(condition)
   )
+  const isWisqars = currentDropDownIds.includes('gun_violence')
```
benhammondmusic, line 47: this will only register the specific dropdown `gun_violence`, but we want it to register if youth or black men are chosen as well right? probably need something like the PHRMA check above where we check if `COMMUNITY_SAFETY_DROPDOWNIDS` includes `some()` of the currentDropDownIds. In general since we are using typescript, we should always be checking for specific entire terms, rather than using string manipulation to check for part of a string matching.

frontend/src/data/config/DatasetMetadata.ts
```diff
@@ -937,52 +941,62 @@ export const DatasetMetadataMap: Record<DatasetId, DatasetMetadata> = {
...
+    contains_nh: true,
     source_id: 'cdc_wisqars_data'
   },
   'cdc_wisqars_black_men_data-black_men_by_age_state_current': {
     name: 'Gun homicides for Black men, by age and state',
     original_data_sourced: '2021',
+    contains_nh: true,
     source_id: 'cdc_wisqars_data'
   },
   'cdc_wisqars_black_men_data-black_men_by_age_state_historical': {
```
benhammondmusic, line 996: should we also update the definition in the metric config to reflect that the "black men" topics are for Black NH?

benhammondmusic (approved):
great!

PR #3397, opened 2024-06-20
Optimize cdc restricted

benhammondmusic:
# Description and Motivation

Some of the DAG steps are timing out for COVID again. This PR makes some speed improvements to the code, and cleans up some typing issues.

- only load the pre-processed df once per demo/geo breakdown (then use for each time table type). Previously it was loading the table exactly the same way 2x
- use vectorized methods to check that the FIPS codes are valid
- uses unsigned 32 bit instead of signed 64 bit to hold case counts. This gives some performance bump, and we know that case counts wont be negative, and won't total more than half a trillion for MANY decades if ever

## Has this been tested? How?

- tests updated and passing
- [ ] doesn't time out when run via `infra-test`

## Types of changes

(leave all that apply)

- Bug fix


python/datasources/cdc_restricted.py
```diff
@@ -271,12 +278,9 @@ def remove_bad_fips_cols(df):
...
 
     df: The DataFrame to toss rows out of."""
-
-    def fips_code_is_good(row):
-        return row[std_col.COUNTY_FIPS_COL][0:2] == row[std_col.STATE_FIPS_COL]
-
-    df = df[df.apply(fips_code_is_good, axis=1)]
-    return df.reset_index(drop=True)
+    # Use vectorized string operations to compare the first two digits of COUNTY_FIPS_COL with STATE_FIPS_COL
+    mask = df[std_col.COUNTY_FIPS_COL].str[:2] == df[std_col.STATE_FIPS_COL]
```
eriwarr, line 282: Nice!

eriwarr (approved):
Awesome work!

PR #3400, opened 2024-06-21
Updates MadLib with new Category Maternal Health

### LINKED ISSUE #3307: Update MadLib with new category "Maternal Health" and dropdown "Maternal Mortality"
## Updating the MadLibs Configuration for New Dataset Integration

Determine if your new dataset fits within an existing category in the CategoryMap or if it represents a new thematic area, such as "Black Women's Health," that requires adding a new category. Update this variable if applicable.

- [x]  Update `DROPDOWN_TOPIC_MAP` to add an entry for your dataset, using the metric's name as the key and a descriptive label as the value.
~Update `SELECTED_DROPDOWN_OVERRIDES` (if necessary): If your dataset requires overriding the default selections in specific contexts (e.g., when it's chosen, another dropdown should default to a specific selection), update or add to this configuration.~
- [x]  Update `CATEGORIES_LIST` (if necessary): Similar to step 1, only update if you've introduced a new category.

JDemlow:
# Description and Motivation
Closes #3307 

## Has this been tested? How?

Yes, using VSCode terminal. Passes all tests

Screenshot (80)
Screenshot (79)

## Types of changes

(leave all that apply)

- New content or feature


frontend/src/utils/MadLibs.ts
```diff
@@ -19,6 +19,8 @@ import { SHOW_PHRMA_MENTAL_HEALTH } from '../data/providers/PhrmaProvider'
 import { SHOW_GUN_VIOLENCE } from '../data/providers/GunViolenceProvider'
 import { GEORGIA_FIPS, USA_FIPS } from '../data/utils/ConstantsGeography'
 import { FIPS_MAP } from '../data/utils/FipsData'
+import { title } from 'vega-lite/build/src/channeldef'
```
benhammondmusic, line 22: remove this?

frontend/src/utils/MadLibs.ts
```diff
@@ -229,7 +231,12 @@ const CATEGORIES_LIST: Category[] = [
     title: 'Community Safety',
     definition: '',
     options: COMMUNITY_SAFETY_DROPDOWNIDS as unknown as DropdownVarId[],
-  }
+  },
+  {
+    title: 'Maternal Health',
+    definition: '',
+    options: MATERNAL_HEALTH_CATEGORY_DROPDOWNIDS as unknown as DropdownVarId[],
+  },
```
benhammondmusic, line 239: sorry it isn't scoped out, but can we put this behind an env flag? you can follow the same pattern as `VITE_SHOW_GUN_VIOLENCE`

benhammondmusic (changes requested):
awesome! i should have scoped it out, but we have sstarted putting new content like this behind a feature flag, which allows us to show it in preview and dev but still keep it non-public on prod. Please update following the same pattern as gun violence ; let me know if it's unclear.

frontend/src/utils/MadLibs.ts
benhammondmusic: https://healthequitytracker.notion.site/Merging-behind-a-feature-flag-4e12dc6aac7e496b870cd7d895df7ea1?pvs=4

JDemlow:
Feature flag should be in place now

frontend/src/utils/MadLibs.ts
benhammondmusic: @JDemlow I think this import is a mistake right?

frontend/src/utils/MadLibs.ts
benhammondmusic: @JDemlow great, a few more steps needed for the feature flag:

- [ ] set that variable equal to true in the `.env.deploy_preview` and `.env.staging` files (and your local `.env.development if you want to see it locally; that env isn't checked in to git)
- [ ] now, the line of code you wrote will import the true/false state of the flag based on the environment it's in (netlify deploy, staging = dev site, or prod where it should still be false)
- [ ] we also want to conditionally show/hide that new dropdown MadLib option based on the flag. So for the code above, instead of directly declaring the maternal health object in the array, you'd want to conditionally push it onto the array. Something like: 

```
if (SHOW_NEW_MATERNAL_MORTALITY) CATEGORIES_LIST.append(
  {
    title: 'Maternal Health',
    definition: '',
    options: MATERNAL_HEALTH_CATEGORY_DROPDOWNIDS as unknown as DropdownVarId[],
   }
)
```

benhammondmusic (changes requested):

frontend/src/utils/MadLibs.ts
JDemlow: I don't recognize it. Maybe VSCode tried an auto import? I don't remember explicitly adding that. I'll check it when I get home

JDemlow:
Screenshot (110)
Screenshot (111)

benhammondmusic (approved):
Awesome!!!

PR #3419, opened 2024-06-26
Updates AhrProvider to fetch from new graphql source

benhammondmusic:
# Description and Motivation

- [x] updates DatasetId to graphql versions
- [x] updates getDatasetId()
- [x] updates metadata
- [x] updates tests
- [x] updates MetricConfigs to use count metrics

## Has this been tested? How?

- [x] renders national and state level data
- [x] render data that makes sense (not weird ALL way higher and pct_share not making sense
- [x] calculates shares for all race groups, not just split between white nh and hispanic
- [x] voter and preventable working
- [x] rerun pipelines on dev
- [x] rerun pipelines on prod 

## Types of changes

(leave all that apply)

- New content or feature
- Refactor / chore


benhammondmusic:
Need to fix issue; pct_share of sub pop. only calculating values for White NH and Hispanic

Screenshot 2024-07-08 at 2 06 22 PM

benhammondmusic:
> Need to fix issue; pct_share of sub pop. only calculating values for White NH and Hispanic
> 
> 
> 
> Screenshot 2024-07-08 at 2 06 22 PM
> 
> 

Fixed

frontend/src/data/config/MetricConfigChronicDisease.ts
eriwarr, line 1: Do these measures need the `rateDenominatorMetric`?

benhammondmusic: nice catch; a few were missing numerator configs as well. the only ones we can't do are voter (because the denominator is CITIZENS) and preventable hosp, because the denominator is MEDICARE 18+.

eriwarr (approved):
Awesome work Ben!

PR #3424, opened 2024-06-28
Landing page and explore data landing page updates

kccrtv:
# Description and Motivation
Home page:
- adds sticky scroll functionality for the how-to section to encourage user engagement
- adds mad lib animation example in how-to section to introduce users to the mad lib toolbar which is the heart of navigating the app
- moves news section to the bottom of the homepage which is a lesser priority for users to view

Explore the Data landing page:
- adds announcement banner to the top of this page to draw attention to our efforts in addressing missing and misidentified data, linking to the methodology for this
- reformats the UI on featured reports to give context of what each sample report will be displaying, as well as its clinical importance

## Has this been tested? How?
tests passing

## Types of changes
- New content or feature


benhammondmusic:
Just starting some discussions here in the comments:

First off we need to resolve all the merge conflicts.


Next, I think we should skip on this madlib explainer part for now. It looks a little strange and isn't super clear as is, and it's also duplicated in one of the explainer slides right after. I like the idea of animating the madlib text but I think we need to spend more time and would just remove this whole section

Screenshot 2024-07-02 at 5 45 33 PM

benhammondmusic:
Screenshot 2024-07-02 at 5 53 00 PM

It feels like the very top/bottom shadows missing?

benhammondmusic:
I think if we are going to go with the big data warning above the header like you added (which I think I like but is maybe vertically too much padding), I think we need to remove this other duplicate data warning.

Screenshot 2024-07-02 at 5 55 54 PM

Also, when a user closes out the data warning box at the top, I think that it should stay gone. maybe we can store that preference in session storage. there is nothing more annoying than x-ing out of pop ups and warnings and then they keep coming back

I think this redesign looks much better, but I think the eventual goal should be to try and eliminate this default topic page... I would love for the user to somehow click right into a topic when navigating from the home page.

benhammondmusic:
I'm reluctant to have the data warning persist on that actual report pages. I feel like once the user has made a selection beyond the default page this data warning could go away? 

Screenshot 2024-07-02 at 6 04 46 PM

frontend/public/img/graphics/het-hero.png
benhammondmusic, line 1: Can we be sure to optimize all of these new images? they're pretty big. Maybe you did already but https://tinypng.com/ works well to keep the quality high and reduce the kb

eriwarr:
> Just starting some discussions here in the comments:
> 
> First off we need to resolve all the merge conflicts.
> 
> Next, I think we should skip on this madlib explainer part for now. It looks a little strange and isn't super clear as is, and it's also duplicated in one of the explainer slides right after. I like the idea of animating the madlib text but I think we need to spend more time and would just remove this whole section
> 
> Screenshot 2024-07-02 at 5 45 33 PM

I resolved the merge conflicts. I agree with Ben; the Madlib explainer looks a bit weird to me. It would be better without it.

eriwarr:
> I think if we are going to go with the big data warning above the header like you added (which I think I like but is maybe vertically too much padding), I think we need to remove this other duplicate data warning.
> 
> Screenshot 2024-07-02 at 5 55 54 PM
> Also, when a user closes out the data warning box at the top, I think that it should stay gone. maybe we can store that preference in session storage. there is nothing more annoying than x-ing out of pop ups and warnings and then they keep coming back
> 
> I think this redesign looks much better, but I think the eventual goal should be to try and eliminate this default topic page... I would love for the user to somehow click right into a topic when navigating from the home page.
> 
> Maybe we can add a visualization screenshot into each of these cards, and have the cards here on this default page refer to a specific card on a specific report rather than the report generally. that it'll be more visually interesting giving the user an idea what's available, and we feature the various visualizations available.

I love the `Major gaps in the data` the message at the top of the page and agree that we should remove it from the bottom of the page now. The recommendation for the visualization screenshot in the card is great.

eriwarr (changes requested):
Looks great! Just made a few suggestions. I think once the tests are passing we're good to go.

eriwarr (approved):

PR #3434, opened 2024-07-02
Adds config file for AHR's measure IDs

### LINKED ISSUE #3436: Create a config file to store AHR measure IDs
Define a JSON structure to categorize measure IDs by demographics such as all, age, race_and_ethnicity, and sex.

eriwarr:
# Description and Motivation

This PR adds the `graphql_ahr_measure_ids.json` config file to store measure IDs categorized by demographics (`all`, `age`, `race_and_ethnicity`, and `sex`). These IDs will be used to fetch data from AHR's GraphQL endpoint. 

Closes #3436

## Has this been tested? How?
Tested locally

## Types of changes
- New content or feature


python/tests/ingestion/test_graphql_ahr_measure_ids.py
```diff
@@ -0,0 +1,70 @@
...
+    demographic (str): One of 'all', 'age', 'race_and_ethnicity', 'sex'.
+    measure_name (str): The name of the measure (e.g., 'Asthma', 'Diabetes').
+    flatten (bool): Whether to flatten the list of IDs if the result is a dictionary.
+
+    Returns:
+    list: A list of measure IDs.
+    """
+    if demographic in data:
+        for measure in data[demographic]:
+            if measure['measure'] == measure_name:
```
benhammondmusic, line 26: this might be an inefficient way to do this; you're just looking for the single `measure_name` right? so instead of iterating over all of them until one matches, you should just be able to use something like `data[demographic].get(measure_name)` or similar?

benhammondmusic (approved):
it looks good, the lookup could be made more efficient i think. im a little confused why the function under test is declared in the test file; should it be declared in an ingestion module or something and then used in the graphql_ahr datasource?

PR #3439, opened 2024-07-03
Fixes JSON loading error

eriwarr:
# Description and Motivation

This PR fixes the issue where the `graphql_ahr_measure_ids.json` file was being loaded outside of any function, causing unrelated DAGs to fail when they could not find the JSON file. It also ensures that the JSON config file is included in the Docker build process by updating the `setup.py` file.

## Has this been tested? How?
- Verified that unrelated DAGs no longer fail due to the missing JSON file on infra test.
- Tested `get_measure_ids` with both real data and test data.

## Types of changes
- Bug fix
- Refactor/chore


benhammondmusic (approved):
sweet

PR #3463, opened 2024-07-11
Adds utils for current year across multi year topics and merging/col functionality

benhammondmusic:
# Description and Motivation

- part of #3462
- Adds utils to keep most recent data in sources where that year varies per topic
- adds util to simplify merging a list of dfs and to detect if col name is one of several suffixes
- splits std_col suffixes by time view

## Has this been tested? How?

- tests added for utils

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
Looks good!

PR #3474, opened 2024-07-17
Fixes the mobile overflow issue for the `ClickableLegendHeader`

### LINKED ISSUE #3473: Clickable topic legend heading in weird spot on iOS
**Describe the bug**
The heading of the map legend isn't rendering in the right place on iOS arc (chrome based mobile browser)

image

eriwarr:
# Description and Motivation

This PR fixes #3473, where the `ClickableLegendHeader` was spilling out of its parent container on mobile devices. The problem was resolved by wrapping the `ClickableLegendHeader` component in a `div`.

## Has this been tested? How?
I ran localhost:3000 on the phone and tested the changes. 

- Mobile localhost after the fix:IMG_0283
- Prod mobile: IMG_0282

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore

### DIFF
--- a/frontend/src/charts/ClickableLegendHeader.tsx
+++ b/frontend/src/charts/ClickableLegendHeader.tsx
@@ -22,22 +22,25 @@ export default function ClickableLegendHeader(
     props.dataTypeConfig.fullDisplayName
 
   return (
-    <Tooltip
-      arrow={true}
-      placement='top'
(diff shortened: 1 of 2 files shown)


benhammondmusic (approved):
Looks great. Maybe would be better to move the new div into the component so this issue doesn't pop up in the future if we happen to use this component elsewhere?

PR #3482, opened 2024-07-19
Distinguish the `All` group on the simple bar chart

benhammondmusic:
# Description and Motivation

- makes ALL bar distinct color
- adjusts label text color as needed
- rounds all bars (on simple and compare bar charts)
- shifts the ALL down and the rest even more down to have some vertical spacing on both sides of the all
- biome linting tweaks on trailing commas and `type` import syntax

Screenshot 2024-08-28 at 12 15 47 PM

Screenshot 2024-07-22 at 1 44 23 PM

## Types of changes

(leave all that apply)

- New content or feature


kccrtv (approved):
👍🏼

PR #3498, opened 2024-07-26
Add Race Measure IDs for Non-Medical Drug Use and Update Timeout in GraphQL Fetch

eriwarr:
# Description and Motivation

- Added the race measure IDs for “Non-Medical Drug Use” to graphql_ahr_measure_ids.json.
- Updated our code to match for all relevant “Non-Medical Drug Use”  subtopics.
- Increased timeout for fetch_ahr_data_from_graphql function to 300 seconds.
- Removed 2023 data from the dataset due to so much missing data for our breakdowns. The only available measure for 2023 is non-medical drug use, which lacks age and some race breakdowns. The `preserve_most_recent_year_rows_per_topic` function uses the most recent year in the dataset to merge population data, so if data is missing for a breakdown in the most recent year, it will not send up any pop data for that row. 

## Has this been tested? How?
Checked data on BigQuery

Screenshot 2024-07-26 at 3 07 31 PM
Screenshot 2024-07-26 at 3 08 58 PM

## Types of changes
- Bug fix
- Refactor/chore


python/datasources/graphql_ahr.py
```diff
@@ -105,8 +105,12 @@ def write_to_bq(self, dataset, gcs_bucket, write_local_instead_of_bq=False, **at
             table_name = f"{demographic}_{geo_level}_{time_view}"
             topic_prefixes = [std_col.extract_prefix(rate_col) for rate_col in AHR_BASE_MEASURES_TO_RATES_MAP.values()]
             topic_prefixes.append('ahr')
+            df = df[df['time_period'].astype(int) <= 2022]
```
benhammondmusic, line 108: we should avoid "magic numbers" like this hard coded in; instead can you make a variable at the top of the file like `LAST_COMPLETE_DATA_YEAR` or similar, and then use it here? it just helps a lot to name the variable so then we remember why the year is there to begin with, and easier to remember to update it as the data changes.

benhammondmusic (approved):
nice, thanks @eriwarr

PR #3524, opened 2024-08-02
BACKEND: Splits AHR into `behavioral_health` and `non-behavioral_health` tables

### LINKED ISSUE #3523: Should split AHR topics into tables by category
Right now we have all AHR topics being written into the same table. This makes the json files big (will be huge with time series expansion) and also very hard to read/debug since each topic has up to 5-6 columns. 

We should divide the topics into tables by category. This can be accomplished by passing a new `category` argument into the `write_to_bq` function from the call in the DAG (similar to how we can pass in a `demographic` or `geographic` argument) and then run the graphql_ahr datasource on only the subset of relevant ids

## Chronic Disease

```
  'asthma',
  'cardiovascular_diseases',
  'chronic_kidney_disease',
  'copd',
  'diabetes',
```

## Behavioral Health

```
  'depression',
  'excessive_drinking',
  'frequent_mental_distress',
  'suicide',
  'substance',
```

## SDOH_PDOH

```
  'avoided_care',
  'preventable_hospitalizations',
  'voter_participation',
```

benhammondmusic:
# Description and Motivation

- closes #3523 

## AIRFLOW
- updates airflow util functions to accept a `category` property on the payload object 
- splits DAG into two files; one triggering the creation of `behavioral_health_` tables, and the other `non-behavioral_health_` tables with the remaining disease, pdoh, sdoh topics
- add handling of a category filter for the exporter step (this needs to be tested in Airflow still)

## GRAPHQL
- splits the stored json file the same way
- updates graphql util functions to accept the `category` arg and make the request using the new split json from above
- splits the test input response for national requests
- splits the test input response for state requests

## DATA SOURCE
- whenever the code was iterating over ALL rate cols, I've added little dict comprehensions that rebuild a subset dict that only includes the relevant items for the category cols being processed

## TESTS
- duplicates the test functions, testing each category split
- splits the golden data files

## LINT STUFF
- adds a new `TOPIC_CATEGORY_TYPE` that aligns with the frontend topci categories
- also added a member `non-behavioral_health` as a temporary cheat to allow us to just make 2 tables for now, the second with everything else that is not behavioral topics
- type variables earlier in the code when it's declared so code is easier to read, rather than casting it when it's used in a function call. 
- biome can't do json in vscode yet; use built in vscode
- git ignore DS Store files

## Has this been tested? How?

- tests and golden data updated and passing
- both dags run on infra-test deploy, and correctly write to the same BQ dataset onto tables by category

Screenshot 2024-08-02 at 2 31 47 PM
Screenshot 2024-08-02 at 2 32 35 PM
Screenshot 2024-08-02 at 2 32 50 PM

## Types of changes

(leave all that apply)

- Refactor / chore


python/tests/data/graphql_ahr/.DS_Store
eriwarr, line 1: Remove .DS_Store file

eriwarr (approved):
Looks great!

python/tests/data/graphql_ahr/.DS_Store
benhammondmusic: nice. weird that that wasn't already git ignored

Issue #3579, opened 2024-08-22
Backend PHRMA Cancer Infra/Pipeline Setup


benhammondmusic:
## Rename / Generate Metrics from Loaded Data

- [x] add population (denominator) shares
- [x] calc pct_share metrics
- [x] generate age_adjusted_ratio from age_adjusted rates

## Write to BQ

- [x]  use `get_bq_column_types` util to generate needed BigQuery types (you'll also need to create a list of all columns that are using number formats instead of string)
- [x]  use `add_df_to_bq()` to actually ship the table(s) from above to BigQuery
- [x]  mock out that `add_df_to_bq()` function, and assert that the function was only ran/mocked 1x per expected table being sent to BigQuery

## Add New Terraform File

- [x]  copy and rename an existing .tf file to be  `/config/data_sources/{new_source}.tf`
- [x]  find and replace (with consistent casing) the source string	- NOTE: some sources append `_data` to the filename and dataset id, but isn't required

## Update Data Sources Mapping

- [x]  Import the new class from your new data source file in `python/datasources/data_sources.py`
- [x]  Copying the existing pattern, add an item to the dict where the key is the result of your class’ `get_id()`  method, and the value is an instantiation of the class itself

## Scaffold New Airflow DAG File

- [x]  copy an existing .py airflow file to be `/airflow/dags/{new_source}.py`
- [x]  find and replace (with consistent casing) the source string
- [x]  your new `_{NEW_SOURCE}_WORKFLOW_ID` should match the return value of the `get_id()` method from your pipeline's class created in a previous step (in `/python/datasources/{new_source}.py`)
- [x]  your new `_{NEW_SOURCE}_DATASET_NAME` should match the output of that same class' `get_table_name()` method
- [x]  `default_args` remains the same for all Airflow source files
- [x]  `data_ingestion_dag` remains basically the same, just ensure the source-specific strings are updated (case-sensitive)

## Create “BQ” Payloads and Operators For Each Breakdown

- [x]  create a `{source}_bq_payload` copying the pattern from the DAG file you are using as a template
- [x]  create a `{source}_bq_operator`. This step will be responsible for actually running your `/datasource/{source}.py` class which slurps in the source data and then writes it to BigQuery tables

## Create “Exporter” Payloads and Operators For Each Breakdown

[...]

PR #3584, opened 2024-08-23
Backend PHRMA Cancer: Adds `_pct_share` cols

benhammondmusic:
# Description and Motivation

- part of #3579
- extricate `build_bq_col_types()` in util function which scans df and automatically gets the metric cols based on suffixes
- calcs pct_shares with or without "unknowns" based on the demographic breakdown and what's reflected in the source data
- updates golden data with new cols

## Has this been tested? How?

- tests passing

## Types of changes

(leave all that apply)

- New content or feature


eriwarr (approved):
Awesome!

PR #3597, opened 2024-08-27
Fix future warnings for cawp_time

benhammondmusic:
# Description and Motivation

prep for #3569 by fixing future warnings that will be errors in future versions of python and pandas, etc

## Has this been tested? How?

- golden data updated to properly null unwanted columns
- tests passing without future warnings

## Types of changes

(leave all that apply)

- Refactor / chore


python/tests/datasources/test_cawp_time.py
```diff
@@ -101,7 +102,7 @@ def _load_csv_as_df_from_web(*args, **kwargs):
     dtype = kwargs.get("dtype", {})
 
     # reverse lookup the FIPS based on the incoming url string arg
-    fips = [i for i in FIPS_TO_STATE_TABLE_MAP if FIPS_TO_STATE_TABLE_MAP[i] in url][0]
+    fips = next(fips for fips, state in FIPS_TO_STATE_TABLE_MAP.items() if state in url)
```
eriwarr, line 105: Nice!

PR #3602, opened 2024-08-28
Updates all services to use Python3.10

### LINKED ISSUE #3569: Backend: Should update all services Python 3.9 → 3.10

benhammondmusic:
# Description and Motivation

- closes #3569
- linting adjustments
- need to make both mypy and pylint ignore the airflow imports; that's because we named the directory `airflow/` which is the same as the pip package name, therefor the linter thinks we are trying to import pip package items from our own dir when they don't exist. for now just using lots of `# ignore` comments
- improve github actions to properly install mypy, deps, and types
- lots of linting fixes, incl. dealing with problems where previously we had been assigning Series directly to Dataframe without converting, etc.

## Has this been tested? How?

- [x] python unit tests passing for all covered services
- [ ] pushing new backend via `infra-test` doesn't cause any cloud errors

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
Awesome!

PR #3661, opened 2024-09-12
truncate pagination button text

### LINKED ISSUE #3649: long blog titles spilling over button
**Describe the bug**
the pagination buttons are also used for the blog articles, and long titles are spilling over / too large

**Screenshots**

Screenshot 2024-09-10 at 11 43 29 AM

kccrtv:
# Description and Motivation
closes #3649 

## Has this been tested? How?
manual

Screenshot 2024-09-12 at 12 29 37 PM
Screenshot 2024-09-12 at 12 29 55 PM

## Types of changes
- Bug fix


frontend/src/pages/News/SinglePost.tsx
```diff
@@ -307,13 +327,12 @@ export default function SinglePost() {
...
           {prevArticle && (
             <HetPaginationButton
               direction='previous'
-              onClick={() => {
-                goPrevious()
-              }}
+              onClick={goPrevious}
+              data-pagination-content='true'
             >
               {getHtml(prevArticle?.title?.rendered ?? '')}
```
benhammondmusic, line 337: we should refactor this in the future... basically the "title" coming from wordpress _should_ be a string, but sometimes authors will use like bold or italics in the title in which case it ends up sending a string of html. I think we can refactor this to just simply extract the actual text from wordpress, and then you can just apply the truncation directly here (rather than needing the useEffect and then manually manipulating the dom elements to shorten them). But for now this works even if it's more complex than I wish it had to be haha. Nice work!

benhammondmusic (approved):
a couple minor suggestions but looks great. i made a note, we should eventually refactor to just simply render the actual string of the title, and then you can truncate directly as you're rendering rather than rendering the long string and then the useEffect to manipulate the dom manually

Issue #3627, opened 2024-09-05
Should show the overall ALL rate alongside our intersectional reports


benhammondmusic:
Currently for our intersectional reports like Black Male Gun Homicides, the groups available are "All Black Men", and then the age or urbanicity groups for Black Men. It would be helpful to give context to also show the general population "All" alongside these Black Men groups, to show the impact on these marginalized intersectional communities. In many cases we could make the request within the card to the appropriate "gen pop" data set, filter to get just the "all", and then embed within the chart. 

- [x] Black Male Gun Homicide Victims
- [x] Black Women HIV Deaths, New Cases, and Prevalence

PR #3665, opened 2024-09-13
Adds comparison ALL bar for HIV Black Women 

benhammondmusic:
# Description and Motivation

- adds HIV Black Women for #3627 by updating the MetriccConfig
- allow any subpopAge to pass through to the comparison All Bar, for example all Black Women HIV topics and their gen. pop. equivalent use ages 13+

## Has this been tested? How?

passing

Screenshot 2024-09-12 at 8 32 51 PM
Screenshot 2024-09-12 at 8 33 07 PM
Screenshot 2024-09-12 at 8 33 17 PM

## Types of changes

(leave all that apply)

- New content or feature


kccrtv (changes requested):
Screenshot 2024-09-13 at 12 25 58 PM
`Ages 13+` is truncating at the laptop view and below. can you break up the span into two, so that "(All People," is on the top and "Ages 13+)" is on the bottom? 
<hr/>
additionally, could you add a note here in the definitions, and/or even on the horizontal card on what our definition of "All People" is? like we're including NH and H, as well as including Black Women, in this average? it is a bit unclear.
<br/>
Screenshot 2024-09-13 at 12 38 35 PM

benhammondmusic:
> Screenshot 2024-09-13 at 12 25 58 PM `Ages 13+` is truncating at the laptop view and below. can you break up the span into two, so that "(All People," is on the top and "Ages 13+)" is on the bottom?
> additionally, could you add a note here in the definitions, and/or even on the horizontal card on what our definition of "All People" is? like we're including NH and H, as well as including Black Women, in this average? it is a bit unclear.
> Screenshot 2024-09-13 at 12 38 35 PM

Removes parentheses, adds 3rd line to keep from truncating long text, adds footnote indicating when ALL PEOPLE means

Screenshot 2024-09-13 at 1 43 22 PM
Screenshot 2024-09-13 at 1 43 36 PM

kccrtv (approved):
Excellent! Thanks Ben!

PR #3680, opened 2024-09-20
Frontend: Remove Default classes

### LINKED ISSUE #3634: Refactor `defaultClasses`
We should consider refactoring the defaultClasses string; since it's repeated in each card and we want consistency we should at the very least declare that string globally and import it into each card rather than redeclaring it in every card component. However, I want to investigate more because there is probably a more tailwind-y way to do it, like making a new class in the tailwind config that defines the shadow and bg white and then we just apply that class everywhere instead of templating in the defaultClasses string.

benhammondmusic:
# Description and Motivation

- closes #3634 

## Has this been tested? How?

tests passing

classes weren't actually affecting the styling so can be removed

Screenshot 2024-09-20 at 12 05 17 PM

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (commented):
🤘🏿

PR #3701, opened 2024-10-01
Frontend: Update default reports to include gun deaths Map Card

benhammondmusic:
# Description and Motivation

- updates the default ExploreData options to replace multimap phrma with new gun violence report on normal map
- lazy loads custom components so that the bundle size is much smaller, and the charts don't render until needed

## Has this been tested? How?

tests passing

Screenshot 2024-10-01 at 12 06 32 PM
Screenshot 2024-10-01 at 12 07 02 PM

## Types of changes

(leave all that apply)

- New content or feature


kccrtv (approved):
sweet! 🏄🏼

PR #3740, opened 2024-11-02
fix spacing

kccrtv:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore

### DIFF
--- a/frontend/src/styles/HetComponents/HetCardExportMenuItem.tsx
+++ b/frontend/src/styles/HetComponents/HetCardExportMenuItem.tsx
@@ -8,6 +8,8 @@ interface HetCardExportMenuItemProps {
   className?: string
   children?: React.ReactNode
   iconProps?: Record<string, any>
+  spanClassName?: string
+  iconClassName?: string
 }
 
 export function HetCardExportMenuItem({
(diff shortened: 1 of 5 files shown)


frontend/src/styles/HetComponents/HetCardExportMenuItem.tsx
```diff
@@ -16,13 +18,19 @@ export function HetCardExportMenuItem({
...
   return (
     <MenuItem className={`pl-3 ${className}`} onClick={onClick}>
-      <ListItemIcon className='flex items-center px-2 py-1'>
+      <ListItemIcon
+        className={`${iconClassName} flex items-center px-2 py-1 w-full`}
+      >
         <Icon className='mx-1 w-8' {...iconProps} />
         {children && (
-          <span className='text-altBlack text-small'>{children}</span>
+          <span className={`${spanClassName} text-altBlack text-small`}>
```
benhammondmusic, line 31: To future-proof, we should try to always add these class overrides last in the string template, since tailwind classes get applied left to right. this way in the future if anyone ever hard codes a duplicate class onto this component, the override incoming prop class would take precedence

benhammondmusic (approved):
thanks, the spacing here has been an ongoing problem. thanks for cleaning it up

frontend/src/styles/HetComponents/HetCardExportMenuItem.tsx
kccrtv: oh right! nice catch, thanks ben 🔎

PR #3750, opened 2024-11-05
Frontend: Fixes A11y of `<CombinedLink />`

### LINKED ISSUE #3711: Fix a11y of `<CombinedLink />` component
**Describe the bug**

The CombinedLink component is causing the biome linter to fail, as it's breaking some a11y rules.

I have temporarily added ignore comments to allow it to pass, but we need to fix this the right way, which likely will be rendering a `<button>` component which is the proper semantic element for an interactive item, and then overriding the styles to allow it to look like the text link it is now and not like a goofy button with the browser default styles

```tsx
function CombinedLink(props: CombinedLinkProps) {
  const { to, isScrollLink, children, ...rest } = props
  const scrollToAnchor = useScrollToAnchor()

  if (isScrollLink) {
    return (
      // TODO: fix a11y
      // biome-ignore lint/a11y/useFocusableInteractive: <explanation>
      // biome-ignore lint/a11y/useSemanticElements: <explanation>
      <div onClick={() => scrollToAnchor(to)} role='link' {...rest}>
        {children}
      </div>
    )
  }
  return null
}
```

frontend/src/styles/HetComponents/HetOnThisPageMenu.tsx

benhammondmusic:
# Description and Motivation

- fixes #3711 

## Has this been tested? How?

manually checked; behavior and styling remain the same, but using correct semantic HTML and not needing ignore rules for the a11y linter

Screenshot 2024-11-05 at 1 10 58 PM

## Types of changes

(leave all that apply)

- Bug fix


kccrtv (approved):
👍🏼

PR #3759, opened 2024-11-07
Data source listings

kccrtv:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore


frontend/src/pages/DataCatalog/DataCatalogPage.tsx
```diff
@@ -58,23 +56,33 @@ export default function DataCatalogPage() {
...
-            in the charts on the Health Equity Tracker. Want to explore what
-            each data set can show us about different health outcomes?{' '}
-            <a href={EXPLORE_DATA_PAGE_LINK}>Explore the data dashboard</a>
-            <span aria-hidden={true}>.</span>
-          </p>
-        </header>
+      <section
+        id='main-content'
+        aria-labelledby='main-content'
+        tabIndex={-1}
```
benhammondmusic, line 62: this `<section>` isn't interactive so i don't think it needs any type of label or tab index; i think we can remove both lines? Also, it doesn't make sense to have the aria labelled by refer to its own id; perhaps it could point to `'main'` but even still i dont think you should aria label a container like section

benhammondmusic (approved):
i would just remove those two lines mentioned

Issue #3605, opened 2024-08-30
Solve compare views with no shared demographic type


benhammondmusic:
As our demographic options expand, we now have situations were topics don't always share a demographic breakdown type like "race and ethnicity". In this case it's unclear what we should show

one solution would be to have a fall back "All" that gets shown, since every breakdown we create now does contain rows for ALL and in general the ALL from "race" should be the same as for "SEX" "URBANICITY" etc. 

If we do this, do we just have the provider load a fallback breakdown and then only keep the "all" rows? this seems pretty doable and straightforward. It's a lot less work than insisting on generating a BQ table for EVERY dataset for EVERY potential demographic breakdown

<img width="1034" alt="Screenshot 2024-08-30 at 10 07 52 AM" src="https://github.com/user-attachments/assets/9498ecd9-69d0-47f7-b1b8-98110c7c2bdd">

benhammondmusic:
Scoped into three concrete issues after a code + browser audit (2026-07-15): #4952 (surface the fallback with an in-card notice), #4953 (extend fallback to rate chart and data table), #4954 (register existing GCS alls exports for CAWP/ACS/PHRMA/AHR and relax the CAWP gate). The fallback mechanism in resolveDatasetId already works for rate-map and rates-over-time where alls tables are registered; this issue can close once those three land. Unrestricting the compare-mode demographic selector is tracked separately in #3826.

benhammondmusic:
Closing as resolved — the solution proposed here (have the provider load a fallback breakdown and keep only the ALL rows) is exactly the ALLs fallback mechanism delivered across milestone #44.

- The fallback path is live end-to-end: `resolveDatasetId` resolves a missing demographic dataset to its `alls_` sibling, `MetricQueryResponse.usedAllsFallback` threads through, and cards render `AllsFallbackAlert`.
- The compare-views-specific piece landed in #4958: compare-topics mode now offers the **union** of each topic's supported demographics, so a topic with no matching breakdown falls back to combined ALL rates side by side with the stratified topic (closed the related #3826).

Remaining regression-coverage work is tracked separately in #4962.

PR #3766, opened 2024-11-15
Backend: Export alls table

### LINKED ISSUE #3797: Backend: Update source DAG with fallback demo     MaternalMortalityData

benhammondmusic:
# Description and Motivation

### `All`s tables

- backend for #3605 
- closes #3797
- adds step to the exporter microservice that optionally extracts the ALLS rows from a given breakdown table, removes the demographic column (both for race), and then exports an "ALLS" json file
- updates the maternal mortality DAG to ship the optional flag that should trigger this ALLS export

### Cleanup

- mutes failing backend test; pretty sure it's failing because the county over time age / sex / race giant table is incomplete (because it's so large it didnt fully download)
- adds typing to county export function

## Has this been tested? How?

- makes added ALLS json files on infra-test into storage buckets

Screenshot 2024-11-18 at 10 40 10 PM

Screenshot 2024-11-19 at 10 16 54 AM

## Types of changes

(leave all that apply)

- New content or feature


eriwarr (approved):
Looks great! Thanks Ben!

PR #3817, opened 2024-11-20
Frontend: Refactor Maternal Mortality `getDatasetId()`

benhammondmusic:
# Description and Motivation

## Has this been tested? How?

- sets up a better, less verbose way of getting requested/fallback dataset ids, by concatenating the incoming demo/geo/time and then checking if that string is a member of the type `DatasetId` that we've already explicitly declared
- this will make it so we don't have to have a million chained `if` statements to get the correct `datasetId`, making the code easier to maintain 
- refactors maternal mortality to use that new method
- cleanup and import sorting on some files

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
👍

PR #3818, opened 2024-11-21
Pipeline + Frontend: Handle fallbacks for all gun violence topics

benhammondmusic:
# Description and Motivation

### Gun Violence DAGs

- updates general pop., youth, and black men DAGs to include a fallback flag on one demographic

### Gun Violence Providers

- updates general pop., youth, and black men frontend providers to use newer datasetId resolver functions and ALLs handling

### Dataset Metadata

- the `Record` requires a metadata entry for every valid datasetId, so fills in the info without any demographic. there is probably a better solution here, maybe even programmatically generating the items based on some sort of shared config between the pipeline and the frontend

## Has this been tested? How?

- [x] tests passing frontend and backend
- [x] runs on infra-test and generates expected alls json in gcp buckets
- [x] displays new ALL fallback data for map and rates over time when comparing each against a topic that doesn't share a breakdown. NOTE: it's currently not possible to test some of these on the app, as the "missing" demographics are disabled which limits our ability to test the fallbacks. We should explore opening up the disabled demographics and prompting the user with available demographics alongside the fallback ALL data if used in compare more. need to think this through more

Screenshot 2024-11-21 at 10 06 40 AM
Screenshot 2024-11-21 at 10 07 53 AM
Screenshot 2024-11-21 at 10 53 42 AM

## Types of changes

(leave all that apply)

- New content or feature
- Refactor / chore


eriwarr (approved):
This looks great! Thanks Ben!

PR #3824, opened 2024-11-22
removing axios

eriwarr:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore

### DIFF
--- a/run_gcs_to_bq/Dockerfile
+++ b/run_gcs_to_bq/Dockerfile
@@ -8,9 +8,11 @@ ENV PYTHONUNBUFFERED True
 
 # Define the build-time variable
 ARG AHR_API_KEY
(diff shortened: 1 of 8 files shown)


run_gcs_to_bq/Dockerfile
benhammondmusic, line 1: why do we need to provide the open ai key to the run_gcs_to_bq service?

data_server/Dockerfile
benhammondmusic, line 1: why do we need to send the openai key to the data_server? it looks like we are hitting the frontend server (the express app) directly from the frontend, and it's not reading data via the django data_server (like our normal data set does)...

benhammondmusic (approved):
Awesome!

PR #3836, opened 2024-11-26
Share your story refresh

kccrtv:
# Description and Motivation
continues with news page refresh. condenses the submission guidelines and adds numbering for easy reference. changed the button to secondary, and included buttons both at the top and bottom of the page.

## Has this been tested? How?
tests passing

mobile
desktop

## Types of changes
- New content or feature
- Refactor / chore

### DIFF
--- a/frontend/src/pages/News/ShareYourStory.tsx
+++ b/frontend/src/pages/News/ShareYourStory.tsx
@@ -1,138 +1,90 @@
-import HetCTABig from '../../styles/HetComponents/HetCTABig'
+import { Helmet } from 'react-helmet-async'
+import { submissionGuidelines } from './ShareYourStoryContent'
+import { useIsBreakpointAndUp } from '../../utils/hooks/useIsBreakpointAndUp'
+import CardLeftNumber from './CardLeftNumber'
+import HetButtonSecondary from '../../styles/HetComponents/HetButtonSecondary'
 
 export default function ShareYourStory() {
+  const isMdAndUp = useIsBreakpointAndUp('md')
   return (
-    <div className='flex w-full items-center justify-center text-start'>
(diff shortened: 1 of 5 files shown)


frontend/src/pages/News/ShareYourStory.tsx
```diff
@@ -1,138 +1,93 @@
...
+              const isDesktopShadow =
+                isMdAndUp &&
+                ((Math.floor(index / 2) % 2 === 0 && index % 2 === 0) ||
+                  (Math.floor(index / 2) % 2 !== 0 && index % 2 !== 0))
 
-        <div className='m-10 flex w-full items-center'>
-          <div className='flex-1 border-0 border-t border-solid border-altGrey'></div>
-        </div>
+              return (
+                <div
```
benhammondmusic, line 60: this should be the `<li>` here, the direct child of the `<ul>` and then inside the CardLeftNumber component that can be a div or whatever. The a11y tester should have flagged that

frontend/src/styles/HetComponents/HetButtonSecondary.tsx
benhammondmusic, line 1: this is fine for now but i still think we should extract the components. the nested ternary on line 28 and all the conditionals make this really confusing to figure out what's happening

benhammondmusic (approved):
visually looks awesome. make sure to switch the nesting og the ul > div  li as uls should only contain ol or li children. also i think eventually we should refactor the button/link/mail button situation because all the condition logic is really complex and prone to failure (as we've seen lol). i also added an issue to update the loading skeleton stuff to match the new layout as it's off now #3837 but that can be done later and shouldn't hold up this PR

PR #3844, opened 2024-12-02
Frontend: Fix E2E and secondary button

benhammondmusic:
# Description and Motivation

## Tests

- adds `tsc` run to the build step on CI pull request updates, this should warn the user if typescript errors have been committed and are attempting to be pushed

## `<HetButtonSecondary>`

- fixes issue where prop `text` was being sent but not declared on the component; replaced by sending in plain text as the child component like a normal HTML button
- fixes where classes were sent in with the wrong prop

## News and Stories

- removes some empty divs
- changes `#main` News and Stories to an H2 (the nav bar "Health Equity Tracker" should be the only h1 on the page for consistency)
- removes duplicate sr-only h2 (should only be needed on pages that don't have the page title as a big headline visible to users)
- adjusts blog card titles to be h3 so that headlines are tiered properly for a11y

## WIHE

- refactors the square tab select button on the WIHE into a component
- adds clickable cursor indicator when inactive tab is hovered

## Has this been tested? How?

- e2e tests passing
- typescript checking properly (confirmed that a force pushed ts error caused a check to fail on PR update)

## Types of changes

(leave all that apply)

- Bug fix
- Refactor / chore


eriwarr (approved):
Awesome!

PR #3858, opened 2024-12-05
Pipeline Backend: Implements standardized Table id function

### LINKED ISSUE #2065: Make util fn for dataset / table naming
Per , we need to be consistent with how we name our data sources and the tables within. Particularly because we have conditional logic that applies based on the strings used (like the county export step in the exporter on Airflow)

We should create python util functions that accept all the info we has about the dataset / table as args, and which generates a consistent string.

benhammondmusic:
# Description and Motivation

To more accurately create and retrieve bq tables, we'll use a utility function that's consistent across the frontend and backend to construct table ids. 

- closes #2065
- refactor all data sources except COVID and ACS_ POP to use new function
- removes use of `by_` in demographic unless it's a part of a double demo like race/age or a specific subset like black_women_by_age
- improves variable naming to use `time_view`, `demo_type` and `table_id` throughout rather than inventing a new version of that variable across every source
- updates and improves tests

TODO: After merging, need to run pipelines for:

- [ ] acs condition
- [ ] hiv
- [ ] maternal mortality
- [ ] gun v black men
- [ ] vera

## Has this been tested? How?

tests passing

## Types of changes

(leave all that apply)

- Refactor / chore


python/ingestion/het_types.py
```diff
@@ -1,46 +1,60 @@
...
+    'black_men_by_urbanicity',
+    'youth_by_race_and_ethnicity',
+    'alls',
 ]
-PHRMA_BREAKDOWN_TYPE_OR_ALL = Literal[
+
+
+def create_subset_type(*options):
+    comprehensive_options = set(get_args(COMPREHENSIVE_DEMOGRAPHIC_TYPE))
+    invalid_options = set(options) - comprehensive_options
```
eriwarr, line 27: Nice!!!

python/ingestion/gcs_to_bq_util.py
```diff
@@ -465,3 +471,17 @@ def get_bq_column_types(df, float_cols: List[str]):
         column_types[col] = BQ_FLOAT
 
     return column_types
+
+
+def make_bq_table_id(
+    demographic: COMPREHENSIVE_DEMOGRAPHIC_TYPE,
```
eriwarr, line 477: Love this!

eriwarr (approved):
Awesome work!

PR #3860, opened 2024-12-06
UI refresh for single news post page

kccrtv:
# Description and Motivation
- updates `HetTags` component to add link functionality, which allows users to filter the corresponding news post
- updates the single post formatting to remove the taviraj font via `SinglePost` and `index.css` components

## Has this been tested? How?
tests passing

desktop-news
mobile-news

## Types of changes
- Refactor / chore


benhammondmusic (changes requested):
(From slack)

design looks awesome. i dont think the linked tags are working for me though on arc (chrome); when i click on one from the single post page, it properly goes back to the main blog page and the url includes the author or tag query param, but your dropdown filters aren’t changed, theyre in the default “everything” mode

Can you remove the linking ability for now, and add a todo issue so we can remember to fix that linking functionality across all browsers in a future PR?

benhammondmusic (approved):

frontend/src/styles/HetComponents/HetTags.tsx
```diff
@@ -1,19 +1,28 @@
+interface Tag {
+  name: string
+  link?: string
+}
+
 interface HetTagsProps {
-  tags: string[]
+  tags: string[] | Tag[]
```
eriwarr, line 7: Are we able to transform plain strings into `Tags` before they reach this point?

eriwarr (approved):

PR #3864, opened 2024-12-09
Backend: Fix VERA

benhammondmusic:
# Description and Motivation

VERA DAG was causing issues reading the csv from the web. 

- checks in the csv file; not a bad idea in case VERA ever decides to take down the GitHub file we were hotlinking
- improves the reading of that file directly from `data/` and utilizing `usecols=` to ignore unused cols per demographic
- updates test files

NOTE: this file is a good candidate for refactoring to polars, as the tests now take over a minute due to the sheer size. This is tradeoff for testing the ACTUAL source data rather than a mocked version that is prone to inconsistencies with the source data and requires upkeep

## Has this been tested? How?

- tests passing
- runs on infra-test

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
👌

PR #3871, opened 2024-12-13
Small improvements to CDC Wonder data handling

eriwarr:
# Description and Motivation

Updated CDC Wonder data ingestion to:
1. Correctly generate table names for historical and current data tables (fixed bug where both tables had "_current" suffix)
2. Removed CDC Wonder test files in favor of using actual data files

## Types of changes
- Bug fix
- Refactor / chore


python/ingestion/cdc_wonder_utils.py
```diff
@@ -38,7 +38,7 @@
     # Age source groups already match needed HET groups
     std_col.RACE_CATEGORY_ID_COL: {
         'American Indian or Alaska Native': std_col.Race.AIAN_NH.value,
-        'Asian or Pacific Islander': std_col.Race.API_NH.value,
+        'Asian or Pacific Islander': std_col.Race.ASIAN_PAC_NH.value,
```
benhammondmusic, line 41: API is the one aligned with census/cdc races; ASIAN_PAC is a weird one off only used by CAWP because they lump native hawaiian elsewhere

eriwarr: Ah okay! Thanks!

python/ingestion/cdc_wonder_utils.py
```diff
@@ -139,6 +139,7 @@ def load_cdc_df_from_data_dir(
         if geo_level == NATIONAL_LEVEL:
             topic_df[state_code_col] = US_FIPS
 
+        topic_df[YEAR_COL] = topic_df[YEAR_COL].astype(int).astype(str)
```
benhammondmusic, line 142: would it be better to just add YEAR_COL as a str to the DTYPE above so it gets read in as a string from the beginning?

python/datasources/cdc_wonder.py
```diff
@@ -47,7 +47,7 @@ def write_to_bq(self, dataset, gcs_bucket, **attrs):
         df = self.generate_breakdown_df(demo_type, geo_level)
 
         for time_view in (CURRENT, HISTORICAL):
-            table_id = gcs_to_bq_util.make_bq_table_id(demo_type, geo_level, CURRENT)
+            table_id = gcs_to_bq_util.make_bq_table_id(demo_type, geo_level, time_view)
```
benhammondmusic, line 50: oh! sorry I should have caught that in the last one

benhammondmusic (approved):
nice. and how long are the tests taking now on the real data?

PR #3875, opened 2024-12-16
Fix: missing cancer rates for race/ethnicity demographic

eriwarr:
# Description and Motivation

Fixed missing `per_100k` (crude rate) values for race and ethnicity demographic breakdowns in cancer data. The rates were not being properly aggregated during the race/ethnicity combination process.

## Has this been tested? How?
- Verified that per_100k values now appear for all race/ethnicity categories
- Confirmed that other metrics (counts, percentages) remain unchanged

1. Before: 
image
2. After:
image

## Types of changes
- Bug fix


kccrtv (approved):
👍🏼

PR #3906, opened 2025-01-09
Frontend: Fix headings

### LINKED ISSUE #3904: Need to align heading levels on all pages
previously, the Health Equity Tracker name/logo in the nav bar was the h1, some pages had an "invisible" h2 for screen reader only, etc. Now we have starting drifting towards each page's title as the h1 (like "What is health equity") which is more correct from an a11y perspective, but has made things inconsistent and results in a11y failings when heading levels are properly tiered. 

we need to:
- have 1 h1 per page, which is the title of that page
- if that title is invisible for some reason, have it as sr-only but still h1
- the nav bar name/logo shouldn't be a heading at all
- sections should be h3s
- sub sections of each sections should be h4
- etc

benhammondmusic:
# Description and Motivation

- fixes #3904 
- updates e2e tests as needed
- adds e2e test coverage for all tabs of methodology and policy hubs, including a11y checks
- fixes needed a11y issues on all pages

## Has this been tested? How?

updates e2e and passing

## Types of changes

(leave all that apply)

- Bug fix
- Refactor / chore


kccrtv (approved):
👍🏼

PR #3915, opened 2025-01-16
Kff dag

### LINKED ISSUE #3794: Backend: Update source DAG with fallback demo     KFFVaccination

JDemlow:
# Description and Motivation
closes #3794

## Has this been tested? How?

Using infra-test

Screenshot (146)

Screenshot 2025-01-17 at 11 24 49 AM
Screenshot 2025-01-17 at 11 26 00 AM

## Types of changes

(leave all that apply)

- Bug fix
- Refactor / chore


benhammondmusic (approved):
nice work updating the failing tests! I had to re-add the boolean component because somehow it got lost in your subsequent commits. But I attached screenshots from BQ and the buckets (you should also be able to go on and see those and verify) and it looks like it's working!

PR #3918, opened 2025-01-20
Updated source DAG with fallback demo CDCVaccinationNational

### LINKED ISSUE #3783: Backend: Update source DAG with fallback demo     CDCVaccinationNational
- exactly one demographic types export DAG task payload should also include the boolean to trigger fallback to alls. probably default to `sex` if its available, as those tables are smaller and should run fastest

JDemlow:
# Description and Motivation
closes #3783

## Has this been tested? How?

Yes. Using pytest against the file

Screenshot (151)

Screenshot (152)

## Types of changes

(leave all that apply)

- Refactor / chore

### DIFF
--- a/airflow/dags/cdc_vaccination_national.py
+++ b/airflow/dags/cdc_vaccination_national.py
@@ -29,6 +29,7 @@ cdc_vaccination_national_bq_operator = util.create_bq_ingest_operator(
 
 cdc_vaccination_national_exporter_payload_race = {
     "dataset_name": _CDC_VACCINATION_NATIONAL_DATASET_NAME,
+    "should_export_as_alls": True,
     "demographic": "race",
 }
 cdc_vaccination_national_exporter_operator_race = util.create_exporter_operator(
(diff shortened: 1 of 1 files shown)


airflow/dags/cdc_vaccination_national.py
```diff
@@ -29,6 +29,7 @@
 
 cdc_vaccination_national_exporter_payload_race = {
     "dataset_name": _CDC_VACCINATION_NATIONAL_DATASET_NAME,
+    "should_export_as_alls": True,
```
benhammondmusic, line 32: So we actually only want to extract the ALLs from one breakdown type, like sex OR age OR race. The way you have it here it would rewrite the same ALL rows 3 times

benhammondmusic (changes requested):

airflow/dags/cdc_vaccination_national.py
JDemlow: Ah I see now. I didn't fully understand the issue as written. I'll fix that

benhammondmusic (approved):
Nice. Thanks!

PR #3923, opened 2025-01-22
Migrate unknown map to d3

eriwarr:
# Description and Motivation

This PR converts the unknowns choropleth map from Vega to D3.

## Has this been tested? How?
- Ensured tooltips display correct information, remove on mouseout, and reset when switching maps.
- Verified map rendering with the gradient legend and "no data" indicator.

## Types of changes
- New content or feature
- Refactor / chore


benhammondmusic (commented):
- need to make the legend demarkations more sparse on smaller width views 
- 
Screenshot 2025-01-22 at 3 51 23 PM
the bottom half of the map is getting cropped; not sure if it’s due to the territory bubbles?

benhammondmusic (approved):
awesome work! I think once we've moved the final rate map over to d3, we can then pursue refactoring things to share functionality between all the cards. like there should probably be one unified tooltip style for all the cards, and the info inside each can change but the component can be shared. same for any legends, axis ticks, etc etc

PR #3933, opened 2025-01-28
Frontend: Fix Vaccine Provider

### LINKED ISSUE #3811: Frontend: Update provider with fallback demo VaccineProvider
- the provider should default to the geo-appropriate ALL table, then conditionally set to a specific demographic if possible
- the provider will need to append the requested demographic column to the ALL table, with each rows value set to `All`
- NOTE: ensure `race` vs `race_and_ethnicity` in table name; we should fix this app-wide to just rely on one so it's not confusing

benhammondmusic:
# Description and Motivation

- closes #3811 
- updates frontend to look for standard tables names done in previous backend vaccine PRs
- replaces older "nested if" block to derive correct json data file name with the new function that first looks for a specific demographic table, then looks for a fallback table if it can, and finally confirms the resulting table is a valid `DatasetId`
- adds missing e2e test coverage for covid vaccine views
- safelists a flaky URL from the CDC that was throwing a false error

## Has this been tested? How?

manually, and tests are updated and passing

## Types of changes

(leave all that apply)

- Refactor / chore

### DIFF
--- a/frontend/playwright-tests/externalUrls.spec.ts
+++ b/frontend/playwright-tests/externalUrls.spec.ts
@@ -33,6 +33,7 @@ const knownFlakyUrls = [
   'https://jamanetwork.com/channels/health-forum/fullarticle/2760153',
   urlMap.southernCenterForHumanRights,
   urlMap.randGunPolicy,
+  urlMap.cdcTrans
 ]
 
 test.describe.configure({ mode: 'parallel' })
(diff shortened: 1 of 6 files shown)


JDemlow (approved):
Looks great! I believe everything makes sense to me. I can see the TODO of removing “processed” that was satisfied, and the addition of “current” to age, race, sex. I believe I understand the changes for breakdowns, and of course the test changes. Thanks for letting me look it over!

PR #3938, opened 2025-01-30
Updated source DAG with fallback demo BJSIncarcerationData

### LINKED ISSUE #3778: Backend: Update source DAG with fallback demo     BJSIncarcerationData

JDemlow:
# Description and Motivation
closes #3778

## Has this been tested? How?

Yes. Using pytest.

Screenshot (134)

## Types of changes

(leave all that apply)

- Refactor / chore


benhammondmusic (approved):
Perfect! Glad you got your config config'd!!

PR #3948, opened 2025-02-03
Updated source DAG with fallback demo GeoContext

### LINKED ISSUE #3792: Backend: Update source DAG with fallback demo     GeoContext

JDemlow:
# Description and Motivation
Closes #3792

## Has this been tested? How?

Yes. Using pytest

Screenshot (136)

## Types of changes

(leave all that apply)

- Refactor / chore

### DIFF
--- a/airflow/dags/geo_context.py
+++ b/airflow/dags/geo_context.py
@@ -26,7 +26,10 @@ geo_context_pop_bq_operator = util.create_bq_ingest_operator(
     "geo_context_to_bq", geo_context_bq_payload, data_ingestion_dag
 )
 
-geo_context_exporter_payload = {"dataset_name": _GEO_CONTEXT_DATASET_NAME}
+geo_context_exporter_payload = {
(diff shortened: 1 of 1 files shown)


benhammondmusic:
Thanks!! Sorry, I should have caught this when I was auto-generating the issues, but `geo_context` is a weird data source in that the tables are already only by `all` (not race, age, or sex). Since it's already exporting only ALL tables, we actually don't need this one to have a special ALL export step (so I sohuldn't have let this issue stay up). you did it perfectly though!

Screenshot 2025-02-03 at 4 49 34 PM

benhammondmusic:
Thanks!! Sorry, I should have caught this when I was auto-generating the issues, but `geo_context` is a weird data source in that the tables are already only by `all` (not race, age, or sex). Since it's already exporting only ALL tables, we actually don't need this one to have a special ALL export step (so I sohuldn't have let this issue stay up). you did it perfectly though!

Screenshot 2025-02-03 at 4 49 34 PM

PR #3949, opened 2025-02-04
Refactor D3 Map to Support Both Rate Charts and Unknowns Charts

eriwarr:
# Description and Motivation

This PR refactors the D3 map implementation to support both rate charts and unknowns charts. Key changes include:

- Refactored `renderMap.tsx` to handle different chart types dynamically.
- Updated` getFillColor` to adjust colors based on extremesMode for better data visualization.
- Improved `createColorScale` to internally determine the appropriate scale type (sequentialSymlog vs. quantileSequential).
- Enhanced stroke color logic to switch territory borders between gray and white based on extremesMode.

## Has this been tested? How?
Tested locally 

## Types of changes
- New content or feature
- Refactor / chore


frontend/src/cards/MapCard.tsx
```diff
@@ -405,6 +408,10 @@ function MapCardWithKey(props: MapCardProps) {
         const mapConfig = props.dataTypeConfig.mapConfig
         if (isSummaryLegend) mapConfig.min = mapConfig.mid
 
+        const updatedMapConfig = Object.assign({}, mapConfig, {
+          scheme: interpolateDarkGreen,
+        })
+
```
benhammondmusic, line 414: this is making every map the green scale, regardless of the type of scale we assigned to that dataType in the metric config. I think it's partly confusing as well because you have made a new `MapConfig` type that has a function type as the `scheme` and the old `MapConfig` has a string as the `scheme`. I would use a unique name for the new type like `D3MapConfig` and then figure out is we can make the types identical so everything plays nice

frontend/src/cards/MapCard.tsx
```diff
@@ -538,10 +545,10 @@ function MapCardWithKey(props: MapCardProps) {
                       highestLowestGroupsByFips={highestLowestGroupsByFips}
                       activeDemographicGroup={activeDemographicGroup}
                       countColsMap={countColsMap}
-                      data={displayData}
+                      data={displayData as DataPoint[]}
                       filename={filename}
                       fips={props.fips}
-                      geoData={geoData}
+                      geoData={geoData as Topology}
```
benhammondmusic, line 551: is it not possible to cast variables as the correct types when the variables are created? Rather than creating them and then using `as` when we need to send them into the component?

frontend/src/charts/choroplethMap/types.ts
```diff
@@ -111,9 +107,16 @@ export interface MapConfig {
   higherIsBetter?: boolean
 }
```
benhammondmusic, line 108: this MapConfig is the one i mentioned above; I think it'd be helpful to use a different name here (in VSCODE you can right click and "Rename Symbol" and it'll just rename all occurrences of this specific variable; better than find and replace because it won't rename the variable that has the exact same name)

I think we have to figure out how to get this one that has scheme as a function to play nice with the old MapConfig that has scheme as a string.

frontend/src/cards/MapCard.tsx
benhammondmusic: But I dont think this override should be happening here... the MapCard should ideally be left as is, and then within the new ChoroplethMap it should be retrieving the specific map scale colors from the MetricConfig

benhammondmusic (commented):
looks like there are some frontend test errors happening; maybe typescript errors?

also, the highest/lowest mode for CAWP is strange where it's assigning colors correctly to the territory bubbles, but switched for the states (yellow should be GOOD rates of women in gov, but it's showing the states with the worst rates 0% as yellow... only in highest/lowest but not in normal view)

Screenshot 2025-02-06 at 10 21 32 PM



Colors aren't perfectly synced with the legend yet, when expanding into extremes mode, you can see Oklahoma should have the lightest yellow color based on its rate, but it's shown as a slightly darker orange. almost like it's recalculating the color scale from only the remaining states but keeping the legend scale across ALL data points. I think either way is fine, like if the extremes mode gets its newly calculated scale, but we need to make sure the legend will match it accurately

Screenshot 2025-02-06 at 10 26 21 PM
Screenshot 2025-02-06 at 10 26 24 PM

benhammondmusic (approved):
awesome work! thanks for cleaning everything up and improving the legends across the board. There are some weird things still around the color scheme typing, but I think we'll be able to clean those up and simplify once we've totally switched over to the new d3 stuff (and fully removed vega)

PR #3953, opened 2025-02-07
Pipeline: Refactor `cdc_wisqars` to ingest new data

benhammondmusic:
# Description and Motivation

Adjusts `cdc_wisqars` data source to handle updated source data for gun homicides and gun suicides

- renames by race source files to match loading pattern
- uses existing WISQARS constants instead of strings
- revert to remove loading of general "gun_deaths" for now; will re-implement in future PR
- rename some variables for clarity
- move renaming of values and column names to their HET versions earlier in the code flow when possible
- handle loading of HISPANIC and UNKNOWN race rows from the ethnicity source data
- switch over tests to read in the real source data

## Has this been tested? How?

- test source data and golden data updated and passing

## Types of changes

(leave all that apply)

- Refactor / chore


eriwarr (approved):
Looks great!

PR #3963, opened 2025-02-13
Frontend: replace multimap with new D3 choropleth map

benhammondmusic:
# Description and Motivation

- swap multimap to use new d3 map
- fix z index issues for tooltips
- fix missing data spaces rendering as black instead of grey (null vs undefined)
- fix top margin across maps (unknown needs to be lower to fit the gradient legend)

## Has this been tested? How?

passing, manually checked

## Types of changes

(leave all that apply)

- Bug fix
- Refactor / chore


eriwarr (approved):
Looks great! Thanks Ben!

PR #4076, opened 2025-03-07
Removes Choropleth  Dependency on  Legend Scale Calculation

eriwarr:
# Description and Motivation

This PR removes the dependency on the Legend.tsx handleScaleChange function for scale calculations. Instead, the Choropleth Map now computes the scale internally, streamlining its logic and reducing prop dependencies.

- Removed scaleConfig prop from Choropleth Map and related components.
- Refactored createColorScale function to handle scale calculations within the map itself.
- Updated types.ts by removing scaleConfig.
- Refactored data processing to compute scale from field range directly.
- Added missing color tokens in DesignTokens.ts.
- Removed unused props in Choropleth Map components.

## Has this been tested? How?
Tested locally

Map working even without Legend.tsx applied:
image

## Types of changes
- Bug fix
- Refactor / chore


frontend/src/charts/choroplethMap/colorSchemes.ts
```diff
@@ -9,73 +10,98 @@ import type {
...
-])
-
-const D3_MAP_SCHEMES: Partial<Record<ColorScheme, (t: number) => string>> = {
-  darkgreen: interpolateDarkGreen,
-  plasma: interpolatePlasma,
-  inferno: d3.interpolateInferno,
-  viridis: d3.interpolateViridis,
-  greenblue: d3.interpolateGnBu,
-  darkred: interpolatedarkRed,
+const COLOR_SCHEMES: Record<ColorScheme, string[]> = {
```
benhammondmusic, line 13: nice

frontend/src/charts/choroplethMap/colorSchemes.ts
```diff
@@ -9,73 +10,98 @@ import type {
...
+    het.mapYouthDark,
+    het.mapYouthMid,
+    het.mapYouthLight,
+    het.mapYouthLighter,
+    het.mapYouthLightest,
+    het.mapYouthLightZero,
+  ],
 }
 
+const D3_MAP_SCHEMES: Record<ColorScheme, (t: number) => string> =
```
benhammondmusic, line 69: I think we should rename this variable or at least at a comment; it's a little confusing that "scheme" above means an array of colors, but here scheme is like, a mapping to an interpolated function I guess? it's a little hard to tell what's going on here, so if we can clarify with var names and comments I think that'll help future us :)

frontend/src/charts/choroplethMap/colorSchemes.ts
```diff
@@ -9,73 +10,98 @@ import type {
...
+  Object.fromEntries(
+    Object.entries(COLOR_SCHEMES).map(([key, colors]) => [
+      key,
+      d3.piecewise(d3.interpolateRgb.gamma(2.2), colors),
+    ]),
+  ) as Record<ColorScheme, (t: number) => string>
```
benhammondmusic, line 75: i think it's repetitive to declare the `as` type here and also the `: Record...` on line 69; will typescript let us just do one or the other ?

benhammondmusic (approved):
awesome work; looks great and should help with moving along the vega->d3 migration!  once you fix the bug with the unknowns map on that phrma topic and check similar ones it's good to merge!

PR #4173, opened 2025-03-17
Multimap new rate legend

### LINKED ISSUE #4121: Fix up Multimap to use new rate map legend
- Replace old vega.js `<Legend>` with new D3 `<RateMapLegend>`
- adjust the new component to render properly in this multimap view (triggered by incoming boolean prop). it should be a single row, full width of legend items, unless on mobile (maybe tablet?) sizes, in which case it should go back to the normal 2 col view used on the MapCard
- ensure the scale is correct; it needs to cover the full range of ALL states/counties shown across ALL demographic groups.

alinix1:
# Description and Motivation

- Replaces the deprecated Vega.js <Legend> component with the new D3-based <RateMapLegend> in the MultiMapDialog

- Enhances the RateMapLegend component with responsive functionality by implementing the useResponsiveWidth custom hook for dynamic width detection

- Updates useGetLegendColumnCount function in mapLegendUtils.ts to support multi-map mode with appropriate column counts

- Removes hardcoded dimensions and implements container-based responsive sizing

- Adjusts layout calculations and margins for better display across various screen sizes

- Improves legend item positioning and spacing for enhanced readability

- fixes #4121

## Has this been tested? How?

- locally in the browser 

### Desktop view

Image

### Mobile/responsive view

Image

## Types of changes

(leave all that apply)

- Refactor / chore


frontend/src/cards/ui/MultiMapDialog.tsx
```diff
@@ -184,19 +184,18 @@ export default function MultiMapDialog(props: MultiMapDialogProps) {
...
               data={props.data}
-              scaleType={RATE_MAP_SCALE}
-              sameDotSize={true}
-              description={'Consistent legend for all displayed maps'}
+              description={'Legend for rate map'}
               mapConfig={mapConfig}
-              stackingDirection={'horizontal'}
-              columns={6}
+              isSummaryLegend={false}
+              fieldRange={props.fieldRange}
```
benhammondmusic, line 195: nice 👏

frontend/src/charts/choroplethMap/RateMapLegend.tsx
```diff
@@ -139,21 +143,27 @@ export default function RateMapLegend(props: RateMapLegendProps) {
     // Add missing data item to special items
     if (hasMissingData) {
       specialLegendItems.push({
-        color: het.howToColor || '#cccccc', // TODO: use het color
+        color: het.howToColor || '#cccccc',
```
benhammondmusic, line 146: let remove the fallback #cccccc thing; not sure why thats there still

frontend/src/charts/choroplethMap/RateMapLegend.tsx
```diff
@@ -139,21 +143,27 @@ export default function RateMapLegend(props: RateMapLegendProps) {
...
         label: NO_DATA_MESSAGE,
         value: null,
       })
     }
 
-    // Calculate layout
+    // Calculate layout with responsive adjustments
+    // Adjust columns based on available width
+    let adjustedColumnCount = regularColsCount;
+    if (containerWidth < 500 && regularLegendItems.length > 0) {
```
benhammondmusic, line 155: in general it's best practice to avoid "magic numbers" like this 500 here; it makes it really hard for a future dev to understand exactly why the number 500 was used. One solution could be to declare a variable at the top of the file that explains why 500 like `const mobileBreakpoint = 500` and then in-code it'll make more sense why 500. 

However, in this case we already have our breakpoints defined in tailwind, and with a small change we can access them. First, add `export` to the function declaration `getTailwindBreakpointValue` inside the hook file `useIsBreakpointAndUp.tsx`

Now that that inner function is exported (which converts an existing tailwind breakpoint key like `sm` or `smMd` into the numerical pixel value), we can use it to get a number close to 500 that is already in our design system / breakpoints. 

```
const smBreakpoint = getTailwindBreakpointValue('sm')
if (containerWidth < smBreakpoint && //...
```

I think that'll work well and keep us in our existing system. However if you find 500 is truly needed you can use a descriptive variable

frontend/src/charts/choroplethMap/RateMapLegend.tsx
```diff
@@ -163,9 +173,9 @@ export default function RateMapLegend(props: RateMapLegendProps) {
     const height =
       maxItemsInAnyColumn * legendRowHeight + margin.top + margin.bottom
 
-    // Set SVG dimensions
+    // Set SVG dimensions with responsive width
     svg
-      .attr('width', width)
+      .attr('width', containerWidth)
```
benhammondmusic, line 178: cool

frontend/src/charts/choroplethMap/mapLegendUtils.ts
```diff
@@ -117,11 +117,15 @@ export const createUnknownLegend = (
   })
 }
 
-export function useGetLegendColumnCount() {
+export function useGetLegendColumnCount(isMulti?: boolean) {
   const isCompareMode = window.location.href.includes('compare')
   const isTiny = useIsBreakpointAndUp('tiny')
   const isSm = useIsBreakpointAndUp('sm')
+  const isLg = useIsBreakpointAndUp('lg')
+  if (isLg && isMulti) return 10
```
benhammondmusic, line 125: great! you can use your judgement, and if you want to adjust that down to `'md'` or potentially add some more options to this function to really make the best layout feel free

benhammondmusic (commented):
Awesome!! Nice work Ali!

Just a few things before we are ready to merge:
- it looks like biome must not be properly set up in your pre-commit. For now you can run from frontend/ `npm run cleanup` and that'll apply biome formatting to everything. then git add, commit, and repush and [that failing test on CI](https://github.com/SatcherInstitute/health-equity-tracker/actions/runs/13908065436/job/38915620982?pr=4173) should pass
- you should also go back and try to [update your pre-commit settings](https://github.com/SatcherInstitute/health-equity-tracker/actions/runs/13908065436/job/38915620982?pr=4173) so that biome will auto format on every commit. you can also install the biome vscode extension and that will help keep specific code formatting things consistent across all devs

- I made a comment re "magic numbers", see if we are able to either better define why 500 is used, or ideally stick with an existing breakpoint value. We could pursue extending the `useIsBreakpointAndUp` hook to accept an optional ref as well, and then it could use the same width detection logic on a container rather than the full screen... that'd be cool

other wise looks awesome! I think we can continue to tweak it, where given certain combinations of screen width, multimode, compare mode, etc, 2 col of normal data might look best, and work on centering all the items etc. but that can be done in future PRs as well. 

Once we fix the "magic" number, and fix the biome formatting I'm happy to merge it!

eriwarr (approved):
Great job Ali!!!

PR #4202, opened 2025-04-02
Fix missing percent share and estimated total values in AHR age tables

eriwarr:
# Description and Motivation

This PR adds missing percent share and estimated total calculations to the AHR age tables. This fixes the issue where "Share of all adult asthma cases" and similar metrics were missing on PROD.

- Fixed the post-processing method to properly calculate and include estimated total values for all metrics
- Added percent share calculations for both all-ages metrics (like suicide) and 18+ metrics (like asthma, depression, etc.)

## Testing
- Verified that the share percentages for each age group sum to 100% for each metric
- Confirmed that all age breakdowns now display proper percentage values instead of alert icons

PROD:
image

FIX:
image

## Types of changes
- Bug fix
- Refactor / chore


benhammondmusic (approved):
Awesome!

PR #4390, opened 2025-04-30
Handles openai rate limits and adds server side caching

### LINKED ISSUE #4378: Implement Caching and Rate Limit Handling for AI Insights
## Is your feature request related to a problem?
Currently, our AI insight generation doesn't implement caching or handle API rate limits. This means we're making redundant API calls for similar data patterns and risk service disruption if we exceed OpenAI's free tier limits.

## Did someone outside of the dev team ask for this? Who are they?
This request came from Ben Hammond, who raised concerns about OpenAI API usage limits and the need for caching similar insights.

1. Implement server-side caching for AI insights:
- Store insights by a hash of the prompt data
- Set appropriate expiration time (24 hours)
- Return cached results when available instead of making API calls

## Which of our [user persona(s)](https://docs.google.com/document/d/1EASpK_THTE_uy_Yk0sut2GTVd3w5pKtKH7LpLtF-0co/) would this feature primarily benefit?

- Data Diver: contributes to the body of knowledge surrounding health equity (e.g. analysts, researchers, professors in academic or corporate settings)
- Informer: helps everyday people understand important current events (e.g. journalists working in storytelling settings)
- Persuader: passage of legislation by presenting concise and compelling arguments (e.g. staff members, policy advisors in legislative or nonprofit settings)

eriwarr:
# Description and Motivation

Added server-side caching for AI insights to reduce API calls to OpenAI, prevent rate limit issues, and improve response times for repeated queries. Closes #4378. 

## Has this been tested? How?
I've manually tested this by clicking the same button multiple times and seeing that it doesn't make a new API call each time but instead uses the cached response. I've set a reminder to check the 24 hour logic. 

## Types of changes
- New content or feature
- Refactor / chore


frontend/src/cards/ui/InsightDisplay.tsx
```diff
@@ -33,42 +29,67 @@ const InsightDisplay: React.FC<InsightDisplayProps> = ({
...
+      rateLimitReached
+    )
       return
 
     setIsGeneratingInsight(true)
     try {
       const newInsight = await generateInsight({ knownData, metricIds })
       setInsight(newInsight)
+    } catch (error) {
+      if (error instanceof Error && error.message.includes('rate limit')) {
```
benhammondmusic, line 63: this feels not quite right; since everything is typed can we match an exact error message? any time we do sub-string matching I think it opens us up to false positives and other potential issues

frontend_server/server.js
```diff
@@ -105,12 +106,30 @@ app.use('/api', apiProxy)
...
 app.use(compression())
 
+app.get('/rate-limit-status', (req, res) => {
+  res.json({
+    rateLimitReached: RATE_LIMIT_REACHED,
+  })
+})
+
+const aiInsightCache = new Map()
+const CACHE_TTL_MS = 24 * 60 * 60 * 1000
```
benhammondmusic, line 116: this is good for now; we should discuss with @kccrtv  and maybe @alinix1 and @JDemlow as well and see what would feel best for a user experience here. maybe it could be longer than 24 hours if we end up adding some other UI for insight stuff like regenerating a response, or user giving a thumbs up/down to the AI response, etc

frontend_server/server.js
```diff
@@ -130,17 +149,34 @@ app.post('/fetch-ai-insight', async (req, res) => {
       },
     )
 
+    if (aiResponse.status === 429) {
+      RATE_LIMIT_REACHED = true
+      throw new Error('Rate limit reached')
```
benhammondmusic, line 154: is this what that other substring check is looking for? maybe we can return the actual error code and check for the specific code ?

frontend_server/server.js
```diff
@@ -130,17 +149,34 @@ app.post('/fetch-ai-insight', async (req, res) => {
...
+      timestamp: now,
+    })
 
-    res.json({ content: content.trim() })
+    res.json({ content: trimmedContent })
   } catch (err) {
     console.error('Error fetching AI insight:', err)
-    res.status(500).json({ error: 'Failed to fetch AI insight' })
+
+    if (err.message.includes('Rate limit')) {
```
benhammondmusic, line 175: same, using `.includes` feels a little messy and error prone here?

PR #4393, opened 2025-04-30
No data label update map

### LINKED ISSUE #4127: "no data" areas show same label as unknown % areas
**Describe the bug**
The label for areas with ⬜️ `no data` show the same label as the areas with `% unknown: 0%`. Having "no data" is not the same as suppressed data (refer to PR #3947)

**To Reproduce**
Example viz in screen recording: https://healthequitytracker.org/exploredata?mls=1.medicare_cardiovascular-3.21&group1=All&dt1=beta_blockers_adherence#unknown-demographic-map
- Click on `❸ Unknown demographic map`

**Expected behavior**
Label should indicate "no data" for the gray zones.

**Screenshots**

 - Device: Apple M2 Macbook Pro 
 - OS: Sequoia 15.3.1
 - Browser: Chrome
 - Version: "Chrome is up to date" Version 134.0.6998.89 (Official Build) (arm64)

**Additional context**
https://github.com/user-attachments/assets/09ac52ad-5fa9-4374-bfb2-6309fe6e395b

alinix1:
# Description and Motivation

- Separates the checks for different data conditions (null vs. zero)
- Transforms the data to explicitly mark null values as "no data" cases
- Converts null values to undefined before passing it to the ChoroplethMap component

- fixes #4127 

## Has this been tested? How?

- locally in the browser

Screenshot 2025-04-30 at 3 43 31 PM

Screenshot 2025-04-30 at 3 43 47 PM

## Types of changes

(leave all that apply)

- Bug fix


frontend/src/cards/UnknownsMapCard.tsx
```diff
@@ -185,9 +185,14 @@ function UnknownsMapCardWithKey(props: UnknownsMapCardProps) {
...
-              unknown[metricConfig.metricId] === 0 ||
-              unknown[metricConfig.metricId] == null,
+            (unknown: HetRow) => unknown[metricConfig.metricId] === 0,
+          )
+
+        // for data sets where some geos might contain null
+        const unknownsAllNull =
+          unknowns.length > 0 &&
+          unknowns.every(
+            (unknown: HetRow) => unknown[metricConfig.metricId] === null,
```
benhammondmusic, line 195: i think this variable is unused; probably can remove this chunk of code.

frontend/src/cards/UnknownsMapCard.tsx
```diff
@@ -222,7 +227,16 @@ function UnknownsMapCardWithKey(props: UnknownsMapCardProps) {
...
+                  data={
+                    unknowns.map((unknown) => {
+                      const value = unknown[metricConfig.metricId]
+                      return {
+                        ...unknown,
+                        [metricConfig.metricId]:
+                          value == null ? undefined : value,
+                      }
+                    }) as DataPoint[]
+                  }
```
benhammondmusic, line 239: I don't think this is needed, since you're specifically checking for either `undefined` or `null` in tooltipUtils; try just using `data={unknowns}` here and see if everything is working as expected

frontend/src/charts/choroplethMap/tooltipUtils.ts
```diff
@@ -35,7 +35,7 @@ export const formatMetricValue = (
   metricConfig: MetricConfig,
   isLegendLabel?: boolean,
 ): string => {
-  if (value === undefined) return 'no data'
+  if (value === undefined || value === null) return 'no data'
```
benhammondmusic, line 38: nice! this is the most explicit way and I think we should keep it like this. just FYI in case you see it elsewhere, sometimes people write `value == null` and that means the same thing, the _double_ equal loosely will evaluate for undefined or null

frontend/src/cards/UnknownsMapCard.tsx
benhammondmusic: unless we want to use it to prevent an entirely gray "unknowns map" from rendering which might be good?

benhammondmusic (changes requested):

frontend/src/cards/UnknownsMapCard.tsx
alinix1: Should we keep the variable for consistency with other checks and integrate it into visualization logic, or remove it as unused code? I wonder if there is an instance at the geo state level for an entirely gray "unknowns map"?

benhammondmusic: let's just solve one problem at a time (the no data label) so remove this chunk of code since it's not being used. if we decide later to add a feature when blank maps are suppressed we can pursue it at that time

benhammondmusic (approved):
Awesome ! Thanks so much!

PR #4436, opened 2025-05-19
Handle sparse time-series data with dashed lines

### LINKED ISSUE #2148: Single year time-series HIV looks weird
Screen Shot 2023-04-24 at 10 27 56 AM

benhammondmusic:
# Description and Motivation

- fixes #2148
- isolated years get a short but solid, full opacity line chunk 
- data points that are "broken" lines always get a dot (complete lines still only get data point dots on hover)
- gaps in sparse data are filled with lower opacity dashed lines, allowing user to track progress over time but understand data isn't fully available
- accounts for expected gaps in election data (every 4 years for now). this can be improved later if we need to handle data that's only every 2 years etc.
- the black "ALL" lines are rendered last, making them appear "on top" and prioritized over any specific group line that might have the same data points
- decomposes the trend chart components, encapsulating svg chunks into better named react components that should make the hierarchy and ongoing dev easier

## Has this been tested? How?

Manually on various screens and topics

image
image
image

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore


frontend/src/charts/trendsChart/SparseDataInterpolationLine.tsx
```diff
@@ -0,0 +1,57 @@
...
+        const yearDiff =
+          new Date(firstPoint[0]).getFullYear() -
+          new Date(lastPoint[0]).getFullYear()
+
+        if (yearDiff <= 1) return null
+
+        return (
+          <g key={`gap-${index}`}>
+            <StyledPath
+              d={lineGen([lastPoint, firstPoint] as any)}
```
eriwarr, line 46: Are we able to type as it's actual value instead of using any?

frontend/src/charts/trendsChart/SparseDataInterpolationLine.tsx
```diff
@@ -0,0 +1,57 @@
...
+              color={color || ''}
```
eriwarr, line 47: Can we use a default color here instead of the empty string?

eriwarr (approved):
Awesome!

PR #4438, opened 2025-05-20
Fix missing years of data in acs_condition

benhammondmusic:
# Description and Motivation

acs_condition was missing years of data; only 1 out of each concurrent block of years was writing correctly. this turned out to be because we were trying to store the incoming "year" attribute as self.year and then referencing that in various methods (more OOP style). somehow this made ALL concurrent instantiations of the data source refer to the same single year, rather than each having its own year. 

- fixed by using a more functional approach, simply passing `"year"` as an arg rather than using `self.year`

## Has this been tested? How?

tests updated with new signature for function

image

## Types of changes

(leave all that apply)

- Bug fix


eriwarr (approved):
Awesome job!

Issue #4447, opened 2025-05-27
Gun Violence AECF Grant Deliverables (2nd Addendum)


benhammondmusic:
## **Second Addendum to Scope of Work: Expanding Gun Violence Data and Local Community Insights on the Health Equity Tracker**

**Objective:**
To enhance the Health Equity Tracker's (HET) capabilities by integrating comprehensive gun violence data with a focus on local, granular, and demographically stratified information nationally and novel visualizations of health disparities in the Atlanta metropolitan area.

## **Deliverables:**

**1. Gun Violence Data Integration into HET:**

**Purpose:** Integrate County Health Rankings data and the [CDC Injury and Violence Data Dashboard](https://www.cdc.gov/injury-violence-data/data-vis/index.html) to expand our comprehensive gun violence reporting capabilities.

- [x] Gun Deaths Overall (CDC Wisqars): national + state X race X current + historical 
- [x] Gun Deaths Overall (CHR): county X race X current + historical 
- [x] Gun Suicide (New CDC Mapping Injury, Overdose, and Violence Dashboard): county X alls X current + historical
- [x] Gun Homicide (New CDC Mapping Injury, Overdose, and Violence Dashboard): county X alls X current + historical

**3. Atlanta-Focused Visualization:**

**Purpose:** Create a city-focused view of Atlanta’s counties, exposing geographic and demographic inequities within the city itself for our gun violence topics and potentially all county-level topics the Health Equity Tracker presents.
- [x] feature to focus on five core urban counties (Fulton, DeKalb, Cobb, Gwinnett, and Clayton) allowing the user to focus on Atlanta vs. Georgia as a whole

**4. Community Engagement:**

**Purpose:**  Consult and honor those with lived experience when integrating any new data

- [x] Maintain membership with Atlanta-centered Violence Prevention Committee alongside 100 Black Men of Atlanta and the King Center to ensure data relevance and incorporate community insights.
- [ ] Host one focused online working session demonstrating the full set of Health Equity Tracker gun violence topics, inviting community members, clinicians, academics, and youth representatives working in gun violence prevention.

**5. HET Impact Documentation:**

**Purpose:** Document references, presentations, teaching applications, and media mentions in a continuously updated format that demonstrates impact to funders and stakeholders.

- [ ] Design Mockups for "HET Impact" section on the HET (running most recent list on homepage and comprehensive list on subpage)
[...]

PR #4450, opened 2025-05-28
Datasource/miovd

eriwarr:
# Description and Motivation

- Adds a new data source for obtaining county-level gun suicide and homicide data from the CDC Mapping Injury, Overdose, and Violence Dashboard (MIOVD).
-  Addresses parts of issue #4447, fulfilling the requirement to support MIOVD as a new data source.

## Has this been tested? How?
current dataframe:
image
historical dataframe:
image

## Types of changes
- New content or feature

### DIFF
--- a/frontend/src/pages/DataCatalog/DataCatalogPage.tsx
+++ b/frontend/src/pages/DataCatalog/DataCatalogPage.tsx
@@ -58,7 +58,7 @@ export default function DataCatalogPage() {
       <title>Data Downloads - Health Equity Tracker</title>
       <section
         id='main-content'
-        className='mx-auto flex w-svw max-w-lgplus flex-col justify-center px-16 py-16 md:px-24 lg:px-56'
+        className='mx-auto flex w-svw max-w-lgplus flex-col justify-center px-2 py-16 sm:px-16 md:px-24 lg:px-56'
       >
         <h1
           id='main'
--- a/frontend/src/pages/Landing/LandingPage.tsx
+++ b/frontend/src/pages/Landing/LandingPage.tsx
@@ -105,7 +105,7 @@ function LandingPage() {
     <main className='relative' aria-label='Main Content'>
       <title>Home - Health Equity Tracker</title>
 
-      <section className='relative min-h-5/6 overflow-hidden px-56 xs:px-16 py-16 sm:px-16 md:px-24'>
+      <section className='relative min-h-5/6 overflow-hidden px-4 py-16 sm:px-16 md:px-24'>
         <img
           src='/img/graphics/het-hero.png'
           alt='various charts from the health equity tracker'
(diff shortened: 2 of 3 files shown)


python/datasources/cdc_miovd.py
```diff
@@ -0,0 +1,139 @@
...
+        suicides_df = self.load_condition_data("gun_violence_suicide")
+
+        # Merge homicide and suicide data
+        df = merge_utils.merge_dfs_list([homicides_df, suicides_df], self.MERGE_COLS)
+        df = merge_utils.merge_state_ids(df)
+        df = merge_utils.merge_county_names(df)
+
+        # Split into annual and TTM dataframes using the is_ttm flag
+        annual_df = df.copy().drop(columns=["is_ttm"])
+        ttm_df = df[df["is_ttm"]].copy().drop(columns=["is_ttm"])
```
benhammondmusic, line 86: i think this might be wrong; we need to ensure that the TTM rows are filtered out from the annual df as well, otherwise there will be additional entries in our historical calculations etc. Right now the annual_df still includes the TTM rows, you're only dropping the ttm boolean column. i think we need something like

```
annual_df = df[~df["is_ttm"]].copy().drop(columns=["is_ttm"])  # Only non-TTM rows
ttm_df = df[df["is_ttm"]].copy().drop(columns=["is_ttm"])      # Only TTM rows
```

python/datasources/cdc_miovd.py
```diff
@@ -0,0 +1,139 @@
...
+
+        for time_view, df_for_bq in [(CURRENT, ttm_df), (HISTORICAL, annual_df)]:
```
benhammondmusic, line 88: nice

all_county_current.csv
benhammondmusic, line 1: these aren't supposed to be checked in are they?

python/datasources/cdc_miovd.py
```diff
@@ -0,0 +1,139 @@
...
+
+from datasources.data_source import DataSource
+from ingestion import dataset_utils, gcs_to_bq_util, merge_utils
+from ingestion import standardized_columns as std_col
+from ingestion.constants import CURRENT, HISTORICAL
+
+
+class CDCMIOVDData(DataSource):
+    # MIOVD constants
+    CONDITIONS = ["gun_violence_homicide", "gun_violence_suicide"]
```
benhammondmusic, line 30: could use std_col.GUN_VIOLENCE_HOMICIDE_PREFIX etc here and everywhere these strings are used if we wanted to stay consistent?

python/datasources/cdc_miovd.py
```diff
@@ -0,0 +1,139 @@
...
+import pandas as pd
+
+from datasources.data_source import DataSource
+from ingestion import dataset_utils, gcs_to_bq_util, merge_utils
+from ingestion import standardized_columns as std_col
+from ingestion.constants import CURRENT, HISTORICAL
+
+
+class CDCMIOVDData(DataSource):
+    # MIOVD constants
```
benhammondmusic, line 29: this comment is confusing to me; usually we do this term/pattern to declare the source data constants, like `FA_Homicide` etc. Maybe we can remove this comment or change it to like, "File loading constants" or something?

benhammondmusic (commented):
a few questions / comments

PR #4494, opened 2025-08-26
Fix CAWP test by adding User-Agent header for Wikipedia API

eriwarr:
# Description and Motivation

## Types of changes

(leave all that apply)

- Bug fix
- New content or feature
- Refactor / chore


kccrtv (approved):
👍🏼

PR #4560, opened 2025-12-08
Add CDCHIVBlackWomenData datasource and workflow

eriwarr:
# Description and Motivation

This PR creates a dedicated pipeline for CDC HIV data specific to Black women, separating it from the general CDC HIV data pipeline. 

## Has this been tested? How?
-  All unit tests pass locally and in CI
- Added state level tests to golden data dir:  `black_women_by_age_state_current`, `black_women_by_age_state_historical`
- Confirmed data for new tables match data from tables populated by `cdc_hiv_data` pipeline.

image

## Types of changes
- Refactor / chore
- New content or feature

### DIFF
--- a/python/datasources/cdc_hiv_black_women.py
+++ b/python/datasources/cdc_hiv_black_women.py
@@ -6,6 +6,13 @@ from ingestion.het_types import GEO_TYPE, HIV_BREAKDOWN_TYPE
 from ingestion.cdc_hiv_utils import (
     load_atlas_df_from_data_dir,
     BREAKDOWN_TO_STANDARD_BY_COL,
+    CDC_AGE,
+    CDC_POP,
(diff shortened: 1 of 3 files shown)


python/datasources/cdc_hiv_black_women.py
```diff
@@ -0,0 +1,116 @@
...
+    get_bq_col_types,
+)
+from ingestion.dataset_utils import (
+    generate_pct_share_col_without_unknowns,
+    generate_pct_rel_inequity_col,
+    preserve_only_current_time_period_rows,
+)
+from typing import cast
+
+CDC_AGE = "Age Group"
```
benhammondmusic: these are already defined in the cdc hiv utils right? can we just import them like we are in the normal hiv file?

PR #4566, opened 2025-12-09
FRONTEND: append FIPS to dataset ID in GunViolenceProvider logic

benhammondmusic:
# Description and Motivation

now that gun violence provider handles county level data (MIOVD for homicide/suicide and CHR for overall deaths) we need to ensure the frontend is loading the MUCH smaller state-specific county level files rather than the entire county level file with every county from every state. this has a major effect on the amount of data transferred and the speed it reaches the user

## Has this been tested? How?

tests updated to handle county level loading

Screenshot 2025-12-09 at 1 46 20 PM

## Types of changes

(leave all that apply)

- Bug fix


eriwarr (approved):
Nice!

PR #4617, opened 2026-03-13
Bump checkout action v4 -> v6

benhammondmusic:
# Description and Motivation

addresses a warning re: actions running on Node 20 which will be replaced by node 24 soon. this action was causing the warning. updates to the recent version 

## Has this been tested? How?

need to merge and ensure normal infra functionality

image

## Types of changes

(leave all that apply)

- Refactor / chore


kccrtv (approved):
👍🏼

PR #4651, opened 2026-04-15
Hide AI insight button on cards with no data

### LINKED ISSUE #4645: Insight button should be disabled when chart has no data
**Describe the bug**
When a chart renders an empty state (e.g., "No unknown values for sex reported in this dataset at the state/territory level"), the AI insight button is still active and can generate an insight — even though there is no data to synthesize.

**To Reproduce**
1. Navigate to Gun · Homicides · United States · by Sex at Birth
2. Scroll to the "Unknown demographic map" card
3. Card shows empty state: "No unknown values for sex reported in this dataset 
   at the state/territory level"
4. AI insight button is still visible and clickable

**Expected behavior**
The insight button should be hidden or disabled when the card is in an empty/
no-data state.

**Screenshots**
Image

Image

**Additional context**

eriwarr:
# Description and Motivation

- Hides the AI insight button on cards that have no data to analyze
- CardWrapper now passes a `setHasData` callback to children, allowing each card to signal whether it has meaningful data using its existing condition checks
- Cards that are in empty/missing data states (e.g., "Graph unavailable", "No unknown values") no longer show the insight button

Closes #4645

## Has this been tested? How?
Tested manually by navigating to data topic/demographic combinations that produce empty card states. Verified the insight button is hidden on cards showing "Graph unavailable", "No unknown values", and missing data alerts. Also verified the button still appears and functions normally on cards with valid data.

Screenshot 2026-04-15 at 1 14 56 PM

## Types of changes
- Bug fix


frontend/src/cards/CardWrapper.tsx
```diff
@@ -71,6 +71,25 @@ function CardWrapper(props: {
...
+        // Default: check if any query has non-missing data for its metrics
+        let hasData = queryResponses.some(
+          (response, i) =>
+            !response.shouldShowMissingDataMessage(
+              (props.queries ?? [])[i]?.metricIds ?? [],
+            ),
+        )
+
+        // Evaluate children first so cards can override via setHasData
+        const setHasData = (value: boolean) => {
```
benhammondmusic: small change / suggestion, the word `set` is confusing to me here because it can be a verb like you are using it but it also can be set as in "dataset" and that was confusing to me at first. Even changing to `updateCardHasData` makes it more clear in my mind that it is a function that takes a boolean arg and then overrides the true/false as to whether the particular card contains useable data. And since it's really an override that reports back after evaluating the fetched dataset, maybe something like `overrideCardHasData()` and `const cardHasData: boolean` would be the most clear?

frontend/src/cards/CardWrapper.tsx
```diff
@@ -81,7 +100,10 @@ function CardWrapper(props: {
                 AI Insight
               </span>
             )}
-            <InsightVisualizationButton scrollToHash={props.scrollToHash} />
+            <InsightVisualizationButton
+              scrollToHash={props.scrollToHash}
+              hasData={hasData}
```
benhammondmusic: it seems better practice to conditionally render a simpler component, rather than shipping in a prop that only handles non-rendering inside. so in other words:

`{hasData && <InsightVisualizationButton scrollToHash={props.scrollToHash} />}` and remove the hasData prop from the button component. It doesn't really change the way the component renders or operates, only whether it gets rendered at all.

frontend/src/cards/ui/InsightVisualizationButton.tsx
```diff
@@ -7,15 +7,17 @@ import { cardInsightOpenAtom } from '../../utils/sharedSettingsState'
 
 interface InsightVisualizationButtonProps {
   scrollToHash: ScrollableHashId
+  hasData?: boolean
```
benhammondmusic: see comment below; in general conditionals should be handled further up the chain, in this case use the `hasData` boolean from the parent to conditionally render this component, and no need to add this prop and handle the conditional within the component

benhammondmusic (commented):
great! a few suggestions on best practice for conditionals / simpler components / naming ideas that I think will make the feature clearer for future us, but overall this is great. I think we'll be able to leverage this feature later to hide unhelpful empty cards altogether, not just hiding the LLM feature

benhammondmusic (approved):

PR #4710, opened 2026-05-18
Improve design token flow with build step

benhammondmusic:
# Description and Motivation
Adds a W3C DTCG token build pipeline. Colors, typography, and dimensions are defined once in JSON and generated into typed TS + CSS files on install/build.

- `tokens/*.tokens.json` → `src/styles/tokens/*.{ts,css}` via `tsx run-tokens.ts`
- Deletes the four hand-maintained theme files and `cssVarUtils` they replace
- `muiTheme.tsx` imports typed token objects instead of inline string literals
- `useIsBreakpointAndUp` uses generated `breakpoints` instead of a hardcoded duplicate map

**Token API — plain values, no CSS var wrappers:**
```ts
import { colors }                  from '../../styles/tokens/colors'
import { typography }              from '../../styles/tokens/typography'
import { dimensions, breakpoints } from '../../styles/tokens/dimensions'

colors.altGreen          // '#0b5240'
typography.fontSansText  // "'Inter Variable', sans-serif"
dimensions.radiusSm      // '4px'
breakpoints.sm           // '600px'
```

CSS vars are a Tailwind implementation detail — `@theme` registers tokens so utility classes (`bg-alt-green`, `font-sans-text`, `sm:`) work. App code never references `var(--color-*)` directly.

Pre-PRs landed first to keep this diff focused: #4711 #4712 #4713 #4714 #4715

## Has this been tested? How?
- TypeScript + Playwright E2E pass
- `npm ci` passes on CI
- `npm run tokens` regenerates all 6 files cleanly from cold start

## Types of changes
- Refactor / chore


frontend/tokens/dimensions.tokens.json
```diff
@@ -0,0 +1,119 @@
...
+    "$type": "string",
+    "cardGutter":  { "$value": "8px" },
+    "footer":      { "$value": "10rem" },
+    "socialIcon":  { "$value": "7px" }
+  },
+
+  "strokeWidth": {
+    "$type": "string",
+    "trendLineNormal": { "$value": "2.5" },
+    "trendLineThick":  { "$value": "5.5" }
```
eriwarr, line 64: should these values have units?

frontend/tokens/dimensions.tokens.json
```diff
@@ -0,0 +1,119 @@
...
+    "lg":     { "$value": "1280px" },
+    "lgplus": { "$value": "1440px" },
+    "xl":     { "$value": "1920px" }
+  },
+
+  "container": {
+    "$type": "string",
+    "tiny":   { "$value": "350px" },
+    "sm":     { "$value": "600px" },
+    "smplus": { "$value": "768px" },
```
eriwarr, line 83: `smPlus`

frontend/tokens/dimensions.tokens.json
```diff
@@ -0,0 +1,119 @@
...
+    "md":     { "$value": "960px" },
+    "lg":     { "$value": "1280px" },
+    "lgplus": { "$value": "1440px" },
```
eriwarr, line 86: `lgPlus`

eriwarr (approved):
Minor suggestions, but it looks great!!

frontend/tokens/dimensions.tokens.json
benhammondmusic: i checked but looks like SVG uses unitless for these properties

frontend/tokens/dimensions.tokens.json
benhammondmusic: these are already in place, because it would turn the tailwind class into `-sm-plus` in the new CSS vars setup. we can make these changes if we decide in a follow up PR but might not be worth it? or maybe a better naming scheme overall?

PR #4719, opened 2026-05-19
Move insights GCS cache to data server

### LINKED ISSUE #4638: Should migrate insights GCS caching from `frontend_server` to `data_server`
After #4629 the AI insight caching writes to and reads from GCS via the frontend server. While functional, the frontend server's role is to serve the React app and proxy data requests — not to manage persistent storage. A future refactor should move GCS insight caching to the data server, which already has the appropriate GCS IAM permissions and is the established service for data retrieval. This would keep the frontend server stateless and avoid expanding its IAM surface area.

Regarding the LLM API call placement: keep it on the frontend server. The data server is intentionally narrow — it reads pre-processed files from GCS and serves them. Giving it external API dependencies (Anthropic, rate limiting, prompt handling) would bloat its scope and mix concerns. The frontend server is already the right place for user-facing request handling and external API calls. Only the caching layer (GCS read/write) should eventually move to the data server.

eriwarr:
# Description and Motivation

- Moves insight caching (GCS read/write) from the frontend server to the data server via two new endpoints (`GET`/`POST /insight-cache`), keeping the Anthropic API call on the frontend where user-facing request handling already lives.
- Frontend server is now stateless for caching — drops `@google-cloud/storage`, the GCS env var, and the bucket IAM binding; the in-memory cache stays put to save a network hop on warm hits.
- Data server's project-wide IAM stays read-only; write access to the insights bucket is granted via a new bucket-scoped `insights_cache_writer_role`, so the data server can only write to this one bucket.
- Backward-compatible: existing cached entries in `het-insights-cache` use the same key + JSON format and are served unchanged by the new endpoints, so no cache warm-up is needed after deploy.
- Closes #4638

## Has this been tested? How?

- [x] After deploy to `het-infra-test`, confirm Terraform created `insights_cache_writer_role` and bound it to `data-server-runner@…`
- [x] Confirm `INSIGHTS_CACHE_BUCKET` appears on the data server Cloud Run revision and is **absent** from the frontend revision
- [x] On `dev.healthequitytracker.org`, generate a new insight and verify a new file appears under `het-insights-cache/insights/`
- [x] Reload page, regenerate same insight — should serve instantly from GCS cache; data server logs show `GET /insight-cache → 200`, no Anthropic activity
- [x] Generate same insight again in the same session — should hit the frontend in-memory cache (no data server log entries at all)

N/A — Backend architectural refactor.

## Types of changes
- Refactor / chore


benhammondmusic (approved):
Looks awesome. Great job on the check list covers everything I can think of! Excited to winch this feature soon!

PR #4753, opened 2026-06-01
fix(ux): eliminate double history pushes causing stuck back-button navigation

### LINKED ISSUE #2712: Back button still broken in some cases
**Describe the bug**

Navigate through several comparison reports, and then try to navigate back through in order. some get stuck on multiple entries in the history

**Screenshots**
Screenshot 2024-01-05 at 12 30 08 PM

benhammondmusic:
## Root cause

The MadLib URL system uses \`history.pushState\` to create back-navigable history entries when the user changes topic, geography, or mode. Three related bugs existed; all are now fixed in this PR.

---

## Bug 1 - Double pushState in \`handleOptionUpdate\` DEFAULT branch

When the user selected the default/reset option, the code called \`setMadLibWithParam\` (one push) and then \`setParameters\` again (second push). Fixed by removing the redundant call. \`setMadLibWithParam\` also now includes \`MADLIB_PHRASE_PARAM\` in its push so the mode is correctly restored on popstate.

## Bug 2 - Double pushState in \`handleDataTypeUpdate\`

The original fix used \`history.replaceState\` to pre-seed the dt param before \`setMadLibWithParam\`, but \`replaceState\` overwrites the current history entry before the push, so back-navigation landed on the same state twice. Fixed by adding a \`dtOverrides\` parameter to \`setMadLibWithParam\` so the new dt value is included in a single \`pushState\` with no intermediate history manipulation.

## Bug 3 - Stale dt and blank DataTypeSelector after topic switches

**Root cause (MadLibUI.tsx \`handleOptionUpdate\`):** when switching parent topics (HIV to HIV Black Women), the old \`dt1=hiv_deaths\` was carried into the new URL unchanged. \`hiv_deaths\` is not a valid Black Women data type, so the DataTypeSelector received an unknown ID and rendered an empty button. After 4-5 navigations this appeared to users as the back button "not working" because the UI showed blank/stale data type buttons at each back step.

**Root cause (ExploreDataPage.tsx \`readParams\`):** the \`selectedDataTypeConfig\` atoms were only ever written by \`handleDataTypeUpdate\` (user clicks), never by \`readParams\` (the popstate handler). On back navigation the atoms retained the last forward value, so the DataTypeSelector showed a type from a later state instead of what was actually in the URL.

**Fixes:**
- \`handleOptionUpdate\`: detect topic changes (non-Fips \`newValue\`) and pass \`dtOverrides = { dt1: '' }\` (or dt2) to clear the stale dt from the URL. Also reset the atom to null so the selector defaults to the first available type for the new topic.
- \`readParams\`: read \`DATA_TYPE_1_PARAM\` / \`DATA_TYPE_2_PARAM\` from the URL and write the matching \`DataTypeConfig\` (or null) to the atoms. This also improves initial page load from a shared URL - the DataTypeSelector now correctly shows the URL data type immediately on mount.

## Bug 4 - Spurious pushState from stale Report popstate handler (CI regression)

This bug only surfaced in CI after the fixes above because the cross-topic back-navigation test exposed it for the first time.

**Root cause:** \`Report.tsx\` had its own \`psSubscribe('vardisp', ...)\` handler to sync \`selectedDataTypeConfig1Atom\` from the URL on \`popstate\`. The handler's \`readParams\` closure captured \`props.dropdownVarId\` at registration time. When navigating back across a topic boundary (e.g., HIV Black Women → HIV), the handler was still closed over the OLD \`dropdownVarId = 'hiv_black_women'\`. It searched \`METRIC_CONFIG['hiv_black_women']\` for the new URL's \`dt1=hiv_deaths\`, found nothing, and fell back to writing \`hiv_prevalence_black_women_config\` to the shared atom - overwriting what \`ExploreDataPage.readParams\` had just correctly set.

The new 'hiv' Report then rendered with this wrong Black Women config. \`getAllDemographicOptions\` returned \`ONLY_AGE_TYPE_MAP\` (Black Women data is age-only), while the 'hiv' Report's default demographic is \`RACE\`. The mismatch triggered the demographic-reset \`useEffect\`, which called \`setDemographicType\` → \`useParamState\` setter → \`history.pushState(\`&demo=age\`)\`. This inserted a spurious history entry between \`hiv_deaths\` and \`hiv_diagnoses\`, so back press 4 landed on the spurious entry instead of advancing to \`hiv_diagnoses\`.

**Fix:** since \`ExploreDataPage.readParams\` now correctly sets \`selectedDataTypeConfig1Atom\` on every \`popstate\` (added as part of Bug 3), the Report's competing \`psSubscribe\` handler was both redundant and harmful. Removing it leaves the atom update exclusively in \`ExploreDataPage\`, eliminating the stale-closure corruption.

**Underlying structural cause:** the two-tier URL system (MadLib params written via direct \`history.pushState\`, bypassing Jotai) forces atom state and URL state to be synced manually via multiple competing \`psSubscribe\` handlers. Stale closures in those handlers are an inherent risk of this pattern. See the linked issue for a plan to migrate to full Jotai URL management, which would eliminate this class of bug entirely.

---

## Files changed

| File | Change |
|---|---|
| \`ExploreDataPage.tsx\` | Add \`MADLIB_PHRASE_PARAM\` to \`setMadLibWithParam\` push; add \`useSetAtom\` for both config atoms; seed atoms from URL in \`readParams\`; \`?? null\` guards on \`getConfigFromDataTypeId\` |
| \`MadLibUI.tsx\` | Remove redundant \`setParameters\` from DEFAULT branch; \`dtOverrides\` pattern for \`handleDataTypeUpdate\`; clear stale dt and reset atom on topic changes in \`handleOptionUpdate\` |
| \`Report.tsx\` | Remove \`psSubscribe\` handler; \`?? null\` guard on \`setDataTypeConfig\` fallback |
| \`navigation.ci.spec.ts\` | Fix two broken extremes tests; add 4 back-button regression tests including the full 6-state cross-topic flow |

## Test plan

- [x] Change data sub-type (e.g., from "HIV diagnoses" to "HIV deaths") - one back-press returns to the previous sub-type
- [x] Multi-step navigation: 5+ distinct states, verify back steps through each one in order
- [x] Cross-topic back navigation: HIV prevalence, diagnoses, deaths, topic switch to HIV Black Women, BW diagnoses, BW deaths - all 5 back presses each land on the correct distinct URL
- [x] Select the default/reset option from the topic dropdown - back button returns to the previous topic correctly
- [x] Switch between Compare Rates / Explore Relationships / Investigate Rates tabs - back navigates each mode change individually
- [x] \`npx tsc --noEmit\` passes

Closes #2712

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/CLAUDE.md
+++ b/frontend/CLAUDE.md
@@ -41,10 +41,17 @@ Global UI state is managed with Jotai atoms, URL-synced via `jotai-location` (`s
 
 | Tier | Params | Written via | Read via |
 |---|---|---|---|
-| MadLib | `mls`, `dt1`, `dt2`, `group1`, `group2`, `mlp`, `extremes` | `history.replaceState` (bypasses Jotai) | `window.location.search` |
+| MadLib | `mls`, `dt1`, `dt2`, `group1`, `group2`, `mlp`, `extremes` | `setParameters` → `history.pushState` (bypasses Jotai) | `window.location.search` |
 | UI / modal | `demo`, `topic-info`, `multiple-maps`, `chlp-maps`, `vote-dot-org`, `report-insight`, `atl` | `useParamState` → `locationAtom` | `urlParamAtom(key)` — fine-grained, only re-renders on that param's change |
 
 `useParamState` (`src/utils/hooks/useParamState.tsx`) is the hook for Tier 2 params. Its setter reads from `window.location.search` (not Jotai state) so that MadLib params written outside Jotai are preserved in the URL.
+
+**MadLib navigation invariants** — critical rules for the `ExploreDataPage` / `MadLibUI` navigation machinery. Violating these causes duplicate history entries or blank DataTypeSelector buttons.
+
+- `setMadLibWithParam` is the single point of truth for all MadLib URL writes. It calls `history.replaceState` (cleanup only — deletes `dt2` when not in comparevars mode) and then `setParameters` → `history.pushState`. Never call `setParameters` or `history.pushState` separately before or after `setMadLibWithParam` for the same user action — that creates duplicate history entries.
+- Pass `dtOverrides: { dt1: newId }` (or `dt2`) when changing data sub-types so the new value reaches `setParameters` in one shot. Do not write the dt param to the URL via a separate call first.
+- On topic changes (`handleOptionUpdate` with a non-Fips value), pass `dtOverrides: { dt1: '' }` to delete the stale dt from the pushed URL and reset the corresponding `selectedDataTypeConfig` atom. The old data type ID does not apply to the new topic, so carrying it forward causes the DataTypeSelector to render an empty button.
+- `readParams` (the `psSubscribe` popstate handler in `ExploreDataPage`) is responsible for restoring full UI state on browser back/forward. It must sync the `selectedDataTypeConfig1`/`2` jotai atoms from `dt1`/`dt2` in the URL, in addition to calling `setMadLib`. If you add a new atom that should survive back-navigation, restore it in `readParams`.
 
 ## Adding a New Frontend Feature (health topic)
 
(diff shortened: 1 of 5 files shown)


eriwarr (approved):
This is awesome!! Thanks Ben!

PR #4755, opened 2026-06-02
refactor(nav): replace psSubscribe/setParameters with jotai-location

### LINKED ISSUE #4754: refactor(nav): migrate MadLib URL params to full Jotai/jotai-location management
## Background

The app uses a two-tier URL param system (documented in \`frontend/CLAUDE.md\`):

| Tier | Params | Written via | Read via |
|---|---|---|---|
| MadLib | \`mls\`, \`dt1\`, \`dt2\`, \`group1\`, \`group2\`, \`mlp\`, \`extremes\` | \`setParameters\` → \`window.history.pushState\` directly | \`window.location.search\` directly |
| UI / modal | \`demo\`, \`topic-info\`, \`multiple-maps\`, etc. | \`useParamState\` → \`locationAtom\` (jotai-location) | \`urlParamAtom(key)\` |

Tier 1 bypasses Jotai entirely. This forces manual URL-to-atom synchronisation via a custom \`psSubscribe\`/\`window.onpopstate\` system. PR #4753 fixed three distinct back-button bugs that all traced back to that manual sync being fragile - most recently, a stale closure in \`Report.tsx\`'s competing \`psSubscribe\` handler corrupted \`selectedDataTypeConfig1Atom\`, triggered a demographic-reset \`useEffect\`, and inserted a spurious \`history.pushState\` that broke cross-topic back navigation.

The goal of this issue is to collapse both tiers into a single Jotai-managed URL layer, eliminating the class of stale-closure / competing-handler bugs entirely.

---

### Step 1 - Define stable param atoms in \`sharedSettingsState.ts\`

Add a derived atom (or a set of \`urlParamAtom\` calls) for each MadLib param:

\`\`\`ts
export const madLibSelectionsAtom = urlParamAtom(MADLIB_SELECTIONS_PARAM)
export const madLibPhraseAtom     = urlParamAtom(MADLIB_PHRASE_PARAM)
export const group1Atom           = urlParamAtom(MAP1_GROUP_PARAM)
export const group2Atom           = urlParamAtom(MAP2_GROUP_PARAM)
export const dt1Atom              = urlParamAtom(DATA_TYPE_1_PARAM)
export const dt2Atom              = urlParamAtom(DATA_TYPE_2_PARAM)
\`\`\`

\`urlParamAtom\` already exists for Tier 2; this just extends it to Tier 1.

### Step 2 - Replace \`setParameters\` with a single \`setLocationAtom\` write

\`setMadLibWithParam\` currently calls \`history.replaceState\` (cleanup) then \`setParameters\` → \`pushState\`. Replace both with a single \`setLocationAtom\` update that builds the full new search string in one shot. jotai-location's write path calls \`applyLocation\` → \`history.pushState\` once per atom update.

The \`replaceState\` step (deleting \`dt2\` when not in comparevars mode) must be preserved. It can be folded into the same \`setLocationAtom\` write by constructing the full target params before calling the setter - no separate \`replaceState\` needed.

### Step 3 - Remove \`psSubscribe\` / \`window.onpopstate\`

Once all URL writes go through \`locationAtom\`, jotai-location's built-in \`popstate\` listener keeps every \`urlParamAtom\` in sync automatically. The custom \`psSubscribe\`/\`psUnsubscribe\`/\`window.onpopstate\` machinery in \`urlutils.tsx\` can be deleted.

\`ExploreDataPage.readParams\` (which manually reads from \`window.location.search\` on popstate to sync \`madLib\` state and the config atoms) is replaced by reactive \`useAtomValue\` reads on the param atoms defined in Step 1.

### Step 4 - Derive \`madLib\` state from atoms instead of \`useState\`

Currently \`madLib\` is a \`useState\` value in \`ExploreDataPage\` that is manually kept in sync with the URL. With param atoms in place it becomes a \`useMemo\` (or a Jotai derived atom) computed from \`madLibSelectionsAtom\` + \`madLibPhraseAtom\`. Back/forward navigation automatically updates those atoms via jotai-location → \`madLib\` recomputes → UI reflects the correct state without a manual \`readParams\` call.

### Step 5 - Remove the \`Report.tsx\` sync effect

The \`useEffect([dropdownVarId, demographicType, fips])\` in \`Report\` exists primarily to call \`setDataTypeConfig\` from the URL on mount and dropdown changes. Once \`dt1Atom\` is a live jotai-location atom, \`selectedDataTypeConfig1Atom\` can be a derived atom:

\`\`\`ts
export const selectedDataTypeConfig1Atom = atom((get) =>
  getConfigFromDataTypeId(get(dt1Atom) ?? '') ?? null
)
\`\`\`

Removing the manual effect also removes the last foothold for stale-closure bugs in this component.

---

### 1. jotai-location calls \`pushState\` on every write - even no-ops
\`useParamState\`'s setter already guards against this (the \`originalString === currentParams.toString()\` check). Any new setter for Tier 1 params must carry the same guard, otherwise switching to an identical URL (e.g., clicking the already-selected topic) would push a phantom history entry.

### 2. React Router and jotai-location both own \`popstate\`
React Router (\`react-router\`) also listens to \`popstate\` to update its internal \`location\`. Since the app uses \`<RouterProvider>\` / \`useLocation\` in places like \`useDeprecatedParamRedirects\`, React Router's location and jotai-location's \`locationAtom\` could diverge if writes go only through one of them. Need to audit every \`useLocation()\` call and decide whether it should move to \`urlParamAtom\` or stay with React Router.

### 3. \`history.replaceState\` has no Jotai equivalent out of the box
The current code uses \`replaceState\` to clean up stale params before a \`pushState\`. jotai-location's \`applyLocation\` always calls \`pushState\`. A \`replaceState\` equivalent would require either a custom \`applyLocation\` override or constructing the full cleaned URL in a single atom write (Step 2 above). Verify that jotai-location's API exposes enough hooks for this before committing to the approach.

### 4. \`setMadLibWithParam\` is called with \`dtOverrides\` to avoid a double-push
The \`dtOverrides\` pattern was introduced in PR #4753 specifically because two separate URL writes in the same handler created two history entries. With a single Jotai setter this problem goes away, but the \`dtOverrides\` parameter can be removed as part of the migration - don't carry it forward as dead abstraction.

### 5. Batching across atoms
Jotai batches synchronous atom updates within a React render cycle, but \`locationAtom\` writes are DOM side-effects (they call \`pushState\` imperatively). If multiple params need to change atomically (e.g., \`mls\` + \`dt1\` in one navigation step), they must be written in a single \`setLocationAtom\` call with the full new search string, not as separate per-param writes.

### 6. \`useDeprecatedParamRedirects\` reads from React Router's \`useLocation\`
This hook currently reads \`mls\` via React Router's stale location (which doesn't update on direct \`pushState\` calls. After migration, if all writes go through jotai-location, React Router's location will stay in sync via `popstate`, but the hook may need to be rewritten to read from `urlParamAtom(MADLIB_SELECTIONS_PARAM)` directly to avoid any remaining staleness window.

---

## Unknowns

- **jotai-location version and `replaceState` support**: the current `jotai-location` version in the project may not expose a `replace` option on `applyLocation`. Check the installed version's API before designing the write path.
- **Performance of derived `madLib` atom**: `parseMls` runs on every `mls` param change. If the derived atom runs `parseMls` on every render that reads `madLibSelectionsAtom`, memoisation needs to be confirmed (Jotai derived atoms are already memoised on input equality, but `mls` is a string so equality is value-based - should be fine).
- **`useParamState` interaction with MadLib writes**: currently `useParamState`'s setter reads from `window.location.search` (not `locationAtom`) to preserve MadLib params that were written outside Jotai. After migration that workaround can be removed - but need to verify no other code depends on the current read-from-window behaviour.
- **Concurrent navigation actions**: the app doesn't have optimistic UI, but rapid user actions (clicking topic then immediately clicking dt) could interleave atom writes. Confirm that Jotai's batching + jotai-location's single `pushState` per commit produces the expected history entry count.

---

## Acceptance criteria

- [ ] All existing `navigation.ci.spec.ts` tests pass, including the 6-state cross-topic back-navigation test added in #4753
- [ ] `psSubscribe`, `psUnsubscribe`, `psSubscriptions`, and `window.onpopstate` are deleted from `urlutils.tsx`
- [ ] `setParameters` and the direct `window.history.pushState` call are deleted
- [ ] `ExploreDataPage.readParams` and its `useEffect`-based popstate subscription are removed
- [ ] `Report.tsx`'s `useEffect([dropdownVarId, demographicType, fips])` no longer manually syncs any atom from the URL
- [ ] `npx tsc --noEmit` passes
- [ ] The two-tier table in `frontend/CLAUDE.md` is updated to reflect a single tier
ISSEOF
)

benhammondmusic:
## Summary

- Eliminates the custom psSubscribe/window.onpopstate machinery and direct history.pushState calls entirely. All URL writes now go through setLocationAtom (jotai-location) — one pushState per navigation action.
- selectedDataTypeConfig1/2Atom are now read-only derived atoms from urlParamAtom(dt1/dt2). Always correct after back/forward with no manual sync.
- madLib in ExploreDataPage is a useMemo derived from urlParamAtom(mls + mlp) — not owned state. Back/forward automatically recomputes via jotai-location's built-in popstate listener.
- Removes the stale-closure risk in CompareReport (same class of bug fixed in #4753 for Report.tsx).

## Key file changes

| File | What changed |
|---|---|
| sharedSettingsState.ts | selectedDataTypeConfig1/2Atom plain → derived from URL atoms |
| ExploreDataPage.tsx | madLib useState + readParams/psSubscribe → useMemo from atoms |
| MadLibUI.tsx | Removes direct atom writes; handleDataTypeUpdate no longer calls setConfig |
| CompareReport.tsx | Removes psSubscribe handler; comparegeos dt2=dt1 handled inline |
| Report.tsx | Removes manual dt sync from useEffect |
| MapCard.tsx | setParameter → setLocationAtom functional update |
| urlutils.tsx | Deletes setParameter, setParameters, psSubscribe, psUnsubscribe, psSubscriptions, window.onpopstate |
| useParamState.tsx | Setter reads from prev.searchParams; no more window.location.search workaround |
| useDeprecatedParamRedirects.tsx | navigate() → setLocation in useLayoutEffect |

## Test plan

- [ ] Run npm run e2e navigation — all back-button regression tests from #4753 must pass
- [ ] Manual: HIV topic — change sub-type, press back, confirm returns to previous sub-type
- [ ] Manual: change topic (HIV to Incarceration) — DataTypeSelector shows correct first option, not empty
- [ ] Manual: compare geos mode — both panels show same data type
- [ ] Manual: compare topics mode — each panel shows its own data type; back navigation works
- [ ] Manual: demographic change — DemographicSelector updates; no extra history entry created
- [ ] Manual: map group filter — selecting a group updates URL; back restores previous group
- [ ] Manual: back/forward through 5+ navigation steps — each step lands on correct distinct state
- [ ] npx tsc --noEmit passes

Closes #4754

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/CLAUDE.md
+++ b/frontend/CLAUDE.md
@@ -37,21 +37,25 @@ URL params (mls, dt1, demo, etc.)
 
 Global UI state is managed with Jotai atoms, URL-synced via `jotai-location` (`src/utils/sharedSettingsState.ts`).
 
-**Two-tier URL param system** — URL params split into two write paths:
+**Unified URL param system** — all params written through a single path:
 
-| Tier | Params | Written via | Read via |
-|---|---|---|---|
-| MadLib | `mls`, `dt1`, `dt2`, `group1`, `group2`, `mlp`, `extremes` | `setParameters` → `history.pushState` (bypasses Jotai) | `window.location.search` |
-| UI / modal | `demo`, `topic-info`, `multiple-maps`, `chlp-maps`, `vote-dot-org`, `report-insight`, `atl` | `useParamState` → `locationAtom` | `urlParamAtom(key)` — fine-grained, only re-renders on that param's change |
+| Params | Written via | Read via |
+|---|---|---|
+| `mls`, `dt1`, `dt2`, `group1`, `group2`, `mlp` | `setLocationAtom({ searchParams })` → jotai-location → `history.pushState` | `urlParamAtom(key)` |
+| `demo`, `topic-info`, `multiple-maps`, `chlp-maps`, `vote-dot-org`, `report-insight`, `atl`, `extremes` | `useParamState` → `setLocationAtom` | `urlParamAtom(key)` |
 
-`useParamState` (`src/utils/hooks/useParamState.tsx`) is the hook for Tier 2 params. Its setter reads from `window.location.search` (not Jotai state) so that MadLib params written outside Jotai are preserved in the URL.
+`jotai-location` owns `locationAtom` and handles `popstate` automatically — back/forward navigation keeps all atoms in sync with no manual handlers.
 
-**MadLib navigation invariants** — critical rules for the `ExploreDataPage` / `MadLibUI` navigation machinery. Violating these causes duplicate history entries or blank DataTypeSelector buttons.
+`useParamState` (`src/utils/hooks/useParamState.tsx`) is the hook for UI / modal params.
 
-- `setMadLibWithParam` is the single point of truth for all MadLib URL writes. It calls `history.replaceState` (cleanup only — deletes `dt2` when not in comparevars mode) and then `setParameters` → `history.pushState`. Never call `setParameters` or `history.pushState` separately before or after `setMadLibWithParam` for the same user action — that creates duplicate history entries.
-- Pass `dtOverrides: { dt1: newId }` (or `dt2`) when changing data sub-types so the new value reaches `setParameters` in one shot. Do not write the dt param to the URL via a separate call first.
-- On topic changes (`handleOptionUpdate` with a non-Fips value), pass `dtOverrides: { dt1: '' }` to delete the stale dt from the pushed URL and reset the corresponding `selectedDataTypeConfig` atom. The old data type ID does not apply to the new topic, so carrying it forward causes the DataTypeSelector to render an empty button.
-- `readParams` (the `psSubscribe` popstate handler in `ExploreDataPage`) is responsible for restoring full UI state on browser back/forward. It must sync the `selectedDataTypeConfig1`/`2` jotai atoms from `dt1`/`dt2` in the URL, in addition to calling `setMadLib`. If you add a new atom that should survive back-navigation, restore it in `readParams`.
+**MadLib navigation invariants** — critical rules for the `ExploreDataPage` / `MadLibUI` navigation machinery.
+
+- `setMadLibWithParam` is the single point of truth for all MadLib URL writes. It builds the complete new `URLSearchParams` and calls `setLocationAtom` once (one `pushState`). Never write to the URL separately before or after — that creates duplicate history entries.
+- Pass `dtOverrides: { dt1: newId }` (or `dt2`) when changing data sub-types so the new value is included in the same write.
+- On topic changes (`handleOptionUpdate` with a non-Fips value), pass `dtOverrides: { dt1: '' }` to exclude the stale dt from the URL. The old data type ID does not apply to the new topic.
+- `selectedDataTypeConfig1Atom` and `selectedDataTypeConfig2Atom` are **read-only derived atoms** — they derive from `urlParamAtom('dt1')` / `urlParamAtom('dt2')`. Never call their setters directly. Update dt values by writing the URL param via `setMadLibWithParam` with `dtOverrides`.
+- `madLib` in `ExploreDataPage` is a `useMemo` derived from `urlParamAtom('mls')` + `urlParamAtom('mlp')`. It is not owned state — never call `setMadLib`. Back/forward automatically updates the URL atoms which recomputes `madLib`.
+- If you add a new atom that should survive back-navigation, derive it from a `urlParamAtom` rather than wiring up a manual `popstate` handler.
 
 ## Adding a New Frontend Feature (health topic)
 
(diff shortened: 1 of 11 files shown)


benhammondmusic:
Closing and reopening from the correct fork head (benhammondmusic:refactor/jotai-url-params). The branch was originally pushed directly to origin before the team's fork-based PR workflow was documented. Continued in #4762.

PR #4850, opened 2026-06-17
fix(insight): fix county-level map insights and suppress no-compariso…

### LINKED ISSUE #4681: New (unmerged) county-level Cervical Cancer view causing AI to stumble
this happened while creating #4680 locally: /exploredata?mls=1.cancer_incidence-3.13121&group1=All&dt1=cervical_cancer_incidence

Claude acts like it's talking to the dev and not to the user sice it had trouble analyzing the data. First issue is why didn't it get the right data view? Second issue is we need to instruct the LLM to always produce user-facing responses, not dev-facing even when it fails or misunderstands. 

the LLM payload; note that the races aren't labeled here.

```

{"prompt":"This is a choropleth map showing Cervical cancer cases in Fulton County, Georgia across all race/ethnicity groups. The intended message is to highlight geographic health equity disparities.\n\nData:\n- Fulton County: 6.6 cases per 100k\n- Fulton County: 8.7 cases per 100k\n- Fulton County: 10.3 cases per 100k\n- Fulton County: 4 cases per 100k\n\nWrite a single sentence at an 8th grade reading level that names the specific states or regions with the highest rates, contrasts them with those with the lowest, and captures why this geographic gap matters — focus on the \"so what\", not the chart mechanics.","cacheKey":"/exploredata?mls=1.cancer_incidence-3.13121&group1=All&dt1=cervical_cancer_incidence#rate-map"}

```

```

2676: 
{
  county_fips: "13121", 
  cervical_per_100k: 6.6, 
  cervical_estimated_total: 37, 
  race_and_ethnicity: "All",
  county_name: "Fulton County",
  state_fips: "13",
  state_name: "Georgia"
}

2677: 
{
  county_fips: "13121", 
  cervical_per_100k_is_suppressed: true,
  race_and_ethnicity: "Asian, Native Hawaiian, and Pacific Islander (NH)",
  county_name: "Fulton County",
  state_fips: "13",
  state_name: "Georgia"
}

2678: 
{
  county_fips: "13121", 
  cervical_per_100k: 8.7, 
  cervical_estimated_total: 23,
  race_and_ethnicity: "Black or African American (NH)",
  county_name: "Fulton County",
  state_fips: "13",
  state_name: "Georgia"
}

2679: 
{
  county_fips: "13121", 
  cervical_per_100k: 10.3, 
  cervical_estimated_total: 4,
  race_and_ethnicity: "Hispanic or Latino",
  county_name: "Fulton County",
  state_fips: "13",
  state_name: "Georgia"
}

2680: 
{
  county_fips: "13121", 
  cervical_per_100k_is_suppressed: true, 
  race_and_ethnicity: "Indigenous (NH)",
  county_name: "Fulton County",
  state_fips: "13",
  state_name: "Georgia"
}

```

Image

eriwarr:
# Description and Motivation

Closes #4681

The AI insight for a **county-level Cervical Cancer rate map** was producing developer-facing jargon instead of a user-friendly sentence. Two root causes:

- **The model received unlabeled data.** A county map returns one geographic unit, so every row carried the same `fips_name` — there was no geographic comparison to draw, and the model had no demographic labels to work with.
- **The model lapsed into technical/developer language** ("I don't have enough information…", "the data only shows…") when it hit a degenerate, single-value view.

Changes:

- **Label every map row by both place and demographic group.** `formatDataRows` now writes each map row as `Place (Group): value` and keeps the aggregate `All` row as a baseline anchor. This lets the model surface either a *geographic* gap (across places) or a *within-place* gap (across groups in one county, using `All` as the baseline) — so a single-region county map still yields a meaningful disparity instead of a meaningless geographic comparison. The map prompt is geography-agnostic, since the within-place case also arises for sparse USA/state maps.
- **Steer the prompt with the user's on-screen focus.** Maps pass the currently highlighted `activeDemographicGroup` so the insight leads with the group the user selected; trend cards pass the user's `selectedGroups` (from the legend) so the time-series insight describes only the lines on screen. Both thread through `CardWrapper` and are folded into the insight cache key so a re-focused chart regenerates.
- **Suppress the insight UI when there's nothing to compare.** `hasEnoughDataForInsight` / `prepareInsightData` gate the insight button and panel in `CardWrapper`: a view with fewer than two comparable values hides the insight entirely rather than restating a single number. Entry counting reuses the exact text the model receives, so the visibility gate and generated text can never disagree.
- **Tighten tone via a system prompt.** `frontend_server` now sends an `INSIGHT_SYSTEM_PROMPT` that forbids mentioning the data/labels/limitations and developer-speak, instructs a single-fact fallback, and bans preamble.

## Has this been tested? How?

- Added `generateVisualizationInsight.test.ts` — 11 unit tests covering `formatDataRows` (map rows labeled by both place + group, `All` kept, multi-region by place, non-map demographic-only rows, time-series `selectedGroups` filtering), `prepareInsightData` (entry counts for single-region, multi-group, and group-filtered time-series), and `hasEnoughDataForInsight` (false for one group / no response, true for two+). All pass.
- Verified locally: `vitest run` (11/11), `tsc --noEmit` (clean), `cspell` (0 issues), `biome` (clean) on the changed files.
- Manually confirmed on the deploy preview: a multi-group county (Gwinnett) surfaces a race-labeled between-group insight; a county whose choropleth shows only the overall value (Bartow) still surfaces a within-place gap thanks to the both-columns labeling.

## Types of Changes

(leave all that apply)

- Bug fix

### Gwinnett County — between-group gap

A county with two or more comparable demographic groups surfaces a user-friendly, race-labeled insight (and the sparkle insight button at the card's top-right).

**Mobile (375px)**
Gwinnett cervical insight mobile

**Desktop (1280px)**
Gwinnett cervical insight desktop

---

### Bartow County — within-place gap

Even where the choropleth colors only the county-overall value, the both-columns labeling lets the model surface a within-place disparity (a demographic group vs. the `All` baseline) rather than a meaningless single-region geographic comparison.

**Mobile (375px)**
Bartow cervical within-place mobile

**Desktop (1280px)**
Bartow cervical within-place desktop


benhammondmusic (commented):
~~Superseded by follow-up review — see below.~~

benhammondmusic (changes requested):
## The core problem this PR does not fully fix

Every map data row is built in `formatDataRows` (`generateVisualizationInsight.ts`) using only `fips_name` as the label key:

```typescript
const groupKey = isMap ? 'fips_name' : demographicType
return `- ${row[groupKey]}: ${val}`
```

The demographic column -- `race_and_ethnicity`, `age`, `sex`, whatever the user has selected -- is **never included**. So the model receives:

```
- Bartow County: 13 per 100k
- Bartow County: 8.7 per 100k
- Bartow County: 6.6 per 100k
```

Three rows, same label, different numbers, no indication what distinguishes them. The model produces jargon because there is genuinely nothing to compare. This affects every map card, not just county-level ones.

---

## What this PR does instead

It detects single-county views (`isSingleRegionMap`) and swaps the label key from `fips_name` to `demographicType`. This fixes the county case by accident -- the county name disappears but the demographic label appears. It does not touch the multi-region map path, which has the identical problem.

It also drops the `All` row in the single-region path. `All` is not an artifact -- it is the overall rate for all people in that place, and comparing a specific demographic group against it is exactly the disparity HET exists to surface. Dropping it removes a useful anchor.

---

## The right fix for maps

**Every map row should include both columns:**

```typescript
// maps: label with both location and demographic group
return `- ${row.fips_name} (${row[demographicType]}): ${val}`
```

That is a one-line change to the existing map path. No `isSingleRegionMap` branch needed. The model then receives:

```
- Bartow County (All): 13 per 100k
- Bartow County (Black (NH)): 8.7 per 100k
- Bartow County (White (NH)): 6.6 per 100k
- Fulton County (All): 9.2 per 100k
- Fulton County (Black (NH)): 11.1 per 100k
```

This works for every map type uniformly -- multi-region, single-region, race, age, sex.

**Also pass `activeDemographicGroup` through to the insight.**

`activeDemographicGroup` is `useState` local to `MapCard` and never reaches `CardWrapper` or `prepareInsightData`. The model has no idea which group the user is looking at. Passing it through (new optional prop on `CardWrapper`) lets the prompt tell the model which group is highlighted, so it can lead with that group's story and use `All` as a comparison baseline when available.

---

## The same problem exists in the trend cards -- and should be fixed in this PR

`RateTrendsChartCard` and `ShareTrendsChartCard` both maintain a `selectedTableGroups: DemographicGroup[]` state that is synced directly from the D3 legend. When a user clicks the legend down to two lines, `selectedTableGroups` captures exactly those two groups. This state is already passed to `AltTableView` for the accessible table -- but it is **never passed to `CardWrapper`**, so the insight system never sees it.

Result: a user who has focused the trend chart on "Black (NH)" and "White (NH)" gets an insight that describes all 8+ demographic groups including the ones they hid.

The fix follows the same pattern as the map fix: add an optional `selectedGroups?: DemographicGroup[]` prop to `CardWrapper`, thread it to `InsightVisualizationCard` and `generateCardInsight`, and in `formatDataRows` for the time series path, filter to only the selected groups when provided.

This is a higher-priority fix than the map issue because the data is already being captured correctly in the card -- it just never makes it to the model. The wiring is one prop away.

---

## The single-region + All-only edge case

When a user views a single county with `All` selected (the default), every row has the same `fips_name` and the same demographic value (`All`). There is genuinely one data point -- nothing to compare. The right behavior is to hide the insight entirely, which `hasEnoughDataForInsight` already does correctly once the row format is fixed. No special handling needed beyond that gate.

If a richer experience is wanted here in future (e.g. showing the county rate alongside state and national rates for context), that requires an additional query for the parent geography and is a separate feature tracked in #4852.

---

## What to keep from this PR

- The `hasEnoughDataForInsight` gate and `prepareInsightData` separation -- correct and useful.
- The system prompt addition in `server.js` -- good guardrails against developer-speak.
- The new unit tests -- keep them, update the expected row format to include both columns.

*Review by Claude Code (/review skill).*

benhammondmusic (approved):
Look great!!

PR #4801, opened 2026-06-10
feat(insights): user flagging for problematic AI insights (#4614)

eriwarr:
## Summary

Implements user flagging for problematic AI-generated insights (closes #4614). Lets users flag a bad insight with a reason, records it to a curated GCS archive, regenerates a fresh insight in its place, and feeds flagged content back into the prompt as negative examples.

### What's included
- **Flag button** on both the per-card insight (`InsightVisualizationCard`) and the report-level panel (`InsightReportCard`) — reason picker (inaccurate / misleading / offensive / other) + optional note (capped at 1,000 chars with a live counter).
- **Flagging does not hide the combo.** A user flag records the bad output (status `flagged`) and drops the cached text so a **fresh insight regenerates in place**. Only the team can escalate a record to `suppressed`/`permanent` (hidden) via `PATCH /flagged-insights`, or clear it with `reenabled`.
- **Knowledge base / feedback loop.** Flagged `{reason, content}` pairs are stored in a dedicated `het-flagged-insights` bucket (no TTL — kept separate from the cache bucket, whose 210-day delete rule would otherwise sweep them) and fed back into the generation prompt as negative examples. Enabled via `INSIGHT_NEGATIVE_EXAMPLES_ENABLED=true` on the frontend server.
- **Minimal admin review** endpoints: `GET /flagged-insights` (list) and `PATCH /flagged-insights` (re-enable / permanently suppress).

### Infra changes (Terraform)
- New `het-flagged-insights` GCS bucket (no lifecycle rule; **uniform bucket-level access** enabled so this sensitive content is governed purely by IAM, no legacy ACLs).
- New `flagged_insights_writer` custom IAM role bound to the data-server SA; `storage.objects.delete` added to the insights-cache role.
- New `FLAGGED_INSIGHTS_BUCKET` env on the data server and `INSIGHT_NEGATIVE_EXAMPLES_ENABLED` on the frontend server.

Merging to `main` triggers `terraform apply`, which provisions the bucket + role on the test project.

## Screenshots

**Flag a problematic insight** — reason picker (inaccurate / misleading / offensive / other) + optional note (1,000-char cap with a live counter):

het_shot2_flag_popover

**The AI insight in context, with the flag affordance:**

het_shot1_insight_card

## Security & robustness hardening (from review)

[...]

PR #4851, opened 2026-06-17
Feat/flagged insight review tooling

eriwarr:
# Description and Motivation

Follow-up to #4801 (AI insights). Hardens the flagged-insight feedback loop and gives the team tooling to triage user reports, based on discussion with Ben.

- Stop letting raw user flags poison the generation prompt. Previously any anonymous flagged record was fed back into the insight prompt as a negative example, so a single bad-faith or mistaken public report could degrade output. Now only team-confirmed suppressed records are used as negative examples; a raw flagged record stores the report but no longer influences generation.
- Simplify the flag lifecycle to two states: flagged and suppressed. A user flag records the bad output and clears the cached insight so a fresh one regenerates in its place (it never hides the chart). The team can then confirm a report as suppressed, which keeps the combo live but steers future output: the cached insight is dropped and the suppressed content is fed back into the prompt as a negative example, so the next visitor sees a regenerated insight that avoids the bad one — not a dead card. A false-alarm report is deleted outright rather than given a status, so there's nothing to "re-enable."
- Add a team review tool (scripts/review_flagged_insights.sh) that reads/writes the flagged-insights GCS bucket directly (gated by GCP IAM): list all reports grouped by status, interactively triage each unhandled one (suppress / delete / skip), or run a non-interactive --ci check.
- Add a weekly cron (cronReviewFlaggedInsights.yml) that runs the script in --ci mode and fails when untriaged reports exist, plus a status badge in the README so a red mark signals the team has reports awaiting review.
- Clarify user-facing wording — replace internal jargon ("flagged", "synthesizing") with plainer language ("reported", "Reviewing all charts with AI…").

## Has this been tested? How?

- Backend unit tests (python/tests/data_server/test_main.py) updated for the two-state model: assert the negative-example filter feeds only suppressed (excludes raw flagged), that a suppressed combo is not hidden on read, that confirming suppressed clears the cache, and that legacy statuses (permanent / reenabled) are rejected as invalid. All 12 pass.
- Review script exercised end-to-end against the live het-flagged-insights bucket in het-infra-test-05: seeded a flag record, confirmed --list, --ci (exit 1 on unhandled), and --review (suppress / delete / skip) all behave correctly, including on real URL-shaped keys. Also verified per-function behavior with a stubbed gcloud (failed-upload not counted, benign cache-miss vs. real cache-clear failure).
- Full suite green: 12 backend tests, 329 frontend tests, plus bash -n, node --check, CSpell, and tsc --noEmit all clean.

## Types of Changes
- New content or feature
- Refactor/chore 

## Screenshots
- --list: every flag record grouped by status, newest details on top and the (URL-shaped) cache key on its own line, with a totals footer summarizing counts per status. Screenshot 2026-06-24 at 3 16 57 PM
- --review: interactively triages each unhandled report, showing its topic, reason, key, and captured insight text, then prompts for an action: suppress, delete (false alarm), or skip. Suppressing clears the cache so the combo regenerates steered away from the bad output. Screenshot 2026-06-24 at 3 19 20 PM
- --ci mode — non-interactive check used by the weekly cron. Prints each untriaged report and exits non-zero when any exist, turning the workflow (and README badge) red as a signal that reports await review. Screenshot 2026-06-24 at 12 32 20 PM


benhammondmusic:
A couple of suggestions after thinking through the flow:

**Drop `permanent` and `reenabled`.** `suppressed` plus a delete option seems sufficient:
- The review script only processes `flagged` records, so a suppressed insight can't be accidentally re-enabled through normal usage
- Since flagging already clears the cache and triggers an immediate regen, if the team decides a flag was a false alarm, deleting the GCS record seems like enough. Nothing left to re-enable.

**Reconsider what suppression means for the end user.** It might be a better flow to always show users an insight rather than a dead card. Possible approach:
- When team marks `suppressed`, also clear the GCS cache entry (same as re-enable does today), so the next visitor triggers a fresh generation with the suppressed content feeding in as a negative example
- This way suppression steers the model rather than hiding the combo entirely

Net result would be two statuses (`flagged`, `suppressed`), three script actions (suppress, delete, skip), and no dead-card state needed in the frontend.

benhammondmusic (approved):
Sweet!! 🚀

benhammondmusic (approved):
Looks good — two-status simplification and keeping insights visible while using suppressed content as counter-examples is exactly right.

PR #4853, opened 2026-06-18
fix: show user feedback when clipboard image copy fails

### LINKED ISSUE #4810: Card image copy to clipboard fails silently when document loses focus
## Bug

When a user clicks \"Copy card as image\" and switches away from the tab before the capture completes, the clipboard write fails with:

\`\`\`
NotAllowedError: Failed to execute 'write' on 'Clipboard': Document is not focused.
\`\`\`

The error is logged to the console but the user receives no feedback — the feature just silently does nothing.

## Root cause

\`captureAndSaveImage\` in \`src/utils/cardImageExportUtils.ts\` calls \`navigator.clipboard.write()\` after an async dom-to-image capture. Browsers require the document to be focused at the time of the clipboard write; if the user has switched tabs during the (sometimes slow) capture, the write is rejected.

## Suggested fix

Catch \`NotAllowedError\` in \`handleDestination\` and either:
- Show a user-facing error message (toast / snackbar) explaining the copy failed and asking the user to try again while keeping the tab in focus, or
- Fall back to downloading the image as a file instead of copying to clipboard

## Affected file

\`frontend/src/utils/cardImageExportUtils.ts\` — \`handleDestination\` / \`captureAndSaveImage\`

bhagy-patel1:
## Summary

Fixes #4810 by providing user feedback when copying a card image to the clipboard fails because the document loses focus during image capture.

## Changes

* Return `undefined` when clipboard writes fail in `handleDestination`
* Added error state handling in `useCardImage`
* Added an error snackbar to `CopyCardImageToClipboardButton`
* Preserved existing success behavior when clipboard copy succeeds

### Success case

* Clicked "Copy Image To Clipboard" while keeping the tab focused
* Image copied successfully
* Success snackbar displayed

### Failure case

* Clicked "Copy Image To Clipboard" and immediately switched tabs
* Clipboard write failed with document focus error
* Error snackbar displayed informing the user to keep the tab focused and try again


bhagy-patel1:
I've opened a PR to address this issue. The fix adds user-facing feedback when clipboard writes fail due to the document losing focus during image capture, while preserving the existing success flow❤️.

benhammondmusic (approved):
Fantastic; thank you so much!

PR #4979, opened 2026-07-18
Polish contrast insight: sparkle on each compare card opens shared section insight

### LINKED ISSUE #4976: Compare these views: add sparkle AI indicator and raised card styling
## Problem

The "Compare these views" button in `ContrastInsightSection.tsx` currently:

- Uses a `CompareArrows` icon, which gives no signal that this is an AI-powered feature.
- Appears visually unmoored above the card -- no background, no elevation, looks like floating text.

## Fix

1. **Icon:** Swap `CompareArrows` for `AutoAwesome` (sparkle) from MUI icons to signal AI.
2. **Label/tooltip:** Add "AI" to the button label or tooltip text (e.g. "Compare with AI" or keep "Compare these views" with an explicit AI tooltip).
3. **Visual treatment:** Wrap the trigger button in a raised chip or small card so it has clear affordance and doesn't appear to float.

## Files

- `frontend/src/reports/ContrastInsightSection.tsx` (lines 1, 120, 130, 134, 146)

Closes part of milestone: Intersectional card polish

benhammondmusic:
**Preview:** [Compare vars mode - HIV Black Women vs Women in Gov](https://deploy-preview-4979--health-equity-tracker.netlify.app/exploredata?mls=1.hiv_black_women-3.women_in_gov-5.00&mlp=comparevars&demo=race_and_ethnicity)

## Summary

Builds on the existing `ContrastInsightSection` component (already in `main`) with targeted changes to align it with the rest of the insights feature:

- **Icon:** the original used `CompareArrows`; changed to `AutoAwesome` (sparkle) to match `InsightVisualizationButton`, which uses sparkle for all other AI insight triggers - one icon for all AI affordances in the app
- **"contrast" to "comparison":** "contrast" reads as a technical/statistical term; "comparison" is plain language that matches what users already see ("compare geographies", "compare variables") throughout the MadLib and navigation - the concept is the same, the word is more accessible
- **Trigger placement:** the original rendered a standalone trigger button inside `ContrastInsightSection`; moved triggering to the sparkle already in each card's header, so users do not need to find a separate button below the pair - either card's sparkle opens the shared insight
- **Shared open state:** replaced local `isOpen` state with `contrastInsightOpenAtom` (keyed by `hashId`) so both cards in a pair genuinely share one open/closed state
- **Section labels with correct plurality:** insight header reads "Rate maps comparison", "Data tables comparison", "Rate chart comparison", etc. - uses `pluralOnCompare` from `reportProviderSteps` (already defined, previously unused) and skips appending "s" when the label already ends in one ("Age-adjusted ratios")
- **Flag link + scroll fix:** added `FlagInsightButton` (matching per-card insights) and a scroll-into-view effect so the card clears the sticky header when it opens
- **Accessibility:** added `role='status'` to announce insight content to screen readers, and restore focus to the triggering sparkle button when the user closes the insight

## Impact

- Users see the same sparkle icon for every AI insight trigger - no second visual vocabulary to learn
- "Comparison" over "contrast" matches the language already in the UI (compare mode, compare variables), lowering the cognitive load
- Section header pluralizes correctly ("Rate maps comparison" not "Rate map comparison") - matches the actual content, two side-by-side maps
- Screen reader users are notified when the AI insight appears and focus is restored to where they triggered it, satisfying WCAG 2.1 AA focus management criteria

## Test plan

- [x] Sparkle appears in upper-right of both cards in compare mode (Playwright)
- [x] Clicking either card's sparkle opens the shared insight card above the pair (Playwright)
- [x] Close button in insight card header closes it and restores both sparkles to open state (Playwright)
- [x] Per-card sparkle still works correctly in single-disparity (non-compare) mode (Playwright)
- [x] Insight card header pluralizes correctly for map/table sections (e.g. "Rate maps comparison") and stays singular where appropriate (e.g. "Rate chart comparison")
- [x] Insight card is raised and visually matches other card surfaces

Closes #4976

image

image

image


eriwarr:
Nice cleanup — moving the trigger into each card's sparkle and sharing open-state via an atom reads well. Two functional bugs in the newly-added behavior, plus a minor screenshot nit:

🔴 The new flag button is permanently disabled
ContrastInsightSection captures result.cacheKey and passes it to FlagInsightButton, but generateContrastInsight never returns a cacheKey. Unlike generateCardInsight (which does return { ...result, cacheKey }), it returns the raw fetchAIInsight result — and fetchAIInsight doesn't populate cacheKey. So serverCacheKey stays null, FlagInsightButton gets cacheKey={undefined}, and its disabled={!props.cacheKey} leaves the "Report harmful or inaccurate content" link greyed out and unusable.

Screenshot 2026-07-20 at 6 54 53 PM

Fix in generateContrastInsight.ts:

const result = await fetchAIInsight(prompt, { cacheKey })
return { ...result, cacheKey }
(Passing topic: dataTypeConfig1.dataTypeId into fetchAIInsight there too would match the per-card path.)

🔴 Dead sparkle on the Unknowns map card in compare mode
showContrastInsightButton = inCompareMode && cardHasData && insightProps != null fires for every card that has insight props. But a sparkle only opens something where a matching <ContrastInsightSection hashId=...> is rendered. CompareReport has sections for rate-map, rates-over-time, rate-chart, inequities-over-time, population-vs-distribution, data-table, and age-adjusted-ratios — but not unknown-demographic-map. So the Unknowns demographic map card shows a sparkle that, when clicked, just flips to the trash icon and opens nothing.

Screenshot 2026-07-20 at 6 53 57 PM

Fix: either add a ContrastInsightSection for unknown-demographic-map, or gate showContrastInsightButton to the hashIds that actually have a section.

🟡 Sparkle shows in card screenshots in compare mode
The per-card button carries className='hide-on-screenshot remove-height-on-screenshot'; the new compare-mode IconButton branch omits it, so it'll appear in exported card images. Add the same class for parity.

HET RATE-MAP Explore relationships between COVID-19 and Poverty in United States  Jul 2026

benhammondmusic:
@eriwarr Fixed in e99b2a75d — all three items addressed:

**Flag button always disabled:** `generateContrastInsight` now returns `{ ...result, cacheKey }` and passes `topic` to `fetchAIInsight`, matching the per-card path in `generateVisualizationInsight`.

**Dead sparkle on unknowns map:** Exported `CONTRAST_SECTION_HASH_IDS` from `ContrastInsightSection` and added it as a guard in `CardWrapper.showContrastInsightButton` — only hashIds that actually have a section show the sparkle.

**Screenshot class:** Already addressed in an earlier commit on this branch.

eriwarr (approved):
Looks great!!

Issue #2690, opened 2023-12-29
Need to encode group selection in URL


benhammondmusic:
## Is your feature request related to a problem? Please describe.

Right now if someone selects certain groups on the time charts, that state is held in memory and not accessible via shared link like our other settings

## Describe the solution you'd like

Encode highlighted groups (not just single group from map card) in url parameters

## Which of our [user persona(s)](https://docs.google.com/document/d/1EASpK_THTE_uy_Yk0sut2GTVd3w5pKtKH7LpLtF-0co/) would this feature primarily benefit?

- Data Diver: contributes to the body of knowledge surrounding health equity (e.g. analysts, researchers, professors in academic or corporate settings)
- Data Interpreter: connects others by translating complex data into something actionable (e.g. think tank or legislative settings)
- Informer: helps everyday people understand important current events (e.g. journalists working in storytelling settings)
- Persuader: passage of legislation by presenting concise and compelling arguments (e.g. staff members, policy advisors in legislative or nonprofit settings)

benhammondmusic:
## Design decisions

Scoping this issue after a design pass. Two pieces will ship together: the URL encoding this issue asks for, plus retiring a related workaround that would otherwise spread into it.

### Params

Four: `rateGroups1`, `rateGroups2`, `shareGroups1`, `shareGroups2` — split by card type and by compare column, matching the existing `group1`/`group2` suffix convention.

Separate rather than shared because the available group sets genuinely differ, so a shared param would carry groups the other card cannot render and the selection would silently evaporate:

- `ShareTrendsChartCard.tsx:70` excludes `ALL`; `RateTrendsChartCard.tsx:84` excludes `AIAN_API` and drops `Unknown race`
- Topics differ in demographic shape — CAWP "women" race groups vs NH groups vs Hispanic-inclusive groups, and entirely different age buckets

Compare columns need the same split for the same reason: in compare-vars, column 1 may be a CAWP topic and column 2 a standard one. Fuzzy "closest match" syncing was considered and rejected as more confusing than independent state.

This is schema size, not URL size. Params are only written when a user narrows, so a typical shared URL carries zero or one.

### Encoding

- **Absent param means all groups.** `charts/trendsChart/Index.tsx:145` already normalizes "all selected" back to `[]`, so the empty state is canonical and needs no sentinel.
- **Values reuse the existing `raceNameToCodeMap` short codes** via the current `getGroupParamFromDemographicGroup` transform.
[...]

PR #4980, opened 2026-07-20
Fold chart focus into the server insight cache key

### LINKED ISSUE #4859: Include active demographic group and selected legend groups in insight cache key
## Background

PR #4850 threaded `activeDemographicGroup` (the highlighted group on a map) and `selectedGroups` (the legend selection on trend cards) into the insight prompt. Both are local React state — not URL params — so different focused views can produce different prompts from the same URL.

## Problem

The insight cache key is built entirely from the URL:

\`\`\`typescript
const cacheKey = \`\${window.location.pathname}?\${params.toString()}#\${hashId}\${cardSuffix}\`
\`\`\`

If a user generates an insight, changes their legend selection (trend cards) or highlighted group (maps), and generates again in the same session, they get the stale cached result.

## Fix

Append a focus suffix to the cache key when either context value is set:

\`\`\`typescript
const focusSuffix = [
  context?.activeDemographicGroup && context.activeDemographicGroup !== ALL
    ? context.activeDemographicGroup
    : '',
  context?.selectedGroups?.length
    ? [...context.selectedGroups].sort().join(',')
    : '',
]
  .filter(Boolean)
  .join('|')

const cacheKey = \`\${window.location.pathname}?\${params.toString()}#\${hashId}\${cardSuffix}\${focusSuffix ? \`-\${focusSuffix}\` : ''}\`
\`\`\`

This is a non-breaking change — when neither context value is set the key is identical to the current format.

## Notes

- Identified by CodeRabbit during review of #4850
- Not blocking the #4850 merge; the main fix (labeling map rows by place + group) is correct
- The `activeDemographicGroup` case may be lower risk in practice since `group1` is already a URL param and often tracks the same value — the trend card `selectedGroups` case is the cleaner win

eriwarr:
## Problem

The AI insight card's client-side cache key already varies by which group(s) the chart is focused on (the highlighted map group, or the selected trend-legend lines — see `InsightVisualizationCard.tsx`). But the **server** cache key that `generateCardInsight` passes to `fetchAIInsight` was built purely from the URL.

Focus is local React state, not a URL param. So if a user generates an insight, changes their legend selection or highlighted map group, and regenerates in the same session, the client correctly builds a fresh prompt but hands the server an identical `cacheKey` — and the server (`server/anthropic.go`) returns the previously-cached answer for the old focus.

## Fix

Extract the suffix logic into a shared `buildInsightFocusSuffix(context)` helper and apply it to both:

- the **server** key in `generateCardInsight` (the actual bug), and
- the existing **client** key in `InsightVisualizationCard` (was duplicating the logic inline).

Both caches now derive the focus suffix from one source of truth, so they can never disagree about what "the same view" means.

**Non-breaking:** with no focus, the suffix is empty and the key is byte-for-byte identical to the current format, so existing cache entries stay valid.

## Notes

- Identified by CodeRabbit during review of #4850; the client half was already done, this closes the server-side gap.
- `sanitizeInsightKey` (`server/anthropic.go`) truncates keys to 500 chars. The suffix is appended last, so a pathologically long report URL could clip it — real report URLs are well under that, so low risk in practice.
- Out of scope: `generateReportInsight` (report-level, no focus state) and `generateContrastInsight` (keys on `-contrast`, takes no group context).

## Testing

- Added a `buildInsightFocusSuffix` unit-test block (empty context, `All` ignored, sorted `selectedGroups`, combined suffix, empty array) — `generateVisualizationInsight.test.ts` passes 17/17.
- Biome clean on all three changed files.

Closes #4859


benhammondmusic (approved):
Great! Eventually we'll need to update this to reflect url encoded focus groups as proposed in #2690

Issue #5045, opened 2026-07-29
refactor(insights): build insight prompts server-side and derive cache keys from them


benhammondmusic:
## Summary

Insight prompts are assembled in the browser today and sent to the server alongside a cache key the client computed. Move prompt construction into the Go server, so the server both renders the prompt and derives the key from it.

## Why

- **One place to change tone.** Three client modules build prompts independently. Server-side templates give a single place to change tone, reading level, and output rules, instead of three that can drift apart. They already have: the report template says "8th-grade reading level" while the card and contrast templates say "8th grade reading level".
- **The server decides what reaches the model.** Today the client sends fully rendered prompt text and the server forwards it. Rendering server-side means the request carries structured data and the server controls the instructions built around it.
- **Smaller payloads.** Clients send a compact structured descriptor rather than a fully rendered prompt.

## Scope

**Phase 1: port the templates, prove against fixtures** — done in #5073 and #5074

- [x] Port the templates from `generateVisualizationInsight.ts` (775 lines), `generateReportInsight.ts` (734), and `generateContrastInsight.ts` (111) into Go
- [x] Port the row-truncation in `insightPromptBudget.ts` (52), which the builders call inline during render
- [x] Go test renders the `server/testdata/insight_prompts/*.json` inputs and reproduces each committed `.prompt.txt` byte for byte (now 12 cases; #5075 added a county report, whose geographic section ranks against sibling counties rather than spreading across children)

**Phase 2: descriptor endpoint and server-derived cache key** — done in #5075, pending merge

- [x] Define a structured request descriptor: topic / data type, geography, demographic breakdown, chart ID, and the metric rows to be described
- [x] Validate each descriptor field server-side against the known enums
- [x] Port FNV-1a 32 and derive the cache key from the rendered prompt server-side
- [x] Keep the flagging flow intact: the server returns the key it actually used, so a flagged insight can be located
- [x] Decide whether the descriptor carries the URL path and params. **Resolved: keep them.** Triage-readable keys and a warm cache both won out; the descriptor sends `urlPathname` and `urlParams`, and keys stay byte-identical to what the browser produces today

**Phase 3: frontend switches over**

- [ ] Update the call sites: `MapCard.tsx`, `CardWrapper.tsx`, `ui/InsightVisualizationCard.tsx`, `InsightReportCard.tsx`, `ContrastInsightSection.tsx`
[...]

PR #5075, opened 2026-08-04
Add /insight, which renders the prompt server-side and derives the key from it

benhammondmusic:
Phase 2 of #5045. Backend only: this adds the endpoint and proves it handles every input shape that exists today. The frontend still calls `/fetch-ai-insight` and is untouched; swapping the call sites is the follow-up PR.

## What changes

Today the client assembles both the prompt text and the cache key that indexes it, and posts both. That splits one decision across two codebases. A template edit needs a client deploy to take effect, and nothing stops a client from minting a key that disagrees with the prompt it is for. The key is a hash of the prompt, so a disagreement does not fail loudly, it just silently misses every warm entry.

`POST /insight` takes a descriptor of the view instead (kind, hash ID, topic, location, metric configs, the rendered rows, URL pathname and params). The server renders the prompt and derives the key from the rendered text, so wording and key derivation now live in one place.

Everything after that point is unchanged. The suppression, cache, ledger, and generation flow is now shared by both endpoints through `resolveInsight` rather than duplicated, so `/fetch-ai-insight` keeps its exact current behavior and serves every existing caller.

`"preview": true` stops after rendering and returns the prompt and key without consulting the cache, reserving a ledger slot, or calling the provider. That is how the frontend will verify parity before it cuts over, and it needs no API key and spends no budget.

## Coverage

`renderInsightPrompt` moved out of the test file and into production, and the golden fixture harness now renders through it. A fixture `.json` **is** a descriptor, the same body the endpoint accepts, so the 12 committed fixtures exercise the served path rather than a test-only mirror of it. The fixture set is the coverage claim, and a new surface (multimap) is covered by adding a fixture.

That retired four duplicated struct types, which removes the class of bug where a field gets added in one place and forgotten in the other.

It also surfaced one gap: a county report has no child places to spread across, so it ranks the place against its sibling counties instead, and the Go port only had the non-county branch. Rather than porting the Fips-dependent peer lookup, the descriptor carries a `peerComparison` object, the same struct the card path already receives. A 12th fixture (`report_county_peers`) pins it, and it is the only committed case whose prompt carries the en dash in the peer range line, which matters for the hash (see below).

## Two things the key derivation has to get right

Both are pinned by tests whose expected values were generated by running the browser's own algorithm under `node`, not by running the Go code. A hash that agrees only with itself proves nothing.

- **The hash runs over UTF-16 code units, not UTF-8 bytes**, because the browser iterates `charCodeAt(i)`. The peer-median line contains a U+2013 en dash, so byte iteration would diverge on every map card with a peer comparison. `TestFNV1a32MatchesBrowser` includes an en dash case and an astral-plane surrogate pair.
- **The params string is opaque text and is never parsed and re-encoded.** `URLSearchParams.toString()` preserves insertion order; Go's `url.Values.Encode()` sorts. So `stripReportInsightParam` edits the string directly, and `TestStripReportInsightParamKeepsSimilarNames` pins that `report-insights=1` and `xreport-insight=2` survive.

Keys therefore stay byte-identical to what the browser produces today, so no warm cache entry is displaced by this PR.

## Verified end to end

Ran the binary locally against the dev project's `het-insights-cache` bucket and the dev Gemini key.

Preview returns the key and the prompt, and the prompt is byte-identical to the committed fixture:

```
$ curl -s -X POST localhost:8080/insight -H 'Origin: http://localhost:3000' \
    -d @card_map_single_region_peers.json   # with "preview": true
{"cacheKey":"/exploredata?mls=1.hiv-3.00&mlp=disparity&dt1=hiv_care&group1=All#rate-map-6ca81c5f",
 "prompt":"This is a choropleth map of HIV diagnoses in Bartow County, Georgia. ..."}

$ cmp served.prompt.txt testdata/insight_prompts/card_map_single_region_peers.prompt.txt   # identical
```

The `6ca81c5f` suffix is the hash the browser algorithm produces for that same text, computed independently in `node`.

Generating writes the insight to GCS under the derived key:

```json
{"outcome":"generated","cacheKey":"/exploredata?...#rate-map-6ca81c5f","topic":"HIV diagnoses",
 "reserved":true,"model":"gemini-3.1-flash-lite","promptTokens":512,"outputTokens":53,
 "dailyGenerations":1,"dailyLimit":400,"durationMs":3358}
```

```
gs://het-insights-cache/insights//exploredata?...#rate-map-6ca81c5f.json   (267 bytes)
{"content":"With an HIV diagnosis rate of 12.4 cases per 100,000 people, Bartow County has a
 higher rate than the typical peer county, ...","timestamp":1785819322754}
```

Repeating the identical request logged `memory_hit`. The demo object was deleted from the dev bucket afterward.

Also exercised, each returning the expected key: a contrast/compare-geos descriptor (`#rate-chart-contrast-...`), and a county report, whose key correctly drops the `report-insight` param while preserving the order of the rest.

An unknown hash ID is rejected with `400 {"error":"unknown hashId ..."}`, logged as `outcome=rejected reason=invalid_descriptor`.

## Test plan

- [x] `cd server && go build ./... && go vet ./... && go test -count=1 ./...` passes
- [x] All 12 fixtures render byte-identically in Go and in TypeScript (`npx vitest run src/utils/insightPromptFixtures.test.ts`, 37 cases)
- [x] `TestFNV1a32MatchesBrowser` and `TestCacheKeyMatchesBrowser` pass against values generated by the browser algorithm in `node`
- [x] `/insight` preview returns a key and prompt via curl with no API key set
- [x] A generating request persists to GCS under the returned key, verified by reading the object back
- [x] A second identical request serves from cache without a second generation
- [x] `/fetch-ai-insight` behavior unchanged (existing tests pass untouched)

## Follow-up

Phase 3 swaps the frontend from posting prompt text to posting a descriptor, using `preview` to confirm the two render the same bytes before the switch.


eriwarr (approved):
Awesome job!

PR #5077, opened 2026-08-05
Batch tool calls in CLAUDE.md and the pr/review skills

benhammondmusic:
## Summary

Sessions in this repo were making one tool call per turn almost universally: across 45 transcripts (27,750 turns), 14,822 of 14,823 tool-invoking turns issued exactly one call. Since every tool call re-reads the whole conversation, a lone `grep` returning 200 tokens costs a re-read of a context routinely over 100K. This is the dominant cost term in a session.

Three changes, all documentation:

- **Root `CLAUDE.md`**: new `## Batch tool calls` section making batching an explicit rule, with concrete guidance (chain dependent shell steps with `&&`, prefer one precise command over a sequence of exploratory ones, treat a solo `git status` turn as a defect). Kept to ~200 tokens deliberately, since this file is re-read every turn.
- **`.claude/skills/pr/SKILL.md`**: Step 1 collapses the repo/fork/fetch/changed-files/behind-main/commit-list derivation into one batched command, and later steps reuse that output instead of re-running `git diff --name-only`. Step 3 batches the reviews, inline comments, and unresolved-thread GraphQL query, and pairs reply-and-resolve in a single call.
- **`.claude/skills/review/SKILL.md`**: Step 1 gathers PR metadata, changed files, and the full diff in one call; Step 2 instructs issuing all per-file `Read` calls in a single message rather than reading and reacting one at a time.

No behavior changes to any application code.

## Test plan

- [x] `/pr` Step 1 batched command runs end to end and prints all six labeled sections
- [x] `/review` Step 1 batched command runs end to end against an open PR
- [ ] Next `/pr` and `/review` invocation completes with the same results as before at fewer turns

### DIFF
--- a/.claude/skills/pr/SKILL.md
+++ b/.claude/skills/pr/SKILL.md
@@ -215,7 +215,7 @@ Then act:
   Reply and resolve in the same call. The mutation needs the **thread's** node ID (`PRRT_...`), not the comment's (`PRRC_...`); Step 3's first block already printed the mapping, so look it up there rather than re-querying:
   ```bash
   gh api repos/$REPO/pulls/<number>/comments/<comment_id>/replies \
-    -f body="Fixed — <one line>." >/dev/null
+    -f body="Fixed — <one line>." >/dev/null && \
   gh api graphql -f query='mutation { resolveReviewThread(input: {threadId: "<PRRT_id>"}) { thread { isResolved } } }'
   ```
   When several threads are addressed by the same commit, put every reply-and-resolve pair in one call.
--- a/.claude/skills/pr/SKILL.md
+++ b/.claude/skills/pr/SKILL.md
@@ -32,28 +32,28 @@ gh pr checkout <number>
 
 If the working tree is dirty and the checkout fails, stop and ask the user to commit or stash first — do not discard changes.
 
-Then derive two variables used throughout the remaining steps:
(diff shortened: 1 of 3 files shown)


benhammondmusic:
Fixed two of the six issues:

1. **Reply-resolve chain** — Now chains `gh api ... replies` and `resolveReviewThread` with `&&` so resolution only happens if the reply succeeds.
2. **Parameterize /review PR fetch** — Step 1 now explicitly passes `<number>` to `gh pr view` rather than relying on current branch.

Not addressing the others:

3. **Step 1b git diff range** — `git diff origin/main --name-only` is correct; it lists files changed on this branch vs main. The three-dot range would exclude files that changed on main after divergence, which is not what we want.
4. **Strict shell handling** — The tool call itself provides error handling. Adding `set -euo pipefail` to multi-line blocks is defensive good practice but the commands are logically independent and don't create risk of stale data.
5. **PR description refresh** — Edge case (would require commits landing concurrently during a single `/pr` run). The current guidance "re-run just `git log` if a commit landed" is correct and avoids verbosity.
6. **Paginate reviews/comments** — 50 is a practical ceiling for individual PRs; the largest PRs in this repo have <15 comments. Adding pagination adds complexity for a rare edge case.

PR #5089, opened 2026-08-07
fix(map): suppress No data legend swatch when orphan FIPS rows have no drawn shape

### LINKED ISSUE #5028: Map legend can show No data when no shape is drawn white
### Problem

A map legend can show a `No data` swatch when no shape on the map is actually drawn in that color, leaving the reader hunting for something that is not there.

Reproduce with Georgia county COVID-19 cases: the legend emits `No data`, but all 159 counties render with a value and zero shapes are white.

### Cause

`processLegendData` derives its absence swatches from the raw query result. `renderMap` derives its shapes from the topojson feature list. When the query returns a null row whose `fips` has no matching feature (a retired or non-standard county code, a state-level row that slipped into a county query, and similar), the legend counts an absence that the map never draws.

PR #5024 fixed the mirror image of this bug, where the map drew an absence the legend had no key for: a national map synthesizes a circle for every entry in `TERRITORY_CODES`, so a territory the source omits rendered white while never appearing in `data`. That direction was cheap because the set of territory circles is known statically. This direction is not, because it needs the actual rendered feature ids.

### Suggested fix

Have the legend reason over the same set the map draws. `renderMap` already filters `features.features` down to what it renders; passing those ids (or the resulting `dataMap`) into `processLegendData` would let it intersect the two before deciding whether an absence swatch is warranted, and would collapse both directions of the mismatch into one rule instead of two special cases.

Worth checking first whether the orphan rows are themselves a data bug. If a county-level query is returning fips codes that no county feature matches, the legend is arguably reporting a real upstream problem, and the fix belongs closer to the source.

### Scope

Frontend only. `frontend/src/charts/choroplethMap/legendDataProcessor.ts` and its caller `RateMapLegend.tsx`, plus whatever `renderMap.tsx` needs to expose.

benhammondmusic:
## Summary

- `renderMap` now extracts the filtered feature list into a named variable and returns `renderedFipsIds` — the set of FIPS codes it actually drew as SVG paths
- `ChoroplethMap` passes those IDs to `RateMapLegend` via a new `onRenderedFipsIds` callback; `RateMapLegend` forwards them to `processLegendData`
- `processLegendData` filters the missing-data check against the drawn set when available, so "No data" and "Suppressed" swatches only appear when a shape on the map actually renders white or grey

Fixes #5028

## Root cause

The CDC COVID county dataset includes rows with valid-looking FIPS codes that have no matching feature in the topojson (retired county codes or similar). `processLegendData` counted those null-value rows as missing data and emitted a "No data" swatch. `renderMap` skips those FIPS when drawing because there is no feature for them, so the swatch described a shape that did not exist. Reproduced with Georgia county COVID-19 cases: legend showed "No data" but all 159 counties rendered with a value.

## Why frontend, not pipeline

`remove_bad_fips_cols` in `cdc_restricted.py` already strips rows whose county prefix doesn't match the state FIPS, but doesn't filter against the canonical topojson feature set — that's a UI concern. The frontend fix is the right boundary: it makes the legend reflect what the map draws regardless of what any dataset emits.

## Test plan

- [x] `legendDataProcessor.test.ts` — 3 new cases: orphan-only missing (no swatch), rendered feature missing (swatch shown), orphan-only suppressed (no swatch)
- [x] `npx tsc --noEmit` — clean
- [x] `npm run cleanup` — clean
- [x] Manual: reproduce the Georgia / COVID case and verify legend no longer shows "No data"

### DIFF
--- a/frontend/src/cards/MapCard.tsx
+++ b/frontend/src/cards/MapCard.tsx
@@ -1,6 +1,6 @@
 import GridView from '@mui/icons-material/GridView'
 import { useAtomValue, useSetAtom } from 'jotai'
-import { useMemo } from 'react'
+import { useMemo, useState } from 'react'
 import { useLocation } from 'react-router'
 import { createColorScale } from '../charts/choroplethMap/colorSchemes'
 import ChoroplethMap from '../charts/choroplethMap/index'
@@ -158,6 +158,9 @@ function MapCardWithKey(props: MapCardProps) {
     ATLANTA_MODE_PARAM_KEY,
     false,
   )
+  const [renderedFipsIds, setRenderedFipsIds] = useState<
+    Set<string> | undefined
+  >()
 
   const resolvedDataTypeConfig = useMemo(
     () =>
@@ -742,6 +745,7 @@ function MapCardWithKey(props: MapCardProps) {
                       isPhrmaAdherence={isPhrmaAdherence}
                       isAtlantaMode={isAtlantaMode}
                       isSummaryLegend={isSummaryLegend}
+                      onRenderedFipsIds={setRenderedFipsIds}
                       updateFipsCallback={props.updateFipsCallback}
                       colorScale={colorScale}
                     />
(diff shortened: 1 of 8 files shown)


benhammondmusic:
Closing without merging. The spurious legend swatch is the only user-visible symptom and it requires reproducing a specific edge case with orphan FIPS codes. The fix touches 7 files and adds prop threading through the component tree, making the codebase harder to reason about for a very minor visual artifact. Not worth the complexity.

PR #5118, opened 2026-08-13
Hide share-based cards and their TOC steps for rate-only topics

eriwarr:
## Summary

Rate-only topics (e.g. **severe maternal morbidity**) listed **"Unknown demographic map"** and **"Population vs. distribution"** in the report sidebar TOC, but the cards themselves rendered nothing — so those TOC steps scrolled to an empty gap.

## Root cause

Both cards gate their content on a share metric (`shareMetricConfig`), but the gate sat **inside** an always-rendered wrapper `<div id=...>`. The sidebar TOC is built by scanning the DOM for those wrapper ids:

```ts
const hashIdsOnScreen = Object.keys(reportProviderSteps).filter(
  (key) => document.getElementById(key)?.id !== undefined,
)
```

So the id was always present → the step was always listed, even when the card inside rendered nothing. Rate-only topics have no `pct_share`, so the cards were empty but the steps remained.

## Fix

Move each gate to wrap the whole `<div id=...>`, matching the pattern already used by `rates-over-time` / `inequities-over-time` / `age-adjusted-ratios`. When there's no share metric, the id leaves the DOM, so the TOC omits the step too.

The unknowns-map gate uses `pct_share_unknown ?? pct_share` (mirroring `UnknownsMapCard`'s own internal guard) rather than just `pct_share`, which also closes a latent inconsistency where a `pct_share_unknown`-only topic would have been suppressed.

This is a general fix — it applies to every rate-only topic (`voter_participation`, `preventable_hospitalizations`, severe maternal morbidity); SMM just surfaced it.

## Verification

- `tsc`, Biome, 517 frontend unit tests — all pass.
- **In-browser** (dev backend): the severe-maternal-morbidity report no longer has `unknown-demographic-map` or `population-vs-distribution` in the DOM; the HIV report (which has a share metric) still renders both. No regression.


benhammondmusic (approved):
nice catch / cleanup

PR #5142, opened 2026-08-16
fix(ingestion): validate JSON content type for JSON file destinations in url_file_to_gcs

### LINKED ISSUE #5139: url_file_to_gcs.py caches non-JSON error responses as successful ingests
## Problem

`url_file_to_gcs.py`'s `get_first_response()` only validates the HTTP status code:

```python
def get_first_response(url_list, url_params):
    for url in url_list:
        try:
            file_from_url = requests.get(url, params=url_params, timeout=120)
            file_from_url.raise_for_status()
            return file_from_url
        except requests.HTTPError as err:
            logging.error("HTTP error for url %s: %s", url, err)
    return None
```

It never checks that the response body is actually the expected content type (JSON for ACS sources) before `download_first_url_to_gcs()` writes it to GCS and logs "Successfully uploaded data to GCS."

This surfaced concretely via #5138: Census's key-wall page is served as a 302 redirect to `missing_key.html`, which `requests` follows automatically and receives as an HTTP 200 HTML response. `raise_for_status()` sees no error, so the HTML page gets cached to GCS as if it were valid ACS JSON. Both `dagAcsConditionPreCache.yml` and `dagAcsPopulationPreCache.yml` reported `conclusion: success` while caching nothing but error pages for every year, 2009–2024.

## Why this matters beyond the Census key issue

This is a general gap: *any* source that returns a 200 with an unexpected body (maintenance page, rate-limit notice, auth wall, malformed response) will be silently cached and reported as a successful ingestion run. There's no signal to a human that the pipeline produced garbage until something downstream breaks - or worse, until a wrong number ships to the site.

## Fix

Add a lightweight content check in `get_first_response()` or `download_first_url_to_gcs()` before treating a response as valid:
- For JSON-expected sources, attempt `response.json()` and treat a parse failure as equivalent to an HTTP error (retry next URL in `url_list`, then fail loudly).
- Consider also checking `Content-Type` header as a cheap first-pass signal.
- Surface a clear failure in the DAG step (non-zero exit / explicit log line) rather than "Successfully uploaded" when validation fails, so the GitHub Actions run goes red instead of green.

This is exactly the kind of gap milestone #63 (Pipeline Trust & Automation) exists to close - "make it structurally hard for bad data to reach production silently."

snowyukitty:
## Problem
In `python/ingestion/url_file_to_gcs.py`, `get_first_response()` only checked HTTP status via `file_from_url.raise_for_status()`. When upstream providers (such as Census key-walls) return HTTP 200 with an HTML error page (e.g. redirected to `missing_key.html`), `url_file_to_gcs` would cache the invalid HTML content to GCS as if it were valid JSON (#5139).

## Solution
- Add `check_json` parameter to `get_first_response()`, `download_first_url_to_gcs()`, and `url_file_to_gcs()`.
- Automatically default `check_json=True` when `dest_filename` ends with `.json`.
- When `check_json` is enabled, attempt to parse `file_from_url.json()`; if it fails, log an error and try subsequent fallback URLs in `url_list` or return `None` (preventing upload of non-JSON error pages to GCS).
- Add unit tests in `python/tests/ingestion/test_url_file_to_gcs.py` covering valid JSON ingest, rejection of non-JSON HTML error responses, and fallback URL handling.

Fixes #5139

### DIFF
--- a/python/ingestion/url_file_to_gcs.py
+++ b/python/ingestion/url_file_to_gcs.py
@@ -13,7 +13,7 @@ def local_file_path(filename):
     return f"/tmp/{filename}"
 
 
-def url_file_to_gcs(url, url_params, gcs_bucket, dest_filename):
+def url_file_to_gcs(url, url_params, gcs_bucket, dest_filename, check_json=None):
     """
     Attempts to download a file from a url and upload as a
     blob to the given GCS bucket.
@@ -24,24 +24,31 @@ def url_file_to_gcs(url, url_params, gcs_bucket, dest_filename):
       gcs_bucket: Name of the GCS bucket to upload to (without gs://).
       dest_filename: What to name the downloaded file in GCS.
         Include the file extension.
+      check_json: Whether to validate that the downloaded content is valid JSON.
 
     Returns: A boolean indication of a file diff
     """
-    return download_first_url_to_gcs([url], gcs_bucket, dest_filename, url_params)
+    return download_first_url_to_gcs([url], gcs_bucket, dest_filename, url_params, check_json=check_json)
 
 
-def get_first_response(url_list, url_params):
+def get_first_response(url_list, url_params, check_json=False):
     for url in url_list:
         try:
             file_from_url = requests.get(url, params=url_params, timeout=120)
             file_from_url.raise_for_status()
+            if check_json:
+                try:
+                    file_from_url.json()
+                except Exception as err:
+                    logging.error("Failed to parse JSON response for url %s: %s", url, err)
+                    continue
             return file_from_url
         except requests.HTTPError as err:
             logging.error("HTTP error for url %s: %s", url, err)
     return None
 
 
-def download_first_url_to_gcs(url_list, gcs_bucket, dest_filename, url_params=None):
+def download_first_url_to_gcs(url_list, gcs_bucket, dest_filename, url_params=None, check_json=None):
     """
     Iterates over the list of potential URLs that may point to the data
     source until one of the URLs succeeds in downloading. If no URL succeeds,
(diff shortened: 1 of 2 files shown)


benhammondmusic:
Hi @snowyukitty — thanks for tackling this! Your diagnosis of #5139 was spot-on, and the fix you wrote is solid.

The timing just worked out that #5143 shipped the same feature while you were wrapping this up. The merged version uses `validate_json` (instead of `check_json`) and also integrated Census API key support from #5145/#5146, so merging this PR as-is would revert those downstream changes.

**But there's a coverage gap we can salvage:** your PR includes `testGetFirstResponse_HttpError_FallsBackToNextUrl`, which validates that HTTP errors trigger a retry to the next URL in the list. That test isn't in #5143, and it's a real requirement worth keeping.

Would you be willing to open a small follow-up PR that adds those test cases to the merged code? That way your contribution lands as an improvement to test coverage, and it's a quick merge. (Bonus: if you're comfortable with it, the merged version catches bare `Exception`, which is too broad—catching `ValueError` specifically would be tighter, since that's what `response.json()` actually raises.)

Either way, thanks for the contribution — the bugfix is live and working.

snowyukitty:
Thanks @benhammondmusic — glad the diagnosis was useful, and no problem about the overlap.

Agreed that merging this as-is would revert #5145/#5146, so I'll close this and open the follow-up you suggested.

One correction first, so the follow-up doesn't churn: the merged `get_first_response` already catches `ValueError` specifically, not bare `Exception` —

```python
except requests.HTTPError as err:
    logging.error("HTTP error for url %s: %s", url, err)
except ValueError as err:
    logging.error("Non-JSON response body from url %s: %s", url, err)
```

The bare `except Exception` was in *my* version, so the merged code is already the tighter one. Nothing to change there.

On the coverage gap, it's a bit wider than the one test. On current `main`, `MockResponse.raise_for_status()` is a no-op, so the `requests.HTTPError` branch is never exercised; and every test passes a single-URL list, so the "retry the next URL" behaviour both docstrings promise is never exercised — for HTTP errors or for invalid JSON.

So the follow-up will be test-only:

1. Give `MockResponse` an optional `status_code` (default 200) so `raise_for_status()` can raise a real `requests.HTTPError`.
2. `testGetFirstResponse_HttpError_FallsBackToNextUrl` — 503 on the first URL, valid response on the second.
3. `testGetFirstResponse_JsonValidation_FallsBackToNextUrl` — HTML body on the first URL, valid JSON on the second, with `validate_json=True`.

No production changes. Happy to trim it to just (1) and (2) if you'd rather.

PR #5148, opened 2026-08-20
test(ingestion): cover URL fallback in get_first_response

snowyukitty:
# Description and Motivation

Follow-up to #5142, at @benhammondmusic's request: port the missing URL-fallback coverage onto the merged #5143 code.

`get_first_response` promises to retry the next URL when one fails — for an HTTP error, and, with `validate_json=True`, for a non-JSON body. Neither promise was exercised:

- `MockResponse.raise_for_status()` was a no-op, so the `requests.HTTPError` branch never ran.
- Every test passed a single-URL list, so the loop never advanced past the first entry.

Tests only; no production change. (`get_first_response` already catches `ValueError` rather than a bare `Exception`, so there was nothing to tighten there.)

- Give `MockResponse` an optional `status_code` (default `200`) so `raise_for_status()` can raise a real `requests.HTTPError`.
- `testGetFirstResponse_HttpError_FallsBackToNextUrl` — 503 on the first URL, valid response on the second.
- `testGetFirstResponse_JsonValidation_FallsBackToNextUrl` — HTML body on the first URL, valid JSON on the second, under `validate_json=True`.

Both assert that the second response is returned and that `requests.get` was called twice.

## Has this been tested? How?

Against `main` at `1761bfcd`:

- `pytest python/tests/ingestion/test_url_file_to_gcs.py` — 10 passed (8 pre-existing + 2 new).
- `black --check` (line-length 120) and `pylint --rcfile=pyproject.toml` — both clean.

Mutation-checked so the new cases are load-bearing rather than decorative:

| Mutation to `get_first_response` | Result |
| --- | --- |
| remove `except requests.HTTPError` | only `testGetFirstResponse_HttpError_FallsBackToNextUrl` fails |
| remove `except ValueError` | `testGetFirstResponse_JsonValidation_FallsBackToNextUrl` fails, alongside the two pre-existing JSON tests |

The second row is the point of the new JSON case: the pre-existing tests only prove a single-URL list returns `None`, not that the loop continues to the next URL.

`status_code` defaults to `200`, so `raise_for_status()` stays a no-op for all eight pre-existing tests.

## Types of Changes

- Refactor / chore


benhammondmusic (approved):
Awesome; thanks!

snowyukitty:
Thanks for merging main and updating the assertions — `max_retries=3`, four calls, and two sleeps all look right.

One optional nit, not worth blocking on. Both fallback tests feed responses through `side_effect` by call order, so they pin *how many* requests happen rather than *which URLs* are asked for. A regression that requested `badurl.com` a fourth time instead of advancing would still consume the fourth response and still count four calls. Keying the mock by URL and asserting the sequence closes that:

```python
by_url = {"https://badurl.com": error_response, "https://goodurl.com": ok_response}
with patch("requests.get", side_effect=lambda url, **kw: by_url[url]) as mock_get, ...
self.assertEqual(
    [c.args[0] for c in mock_get.call_args_list],
    ["https://badurl.com"] * 3 + ["https://goodurl.com"],
)
```

Verified locally: with that form, removing the fallback fails exactly the two fallback tests, and removing the same-URL retry also fails the HTTP fallback test and the two added in #5153. No need to hold this PR; I can send it as a follow-up if useful.

PR #5209, opened 2026-09-02
Add per-tab URL-param overrides for every feature flag

### LINKED ISSUE #5065: Enable AI insight generation on production
## Summary

The final step of milestone #51: turn on insight generation for the public site, once every other issue in the milestone is closed.

## Why

Insights are dark in production because `VITE_SHOW_INSIGHT_GENERATION` is set in `.env.localhost`, `.env.dev`, and `.env.deploy_preview` but not in `.env.prod`. Keeping the launch as its own issue means it is a deliberate, reviewable act with an agreed set of preconditions, rather than something that rides along inside an unrelated change.

Since #5212, prod ignores env-set flags entirely, so **launching means deleting the flag gate from the code**, not adding a line to `.env.prod`. Anyone can still preview the feature against real prod data in a single browser tab with `?VITE_SHOW_INSIGHT_GENERATION=1`, which is worth doing as the last check before the gate comes out.

## Preconditions

Every other open issue in milestone #51 must be closed first. That is the team's decision on this: the feature launches complete, including the server-side prompt port in #5045, rather than in stages.

- [x] #5060 production infrastructure provisioned and verified
- [x] #5062 usage reporting in place, so spend and cache hit rate are readable from day one
- [x] #5064 prompt regression harness in place
- [x] #5045 server-side prompt construction landed and checked against the harness
- [x] #5063 generation ceilings sized against measured free-tier quota — closed via #5170
- [x] #5201 production has its own Gemini project, so internal testing cannot spend the public site's quota
- [x] #5168 burst guard, plus hardening the unprotected `/flag-insight` endpoint — closed via #5228
- [x] #5061 flagged-insight triage covering production
- [ ] #4615 internal review pass run, meeting the agreed bar
- [x] #5213 serve-side kill switch, so there is a fast off switch after the gate is gone
- [x] #5214 minimal public documentation of the feature
- [x] #5215 end-to-end coverage that insights render

`#4609 export insight` was removed from this list and closed. It was requested to support
internal testing workflows, which belong to #4615, and exporting the displayed insight to
PDF adds little over the browser's own print. A composed "generate PDF report" feature is
worth wanting but is a different and larger thing, and should start as a Discussion.

The earlier version of this list also required #5017, the ability to flush a production cache entry. That is no longer a precondition: #5196 established that the prompt-hash cache key self-invalidates, and #5197 removed the flush modes from `scripts/review_flagged_insights.sh` entirely. Removing a bad cached insight now happens through the cache-key deletion path in the review script, which #5061 already covers for production.

## Scope

- [ ] Preview on the live prod site first, via `?VITE_SHOW_INSIGHT_GENERATION=1` in one tab, before anything is removed
- [ ] Delete the `flag('VITE_SHOW_INSIGHT_GENERATION')` gate and every branch it guards, so the feature is unconditionally the product
- [ ] Remove `VITE_SHOW_INSIGHT_GENERATION` from `.env.localhost`, `.env.dev`, and `.env.deploy_preview`, since the flag no longer exists
- [ ] Cut a release
- [ ] Verify on the live site: an insight generates, the flag control works, and a flagged insight reaches the production bucket
- [ ] Watch usage and hit rate over the first week against the projections from the sizing issue
- [ ] Confirm the rollback path works before it is needed

### The first days are the atypical ones

Production starts with an empty cache, so early traffic generates at a far higher rate per pageview than steady state. The 92% hit rate measured in #5170 comes from dev traffic and is an upper bound, not a forecast — real users spread across many more topic, geography and demographic combinations. The provider's 500 requests per day is a hard wall on how fast a cold cache can be filled, whatever the ceiling is set to.

Two consequences worth planning for rather than discovering:

- Judge the first week against a cold-cache expectation, not against the dev hit rate. #5202 seeds the production cache from dev shortly before the flip, which softens this but does not remove it, and is not a precondition for launching.
- If this launch is timed to a conference or demo, that traffic arrives concentrated in time, which is the per-minute axis #5168 covers rather than either ceiling.

## Rollback

Because the launch deletes the flag rather than setting it, there is no env var to unset. The levers are:

- The operator off switch, an `insights-generation-disabled` object in the insights cache bucket, stops new generation within about a minute without a deploy. **Cached insights continue to serve.** It is checked in `resolveInsight` after the suppression check and both cache lookups, so it is a quota lever, not a content lever.
- The serving switch in #5213, checked before the cache lookups, is the one that actually takes the feature down. Production already holds 255 cached insights, so without it a content emergency has no fast answer.
- Reverting the gate-removal commit and cutting a release removes the feature from the UI entirely. Slower than the old "unset the env var" path, which is the tradeoff for prod never carrying a flag.

The distinction between the first two is the reason #5213 is a precondition. If the emergency is "we are burning quota", the existing switch is right and hiding good cached content would be a needless availability cost. If the emergency is "this content is wrong", the existing switch changes nothing a visitor sees.

Closes milestone #51.

---

## Launch date and state of play, 2026-09-03

Targeting the conference on **2026-09-24**, three weeks out.

**Infrastructure is done and was further along than this issue recorded.** Verified live
against `het-infra-prod-f6`: `frontend-service` revision `frontend-service-00008-8gb`
carries `GEMINI_API_KEY` from Secret Manager, `GEMINI_MODEL=gemini-3.1-flash-lite`,
ceilings of 300/day and 6000/month, both `het-prod-insights-cache` and
`het-prod-flagged-insights` wired, and the origin allowlist covering the apex, `www`, dev,
localhost and Netlify previews. Insight generation works on production today through the
URL-param flag.

**The cache is already seeded.** 255 insight objects landed in `het-prod-insights-cache`
on 2026-09-01, so #5202's copy has run. That issue is now rescoped to make it a repeatable
mode of the review script and to run a second pass closer to the date.

**What is genuinely left is quality and safety, not plumbing.** Of the remaining
preconditions, #5168 and #5213 close abuse and rollback gaps that only become reachable
once the feature is public, #5214 is the public disclosure, and #4615 is the only one that
asks a human judgment question. #4615 is the long pole and has been rescoped to something
one reviewer can complete and still treat as a real gate.

### Sequencing note

Do not land the gate deletion and the release cut on the same day as the conference. Cut
the release with enough margin to watch a full quota day of real traffic, so
`ceiling_approaching` and the cold-cache hit rate have a chance to say something while
there is still time to act on them.

benhammondmusic:
**Preview:** [Report with an undeclared flag armed by param](https://deploy-preview-5209--health-equity-tracker.netlify.app/exploredata?mls=1.hiv-3.00&group1=All&VITE_SHOW_NOT_A_REAL_FLAG=1)

Click the 🧑🏽‍🔬 in the toolbar and check the console: `VITE_SHOW_NOT_A_REAL_FLAG` is on via `param` even though it is declared nowhere in the codebase, alongside the two real flags this environment turns on via `env`.

## Summary

- **Any `VITE_SHOW_*` var is a feature flag, and nothing else is.** The prefix is the whole convention — there is no registry, so adding a flag is the `.env` line alone, or just the URL param.
- Every flag takes a **URL param of exactly the same name as its env var**, overriding the env for that browser tab only. Prod can now demo AI insights without a redeploy that makes them public for everyone at once.
- The override can force a flag *off* (`=0`), so the prod experience is reachable on dev — something a one-way preview switch could not do.
- `HetFeatureFlagIndicator` renders a 🧑🏽‍🔬 in the desktop and mobile toolbars whenever any flag is on, from either source, so a flagged environment is never silently flagged. Hovering, tapping, or clicking it opens a tooltip listing every armed flag and its source, and the same interaction logs a full table to the browser console, so a reviewer can find out what is on without opening devtools.

## Why this shape

Flag params are **stripped from the URL** once read. `setMadLibWithParam` rebuilds the query from a fixed allowlist (`demo`, `dt1`, `dt2`, `onboard`) on every mode change, so a param left in place would silently vanish on the first mode switch while the flag stayed on, and the URL would stop describing the state. `sessionStorage` is the single source of truth instead. Useful side effect: a link someone copies, or a screenshot they post, does not arm a flag for anyone else.

Vite emits `import.meta.env` as a whole object literal, so the computed lookup `ENV[key]` resolves fine at runtime, verified against a minified prod bundle with a runtime-opaque key. What does *not* work is enumeration: a var unset in that environment's `.env` is absent from the object entirely, so `describeFeatureFlags()` unions the env keys that are present with the override keys.

`sessionStorage` access is wrapped in try/catch. This module is imported by `CardWrapper` and `Report`, and `sessionStorage` throws outright when cookies are fully blocked, so an uncaught throw at module scope would white-screen the app for every visitor rather than just the one arming a flag.

The tooltip is controlled (state-managed) rather than left to MUI's own listeners, since a touch device never fires hover. A `ClickAwayListener` dismisses it on touch, and the tooltip carries `describeChild` so it does not take over the button's accessible name while open.

Frontend only. `INSIGHT_ALLOWED_ORIGINS` on the server already permits prod origins.

## Impact

- **Turning on a flagged feature in prod no longer requires a rebuild and redeploy.** Today the only way to see AI insights under real prod conditions is to ship them to every visitor at once; a param override makes a private dry run possible before the public flip.
- **Declarations per new flag: 2 to 1** (or 0, for a param-only experiment). Previously a flag needed both its `.env` line and a registry entry in code.
- `VITE_SHOW_CORRELATION_CARD` is unset in `.env.localhost`, so `CompareBubbleChartCard` could not be exercised locally at all. It can now.
- The indicator is a `<button>` with an accessible name and the emoji marked `aria-hidden`, satisfying WCAG 2.1 4.1.2 (Name, Role, Value) rather than shipping a bare decorative glyph.

## Test plan

New `playwright-tests/featureFlags.ci.spec.ts` runs on every PR push:

- [x] A param arms a flag declared **nowhere** (`VITE_SHOW_NOT_A_REAL_FLAG`), shows the indicator, strips the param, leaves the rest of the report state intact, and logs the exact `{on, source: 'param'}` payload
- [x] The tooltip lists every armed flag and its source on hover/tap/click, and also logs to console; tooltip dismisses on mouse-out or click-away
- [x] Forcing every flag off with `=0` removes the indicator, proving an override beats an env-on flag
- [x] The override survives navigation to a report URL that no longer carries the param
- [x] On a touch device, tapping the indicator opens the tooltip and clicking away dismisses it

Regression-checked the existing specs that touch the app bar, nav, and modals, since the desktop toolbar layout changed to seat the indicator:

- [x] `navigation.ci`, `internal_routes.ci`, `modals.ci`, `wihe.ci`, `faqs.ci`, `data_catalog.ci` — 35 passed, including the WIHE accessibility scan

Also verified in a browser:

- [x] `?VITE_SHOW_CORRELATION_CARD=1` renders the correlation card in comparevars mode without turning insights on
- [x] Indicator present and toolbars intact at 1280px and 390px viewports

Closes #5065


eriwarr (approved):

benhammondmusic:
Re CodeRabbit on `featureFlags.ci.spec.ts:11` — not changing. The two flag names are hardcoded on purpose: if a third `VITE_SHOW_*` is ever turned on by default in the preview env, this test failing is the correct signal to update it, and the comment above the test already explains the coupling. Discovering the flag list at runtime to disable it would just make the test assert against itself.

Issue #4615, opened 2026-03-07
Internal user testing infrastructure


eriwarr:

eriwarr:
Tooling for the rescoped launch pass is on branch `claude/affectionate-clarke-6xct6u`, under `frontend/scripts/insight-quality-pass/`:

- **The bar is written down** in `README.md`: four blocking codes (factual error, stigmatizing language, unsupported causal claim, revealed suppressed value; a single occurrence is a NO-GO) and three non-blocking codes (empty, above reading level, awkward; rate recorded, not gated), with a decision rule.
- **The 30-view set is recorded** in `review-set.json`: five views in each of ordinary, sparse, suppressed, county, compare and report, each with the reason it is there. Most match a committed prompt fixture. Ten carry `coldRead: true`.
- **The harvester** (`npm run insight-pass`) opens each view on production, captures the sentence, highlight, the rendered prompt via the transparency dialog preview (spends nothing) and a screenshot, and writes a worksheet with the rubric as checkboxes. `--cold-read` produces the one-question file for the non-author reader. `--tally` reads a filled worksheet back into counts and the go/no-go, and exits non-zero on a BLOCK or while any view is still PENDING.
- **Post-launch programme** opened as #5251.

Still to do, by a person, after the pre-launch re-seed:

- [ ] Run `npm run insight-pass` against production (about 7 minutes, at most 30 generations)
- [ ] Fill in the worksheet and run `--tally`
- [ ] Hand `cold-read.md` plus screenshots to one non-author for the 10-view read
- [ ] Commit `results/<run>/` and post the decision block here

The harvester was verified end to end against a local page mirroring the insight markup (card, contrast, report, no-insight path, prompt capture, tally). It has not yet been run against dev or prod from this environment, which cannot reach either site; the first live run should use `--only` on two or three views before the full set.

---
_Generated by [Claude Code](https://claude.ai/code)_

eriwarr:
The quality-pass tooling merged in #5261 (which superseded #5254 after an accidental auto-close). Relative to the earlier note here, two things changed in review: a timed-out or failed generation now stays `PENDING` rather than being recorded as `NO_INSIGHT`, and a failed prompt capture logs its reason to stderr. What remains for this issue is unchanged: run `npm run insight-pass` against production after the cache re-seed, fill the worksheet, have a non-author do the cold read, commit the `results/<run>/` folder, and post the go/no-go here.

---
[...]

PR #5234, opened 2026-09-10
Remove the AI insight feature flag, making insights unconditional

eriwarr:
**Preview:** [Insights on a report, with no flag set](https://deploy-preview-5234--health-equity-tracker.netlify.app/exploredata?mls=1.incarceration-3.00&group1=All&mlp=disparity&dt1=prison)

## Summary

- Deletes the `flag('VITE_SHOW_INSIGHT_GENERATION')` gate and every branch that only existed to hold it — six call sites across five components.
- Removes the var from `.env.localhost`, `.env.dev` and `.env.deploy_preview`, since the flag no longer exists to be read.
- Updates the four test files that used the flag as a live mechanism rather than a mention.
- Repairs `insights.spec.ts` against the #5220 transparency dialog — a pre-existing nightly failure, not caused here.

## Why deleting rather than switching on

Since #5212 production ignores env-set flags entirely, so there is no `.env.prod` line to add. Launching the feature means removing its gate.

## Impact

**No user-facing change on merge.** Every non-prod env already set the flag to `1`, so dev, localhost and deploy previews behaved exactly this way before. The only environment whose behavior changes is production, and that lands at the release cut, not here.

Removes 6 conditionals, 5 now-unused imports and 3 env declarations — 85 lines deleted against 56 added.

## Test plan

- [x] Insight spec passes against a dev server built from `.env.localhost`, which after this change declares **no flags at all** — the proof the feature is unconditional rather than differently gated
- [x] All 5 `E2E_CI` specs pass against a local `build:deploy_preview`, the exact build CI makes (localhost has zero flags now and would have passed for the wrong reason)
- [x] All 4 `insights.spec.ts` nightly tests pass, including the two repaired for the #5220 dialog
- [x] Deploy preview verified in a real browser: the insight control renders on a build with no flag declared ([screenshot](https://deploy-preview-5234--health-equity-tracker.netlify.app/exploredata?mls=1.incarceration-3.00&group1=All&mlp=disparity&dt1=prison))
- [ ] Post-release: verify on the live site that an insight generates, the flag control works, and a flagged insight reaches the production bucket

## Not included

Part of #5065, which is deliberately **not** closed by this PR. Its remaining scope is the release cut, live verification, the first week of usage watching, and confirming the rollback path. #4615 is also still open.

Two things from #5065 worth carrying forward: judge week one against a cold-cache expectation rather than the 92% dev hit rate, and do not land the release on conference day — leave a full quota day of real traffic to read first.


frontend/CLAUDE.md
```diff
@@ -204,7 +204,7 @@ VITE_SHOW_MY_FEATURE=1            # in .env.localhost / .env.dev / .env.deploy_p
 
 Read it with `flag('VITE_SHOW_MY_FEATURE')`. The identical string appears in the `.env` file, the URL param, and the call site. It is verbose, which is appropriate: a flag reference should look temporary at the point of use. The tradeoff of having no registry is no typo safety — a misspelled key reads `false` forever rather than failing to compile.
 
-One rule covers env values and params alike: present, non-empty, and not `0` means on. A param **overrides** the env for that browser tab only, so `?VITE_SHOW_INSIGHT_GENERATION=0` is how you see the prod experience on dev.
+One rule covers env values and params alike: present, non-empty, and not `0` means on. A param **overrides** the env for that browser tab only, so `?VITE_SHOW_CORRELATION_CARD=0` is how you see the prod experience on dev.
```
benhammondmusic, line 207: This doc was wrong before and still is. It's not really seeing "prod on dev" it's enabling a flag per browser tab without needing to commit an env to the repo and enable the flag for everyone. Useful for previewing a feature on public facing environments (prod)

benhammondmusic (approved):
Awesome! If you don't mind updating that Claude doc to be more accurate before you merge please do.

